PR #2, opened 2017-09-16
Refactor imports to follow pep8

brianspiering:
I had trouble getting the env setup and debugging within in this file.

It was unclear which items were Standard Library packages and which items where 3rd party packages for me to install. There was also unnecessary duplication.

I hope this helps.


jph00:
Whilst I know it's not officially recommended, I'm not a fan of using lots of vertical space, and like to be able to see everything in one screen where possible. Thanks anyways for the PR! :)

PR #3, opened 2017-09-18
Add requirements.txt

ecvgit:
I added a requirements.txt file that works with the notebooks. should help in setting this up.


jph00:
Could you please remove the specific version numbers unless there are cases where they're needed? Or perhaps replace with '>='?

ecvgit:
I just realized that the project only works with Python 3.6 and also pytorch requires installation from a specific channel. We can specify in the README about those two requirements (and use a requirements.txt file for all other dependencies) or I can export a conda yml file that takes care of all these things automatically. I am not sure which approach would you like to use in the class. Probably the conda approach is newbie friendly but requirements.txt is more flexible..

jph00:
Thanks for checking in. I would prefer the conda approach for the class since that's what we'll be teaching the students. We can still have a requirements.txt for those that are familiar and comfortable with that, of course.

ecvgit:
I add both an environment.yml for conda as well as requirements.txt

jph00:
Thanks for the update. I'm not an expert in conda config by any means, but I think that you can:

- Remove the version numbers from all the conda packages, except for pytorch, which we should keep to 0.2 (since we're using some stuff there that's likely to change)
- Move everything from the 'pip:' section to the main dependencies section, except for opencv-python (which isn't in conda) and anything else that's not in conda
- Replace all the '==' with '>=' except for pytorch, for requirements.txt
- For conda's python, 3.7 works fine too, so perhaps anything >=3.6 should be acceptable

ecvgit:
I have removed the versions from environment.yml file and moved several packages from the pip section to the conda section.  It seems that the bcoltz that comes with conda doesn't work very well (It is an older version) -- so I have added it to pip. 
BTW, I personally prefer the setup where all the version numbers are explicitly specified -- so that a student can start with a new environment for the course and we know that the particular environment will definitely work.

eduedix:
How do you manage to install packages in environment.yaml? I am currently using pipenv, a wrapper around virtualenv. It seems installing packages in `requirements.txt` was not enough. For example `graphviz` is missing from it. Would it be possible for you to update `requirements.txt` file?

PR #9, opened 2017-10-04
Learner.predict fix

szymonm:


szymonm:
hey @jph00 @yanneta, this is a n obvious bug fix. Can you have a look?

jph00:
Not quite right - you need to pass is_test too. I'll fix and commit separately now.

PR #10, opened 2017-10-05
working on fixing transforms

yanneta:
Hi Jeremy,

Here I am working on fixing the basic transforms: 
ScaleXY
RandomScaleXY
CenterCropXY
NoCropXY

The notebook test_transforms has a few test cases. Here is one question.

What is the best way to handle "transforms_basic" and other set of transforms now that we have a transform type.

I am leaving the old transforms until we agree on the replacement. I will clean up at the end.
Please take a look and let me know what you think.

Best,
Yannet

### DIFF
--- a/fastai/transforms.py
+++ b/fastai/transforms.py
@@ -119,7 +119,7 @@ class Normalize():
     def __init__(self, m, s):
         self.m=np.array(m, dtype=np.float32)
         self.s=np.array(s, dtype=np.float32)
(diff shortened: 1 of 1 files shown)


jph00:
This looks pretty good to me. re: your question... I'm not really sure of the answer - I'd suggest try replacing the current methods and then refactoring it until you've got it simplified as much as you can in the Planet and Dogs v Cats notebooks, then let's discuss to see if we can make it even simpler!

PR #11, opened 2017-11-01
Using train and valid arguments.

trusty:
Looks like the method argument was not used. hope this is correct fix.


sampathweb:
@trusty - Your change looks right, but looks like recent change to the library has created conflicts. 
 Can you resolve the and push again for the maintainers to see.  I am just commenting here on your change.  I am not the maintainer of the project.

trusty:
@sampathweb done

trusty:
I commited another change in my master and it is part of the pull request now. that new changes are also tested. that version is atleast 3x faster than the version it replaces atleast on my machine.

trusty:
I was referring to https://github.com/fastai/fastai/pull/11/commits/e9b79ce4825283604e066b22e5c3e12a92ed6466 in my previous comment.

fastai/dataset.py
```diff
@@ -40,12 +40,18 @@ def read_dir(path, folder):
...
-                  for f in iglob(f"{full_path}/*/")])
-    fnames = [iglob(f"{full_path}/{d}/*.*") for d in all_labels]
-    pairs = [(os.path.relpath(fn,path), l) for l,f in zip(all_labels, fnames) for fn in f]
-    return list(zip(*pairs))+[all_labels]
+    labels = []
+    filenames = []
+    all_labels = []
+    pjoin = os.path.join
+    full_path = pjoin(path, folder)
+    os.chdir(full_path)
```
sampathweb: Change Directories in the middle of code may move the current working directory without explicit consent of the user of the API

sampathweb:
@trusty - I think Yannet may be referring to couple of things -
1. Jeremy likes the use of list comprehension to keep the code concise with fewer lines of code.
2. How often `read_dirs` is the bottleneck in the overall model training process.

Coming from a Software Eng. background, I agree with your spacing and style of code, but that may not consistent with the way library is written.

fastai/dataset.py
trusty: I realised that and put in a new version without these environment changes.

trusty:
Thanks, I think I have made change which also condense the function. @jph00 please comment if the latest change is ok.

I agree list comprehension are good, in this case, i had a huge training set and difference between old and new code was 32 secs vs 4 mins on my machine.

yanneta:
I merged the code. Do you know what was the slow part?

trusty:
looks like we were looping over same list multiple times.

PR #12, opened 2017-11-02
GPU / CPU compatibility

sampathweb:
I created a new function `to_gpu` in `core.py` that checks if `cuda` is available and transfers the model or tensor to GPU.  Otherwise keeps the model / tensor in CPU.  This makes the code compatible for both GPU and CPU.   The API of `to_gpu` is desiged after `to_np` that's there in `core.py`.

### DIFF
--- a/fastai/column_data.py
+++ b/fastai/column_data.py
@@ -42,7 +42,7 @@ class ColumnarModelData(ModelData):
     def get_learner(self, emb_szs, n_cont, emb_drop, out_sz, szs, drops,
                     y_range=None, use_bn=False):
         model = MixedInputModel(emb_szs, n_cont, emb_drop, out_sz, szs, drops, y_range, use_bn)
-        return StructuredLearner(self, StructuredModel(model.cuda()), opt_fn=optim.Adam)
+        return StructuredLearner(self, StructuredModel(to_gpu(model)), opt_fn=optim.Adam)
 
 
 def emb_init(x):
@@ -134,7 +134,7 @@ class CollabFilterDataset(Dataset):
 
     def get_model(self, n_factors):
         model = EmbeddingDotBias(n_factors, self.n_users, self.n_items, self.min_score, self.max_score)
-        return CollabFilterModel(model.cuda())
+        return CollabFilterModel(to_gpu(model))
 
     def get_learner(self, n_factors, val_idxs, bs, **kwargs):
         return CollabFilterLearner(self.get_data(val_idxs, bs), self.get_model(n_factors), **kwargs)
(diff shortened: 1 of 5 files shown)


apiltamang:
i think it's a decent start. I've always wondered why PyTorch makes it hard to separate the logic between GPU and CPU in this manner.

I will test your commit out, and if you're cool, I might add in a support for multi-gpu resource utilization. PyTorch makes using multi-gpu super-easy, so maybe worth giving it a shot. But, instead of utilizing all GPUs by default, wonder how hard it might be to specify how many GPU's to use!

yanneta:
Ramesh,
Have you test the code with and without GPU to make sure it works?

Best,
Yannet

sampathweb:
@apiltamang - I would prefer to leave out the multi-gpu support out of this Pull request.  This pull request is intended to make fastai work both in CPU and GPU.  Adding multi-gpu might be better served in a separate pull request.

sampathweb:
@yanneta - Thanks for reviewing my pull request.  I am working on it now to test with the New AMI that Jeremy has created.  I will confirm here once I run through couple of lesson notebooks in courses/dl1

sampathweb:
@yanneta - I tested it on an P2 GPU instance using the AMI that Jeremy created for the course.  I ran the lesson1.ipynb for the dogs/cats notebook.

yanneta:
Ramesh,
I am experiencing a slower notebook than before. I am not sure this is due to your changes but I want to investigate before merging. I will be working on this later today. Thanks!

yanneta:
I merge the changes. Thank you!

Coderx7:
The current implementation needs to also check for old gpus. Currently if a system has both CUDA and cuDNN installed but uses a GPU with compute capability <3 will result in failure while all the checks are seemingly OK. Incorporating a simple check such as this should solve the issue
```
def is_gpu_supported():
    for d in range(torch.cuda.device_count()):
        capability = torch.cuda.get_device_capability(d)[0]
        return False if capability < 3 else True
```

Coderx7:
The current implementation needs to also check for old GPUs. Currently if a system has CUDA and cuDNN installed but has a GPU with compute capability <3, it results in failure, while all the checks are seemingly fine! Incorporating a simple check like the one below should fix this issue : 
```
def is_gpu_supported():
    for d in range(torch.cuda.device_count()):
        capability = torch.cuda.get_device_capability(d)[0]
        return False if capability < 3 else True

USE_GPU = torch.cuda.is_available()
def to_gpu(x, *args, **kwargs):
    '''puts pytorch variable to gpu, if cuda is available and USE_GPU is set to true. '''
    return x.cuda(*args, **kwargs) if (USE_GPU and is_gpu_supported()) else x
```

PR #13, opened 2017-11-04
better loss prints: add train_loss, val_loss, and val_acc during fitting

apiltamang:
Just better print statistics during learn.fit(..). Now, there's just more information, which otherwise could be confusing.

To Test:
- Just learn notebooks as is. Notice that at the end of each epoch, you get an output that looks as the follows:

screenshot 2017-11-04 16 46 39

### DIFF
--- a/fastai/model.py
+++ b/fastai/model.py
@@ -88,13 +88,14 @@ def fit(model, data, epochs, opt, crit, metrics=None, callbacks=None, **kwargs):
             loss = stepper.step(V(x),V(y))
             avg_loss = avg_loss * avg_mom + loss * (1-avg_mom)
             debias_loss = avg_loss / (1 - avg_mom**batch_num)
-            t.set_postfix(loss=debias_loss)
+            t.set_postfix(train_loss=debias_loss)
             stop=False
             for cb in callbacks: stop = stop or cb.on_batch_end(debias_loss)
             if stop: return
 
         vals = validate(stepper, data.val_dl, metrics)
-        print(np.round([epoch, avg_loss] + vals, 6))
+        print("epoch: {:4d}, train_loss: {:10.6f}, val_loss: {:10.6f}, val_acc: {:10.6f}".\
+              format(epoch, avg_loss, vals[0], vals[1]))
         stop=False
         for cb in callbacks: stop = stop or cb.on_epoch_end(vals)
         if stop: break


yanneta:
Hi,
There are a couple of problems here. We often don't use accuracy as the metric so "val_acc" would be misleading. If you read the code carefully, "metrics" is a list of arbitrary length. "metrics" can be defined by the user.  I am not sure the best way to solve it. 

Thank you,
Yannet

apiltamang:
That is true. I think though that "train_loss", and "val_loss" would be fairly certain to be always applicable, no?

Anyways, I think there could be a design improvement here. Just printing out an array of numbers definitely seems a little crude. Let me know what a best possible alternate may be.

yanneta:
At the moment I don't have a good suggestion. Let's keep thinking about it. I think that after one day of training you get used to the numbers.

jph00:
I think this is fine as is, except that you should just write 'val_metrics:' at the end, and then show all metrics in the array there. You should test it to ensure it works correctly with multiple metrics.

jph00:
@apiltamang are you planning to fix up this PR? Or should I just close it for now?

apiltamang:
@jph00 
I can do a basic rewrite.. no problem.
As for the metrics, unfortunately, I'm not aware of all the metrics that can come by. So, definitely blindsided a bit. If it can be merged for now, I'll continue to test as I learn more of the possible metrics.

Anyways let me know.. it's not a big deal. If you'd rather close it, I'm fine.

alexandrecc:
@apiltamang , imho, inspiration from keras could be useful :
https://github.com/fchollet/keras/blob/master/keras/metrics.py
https://github.com/fchollet/keras/blob/master/keras/losses.py

oostopitre:
Can we display it as a header instead on every line? that way it does not waste space on the row and it pulls the headers from the metrics array?

jph00:
@apiltamang you don't need to know what metrics might come (and indeed you can't, since the user can write their own!) That's why I suggested "just write 'val_metrics:' at the end, and then show all metrics in the array there" - that is, simply join the metrics up into a single string, just like I do now, and don't try to name each one individually. Does that make sense?

jph00:
Closing this for now since it's not getting any action. Feel free to create a new PR fixing issues mentioned.

PR #14, opened 2017-11-08
added support for VGG16. Import as vgg_surrogate

apiltamang:
Added support for using a surrogate vgg16 model. User should be able to use a pretrained vgg16 model, just as they could use resnet, or inception.

I feel like this would be useful because many articles have been published that use the feature extraction features of VGG-16 for other analysis such as image segmentation, multi-label classifications etc. Thus, this would be a useful thing to have.

@yanneta @jph00


jph00:
This is really cool that you got this working - I've just committed a different approach instead however which requires less code, which hopefully you'll find find useful to take a look at :)

PR #15, opened 2017-11-08
CPU / GPU Compatibility in Model Summary

sampathweb:
Changed `.cuda` to `to_gpu` in model_summary


sampathweb:
There's an error in `model_summary` code, even with current codebase, so have not been able to test it.  @jph00 @yanneta.  As you can see this code doesn't have the pull request and uses `.cuda` method.  The notebook is the `lesson1.ipynb` 

screen shot 2017-11-08 at 6 24 11 am
screen shot 2017-11-08 at 6 25 27 am

jph00:
summary() currently assumes precompute=False (not intentionally - you just found a bug!)

PR #18, opened 2017-11-09
Fix/use `cls` instead of `self` in @classmethod decorated methods

suvash:
- and use `@staticmethod` where the first argument is never really used.

I was trying to follow along the codebase, and the `self` got me quite confused when reading the classmethods. Contributing this, so that other python users don't have to get as confused when codereading.

Not a big deal, once I figured that it worked fine, just that `self` didn't mean anything important in `@classmethod` methods.

Here's a reference that makes this a bit more clear. http://stackabuse.com/pythons-classmethod-and-staticmethod-explained/

### DIFF
--- a/fastai/column_data.py
+++ b/fastai/column_data.py
@@ -30,14 +30,14 @@ class ColumnarModelData(ModelData):
             DataLoader(val_ds, bs*2, shuffle=False))
 
     @classmethod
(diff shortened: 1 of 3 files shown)


suvash:
Had some rebase conflicts here. Hastily closed the PR, and couldn't reopen it. So, a new one at #19

PR #19, opened 2017-11-09
Fix/use `cls` instead of `self` in @classmethod decorated methods

suvash:
- and use `@staticmethod` where the first argument is never really used.

I was trying to follow along the codebase, and the `self` got me quite confused when reading the classmethods. (since, `self` makes more sense in the context of an object, and classmethods don't have one) Contributing this, so that other python users don't have to get as confused when codereading.

Not a big deal, once I figured that it worked fine, just that `self` didn't mean anything important in `@classmethod` methods. Please correct me if I'm wrong here.

Here's a reference that makes this a bit more clear. http://stackabuse.com/pythons-classmethod-and-staticmethod-explained/

### DIFF
--- a/fastai/dataset.py
+++ b/fastai/dataset.py
@@ -269,8 +269,8 @@ class ImageClassifierData(ImageData):
     @property
     def is_multi(self): return self.trn_dl.dataset.is_multi
 
-    @classmethod
-    def get_ds(self, fn, trn, val, tfms, test=None, **kwargs):
+    @staticmethod
+    def get_ds(fn, trn, val, tfms, test=None, **kwargs):
         res = [
             fn(trn[0], trn[1], tfms[0], **kwargs), # train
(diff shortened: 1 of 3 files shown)


fastai/dataset.py
```diff
@@ -269,8 +269,8 @@ class ImageClassifierData(ImageData):
     @property
     def is_multi(self): return self.trn_dl.dataset.is_multi
 
-    @classmethod
-    def get_ds(self, fn, trn, val, tfms, test=None, **kwargs):
+    @staticmethod
```
sampathweb: You may be right in changing this to Static Method, but it might be better to leave it as classmethod since this is only called by other classmethods.

yanneta:
I am leaving this for Jeremy @jph00 .

jph00:
Thanks - that's helpful! Feel to to fix any other places you see this

PR #20, opened 2017-11-11
Test Activations Not Creating

kevinbird15:
If a data element is created with no test directory defined, it cannot be added in afterwards.  I have a fix ready for this, but I want to tie it back to an issue as I think that is the easiest way to use Hub?


jph00:
Looks good!

PR #31, opened 2017-11-19
Fix missing ConstantPad

azupanc:
Replaced ConstantPad with AddPading. The former was removed in b1ba89663c16f81ae33220319d08a4f87988eb94 when opencv replaced PIL.


yanneta:
I will let @jph00 look at it.

jph00:
Thanks for the fix! BTW AddPadding does reflection padding by default, which could change your existing models. You can add a flag to get constant padding again.

PR #34, opened 2017-11-21
Bug: get_cv_idxs error when val_pct > 0.3

sampathweb:
`get_cv_idxs` returns empty list when `val_pct` specified is > 0.25.  This is because of the `cv_idx` parameter is defaulted to 4.  More details - http://forums.fast.ai/t/understanding-get-val-idxs/8148

Fix is based on suggestion by Adrian Galdran in the same forum post.

Testing:
Current Status:
screen shot 2017-11-21 at 11 50 45 am

Fix:
screen shot 2017-11-21 at 11 47 47 am

### DIFF
--- a/fastai/dataset.py
+++ b/fastai/dataset.py
@@ -5,12 +5,15 @@ from .transforms import *
 from .layer_optimizer import *
 from .dataloader import DataLoader
 
-def get_cv_idxs(n, cv_idx=4, val_pct=0.2, seed=42):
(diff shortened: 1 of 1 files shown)


yanneta:
What about a default of cv_idx=0 instead. Would that solve the problem?

sampathweb:
@yanneta - Yes, that totally works.  I also updated the pull request.  It's just a change to the Default value for the function.

Testing:
screen shot 2017-11-21 at 2 09 12 pm

PR #35, opened 2017-11-22
added doc for mapper

AdityaSoni19031997:
Need to add the return type of mapper also..
But was not sure whether it's a dict or a list or something else..

### DIFF
--- a/fastai/structured.py
+++ b/fastai/structured.py
@@ -326,6 +326,7 @@ def scale_vars(df, mapper):
 
 def proc_df(df, y_fld, skip_flds=None, do_scale=False, na_dict=None,
             preproc_fn=None, max_n_cat=None, subset=None, mapper=None):
+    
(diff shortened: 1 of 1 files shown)


jph00:
Your docs aren't quite right. Try running the code and make sure you understand how it works. You'll be able to see the exact return type too.

Basically the idea is: when training, you don't pass in a mapper, but one gets calculated, containing the means and standard deviations of the continuous variables in the training set. Then, for the test set, you pass in that mapper, so that you scale the test set in the same way.

AdityaSoni19031997:
updated

AdityaSoni19031997:
https://stackoverflow.com/a/40352356

AdityaSoni19031997:
The DataFrameMapper transformation tuple describes which columns should be run through the transformation.

AdityaSoni19031997:
edited again..

PR #36, opened 2017-11-22
Aditya soni19031997 patch 2

AdityaSoni19031997:
Added doc support for mapper and an example from sklearn docs


AdityaSoni19031997:
Tried to correct it further

PR #37, opened 2017-11-23
Added ability to create LanguageModelData objects from DataFrames

ohmeow:
Added ability to create LanguageModelData objects from DataFrames.

It works just like LanguageModelData except that you pass in DataFrames to be used for train, validation, and test datasets and the column name for where the text data is.  I've already tested against the "spooky author" kaggle comp as such:

```
dfs = dict(train_df=train_df_sample, val_df=train_df_sample, test_df=test_df_sample)
md = LanguageModelDataFromDataFrames(TEXT, 'text', **dfs, bs=bs, bptt=bptt, min_freq=10)

md.trn_dl.n, md.nt, len(md.trn_ds), len(md.trn_ds[0].text)
```
Using DataFrames means that this feature can support a bunch more formats (anything that can be massaged into a DataFrame) and its super trivial to create sample datasets on the fly for testing.


ohmeow:
This is going to break some things, but I think the LanguageModelData will be better for it.

Updated so that it can be created from text files or dataframes.

PR #45, opened 2017-11-28
Add the updates to plot the most uncertains

alessaww:
Feature Request: http://forums.fast.ai/t/understanding-softmax-probabilities-output-on-a-multi-class-classification-problem/8194/18

ImageModelResults class is used for Visualizing the results of an image model: 
1. Plot most correct examples
2. Plot most incorrect examples
3. Plot most uncertain examples

**Before** there was no class, but only some functions written to handle just 2 classes problems:
1. Plot the most correct cats
image
2. Plot the most incorrect cats
image
3. Plot the most uncertain, both cats and dogs
image

**Now** we instantiate an object as following `imr = ImageModelResults(data.val_ds, log_preds)` 
We exemplify by using the dataset from kaggle `the-nature-conservancy-fisheries-monitoring_sample` which has 8 classes
1. Plot the most correct images from class 0
2. Plot the most incorrect images from class 0
3. Plot the most uncertain images from class 0
image

### DIFF
--- a/fastai/plots.py
+++ b/fastai/plots.py
@@ -87,6 +87,7 @@ class ImageModelResults():
         self.ds = ds
         self.preds = np.argmax(log_preds, axis=1)
         self.probs = np.exp(log_preds)
+        self.num_classes = log_preds.shape[1]
 
     def plot_val_with_title(self, idxs, y):
         imgs = np.stack([self.ds[x][0] for x in idxs])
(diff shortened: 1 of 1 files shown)


yanneta:
The docstrings look weird. Here are two examples.

http://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html
http://docs.python-guide.org/en/latest/writing/documentation/

alessaww:
Thanks for your reply!
I will update it with the right docstrings.
+ the updates to plot the most uncertain examples by class. Because now it is displaying the most uncertain examples from all classes.

PR #46, opened 2017-11-28
Implementation of AdamW/SGDW paper in fastai

anandsaha:
Paper: https://arxiv.org/abs/1711.05101
Code associated with paper: https://github.com/loshchil/AdamW-and-SGDW/

On CIFAR10 with resnet152 compared in the notebook adamw-sgdw-demo.ipynb (see the last graph). Also available here:
https://github.com/anandsaha/fastai.part1.v2/blob/master/courses/dl1/adamw-sgdw-demo.ipynb 

I looked at the regularizer we use in the language model - there we are creating a new layer in the RNN model. That approach I thought won't fit in the Adam model. Here we are using a callback called 'WeightDecaySchedule()' to apply the weight decay. Let me know what you think.

Observations:
- wds=0.025 or below seems to give good results
- Got good results in the above notebook
- I compared SGDM with SGDW on lesson1.ipynb: SGDW had lower loss, but SGDM had slightly higher accuracy (only 3 cycles though)
- I tried to beat the results of lesson1-rxt50.ipynb (Dogs v Cats super-charged!), but could not till now. More experimentation needed. 

Usages:

Same value
```
lr=0.01
wds=0.025
learn.fit(lr, 1, wds=wds, use_wd_sched=True)
```
Differential
```
lr=[0.001, 0.01, 0.1]
wds=[0.0025, 0.025, 0.25]
learn.fit(lr, 1, wds=wds, use_wd_sched=True)
```


anandsaha:
Fixing the conflict

sampathweb:
@anandsaha - There's still a conflict

anandsaha:
A note on performance: I tried to compare performance with and without the weight decay. I considered two factors:

a) In `on_batch_begin()`, I needed to make a copy of `param_groups`. This is because we need this in 'on_batch_end()' when the optimizer would have changed the weights already. If we have 5 million parameters, then the size of `param_groups` would be approximately 40MB.

b) In `on_batch_end()`, I am applying the decay, where-in we have a loop. This will loop as many times as the number of layers. This does not add to the complexity of the algorithm.

Both the operations being vectorized and happening in GPU, I did not find any slowdown in performance.

jph00:
Wow this is genius :) Can you confirm:

- Without this enabled, the existing lessons work as before
- Weight decay with this method is effectively regularizing an over-fitting model

anandsaha:
Sure, @jph00, I will run all the notebooks and report back.

For checking overfitting, I believe the criteria to check would be if validation loss is decreasing hand-in-hand with training loss, right? I will create 4 FC layers of 1024 dim to over-fit the model. Would you suggest something else?

jph00:
I guess start with a model that over-fits, then add more and more weight decay, and you should see the training loss after a few epochs gets worse, and validation gets better. e.g. use a resnet 50 with no dropout on dogs v cats and unfreeze all layers.

anandsaha:
Hi @jph00 here is an update on the weight decaying and regularization.

With a resnet50, I switched off dropout, `unfreeze`ed all layers and added two FC layers to help overfit. First I overfitted with 8/12 epochs, then introduced weight decay for 8/12 epochs. Results are below. 

I could overfit better with Cifat10, somehow.

I also wanted to find out how does Adam and AdamW compare when doing the regularization. I did it on Cifar10 below. AdamW could recover better.

For the loss part during weight decay, I did not see training loss getting any worst than validation loss - but both were going down (almost) uniformly and the accuracy improving.

I have updated the `adamw-sgdw-demo.ipynb` notebook where these were run. 

In the meantime, I am running the other notebooks without enabling weight decay to check for regression. Should be done by tomorrow.

### Cats And Dogs (AdamW)
[](https://postimg.org/image/l6w01hoff/)

### Cifar 10 (AdamW)
[](https://postimg.org/image/ran7w2fhn/)

### Cifar 10 (Adam)
[](https://postimg.org/image/6qidxn4wb/)

PR #47, opened 2017-11-28
added some docs around the Learner object

apiltamang:


yanneta:
The docstrings look weird. Here are two examples.

http://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html
http://docs.python-guide.org/en/latest/writing/documentation/

PR #49, opened 2017-11-29
CollabFilterClassifier

lrajlich:
Made a stab at the classifier challenge jeremy suggested in lecture on monday night. My understanding is that for a classifier the differences we need are `F.cross_entropy` instead of `F.mse` for calculating loss and to use `F.sigmoid` directly to calculate the activations. 

Looking at the `EmbeddingDotBias`, the activations look like they should evaluate to sigmoid without any code change if your input is all 0 or 1.

Ran through movie lens database and got a loss of .2. 
https://github.com/lrajlich/fastai/blob/CollabFilterClassifier-notebook/courses/dl1/lesson5-movielens.ipynb

Let me know if this is correct or point me in the right direction if I'm off.

If this is right, don't merge right away, I want to validate the input that the ratings are all 0 or 1 (otherwise I assume the model won't work properly).

### DIFF
--- a/fastai/column_data.py
+++ b/fastai/column_data.py
@@ -192,21 +192,5 @@ class CollabFilterLearner(Learner):
         super().__init__(data, models, **kwargs)
         self.crit = F.mse_loss if is_reg else F.binary_cross_entropy
 
-
-# class CollabFilterClassifierLearner(Learner):
-#     def __init__(self, data, models, **kwargs):
-#         super().__init__(data, models, **kwargs)
-#         self.crit = F.cross_entropy
-
-# class CollabFilterClassifierDataset(CollabFilterDataset):
-#     def __init__(self, path, user_col, item_col, ratings):
-#         super().__init__(path, user_col, item_col, ratings)
-#         # TODO VALIDATE ratings are all 0 or 1
-#
-#     # I did not override get_model or modify EmbeddingDotBias since it will evaluate to F.sigmoid as is
-#
-#     def get_learner(self, n_factors, val_idxs, bs, **kwargs):
-#         return CollabFilterLearner(self.get_data(val_idxs, bs), self.get_model(n_factors), **kwargs)
-
 class CollabFilterModel(BasicModel):
     def get_layer_groups(self): return self.model
--- a/fastai/column_data.py
+++ b/fastai/column_data.py
@@ -129,8 +129,8 @@ class StructuredModel(BasicModel):
 
 
 class CollabFilterDataset(Dataset):
-    def __init__(self, path, user_col, item_col, ratings):
-        self.ratings,self.path = ratings.values.astype(np.float32),path
+    def __init__(self, path, user_col, item_col, ratings, is_reg=True):
+        self.ratings,self.path = ratings.values.astype(np.float32),path,is_reg
         self.n = len(ratings)
         (self.users,self.user2idx,self.user_col,self.n_users) = self.proc_col(user_col)
         (self.items,self.item2idx,self.item_col,self.n_items) = self.proc_col(item_col)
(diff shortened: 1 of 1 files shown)


yanneta:
You can look at how this is handled in convolutional networks. We just pass an argument instead of writing a new class.

lrajlich:
@yanneta sounds good, I assume you mean to accept an `is_reg` parameter similar to https://github.com/fastai/fastai/blob/master/fastai/conv_learner.py#L23., which I'll have default as `true` to preserve existing usage. Let me know otherwise. I'll also modify the model activation function so that the model will work with classes other than just `0` and `1`, similarly to ConvLearner.

yanneta:
Yes!

jph00:
@lrajlich are you planning to update this PR with @yanneta 's suggestion?

lrajlich:
yes, I haven't had a chance to work on it until today, I'm running into some issues which I'm still resolving but I should have something in a day or two.

lrajlich:
I had a chance to update the PR. The amount of places where `is_reg` has to be passed around is too high in my opinion, which can be resolved by modifying `PassThruDataset`. A more long winded way to explain it is that in ConvLearner is_reg is determined via `data.test_ds.is_reg`; however, with `CollabFilterDataset` the test_ds is a `PassThruDataset` and thus does not have the `is_reg` property on it, so I've had to pass it around. Modifying the `PassThruDataset` to try to make it inherit `BaseDataset` will resolve this issue, which I don't think will be too crazy.

More pertinently, I was seeing issues with `F.cross_entropy`... it expects a rank 2 or rank 4 tensor - the rank 2 seems obvious to me (it's a probability for each class, across each item in the test dataset), rank 4, not so much. The model in this case spits out a rank 1 tensor. It appears that ConvLearner uses `F.binary_cross_entropy` which works with arbitrary sized input and I was able to have the code run here using that. HOWEVER, in ConvLearner, the model is generating a prediction for each class, whereas the model in this case is only generating a single prediction... In order to appropriately build a classifier here, I believe the model would need to be changed to generate a vector of predictions for each class, is this correct? It's not immediately clear to me how to change the model for this. Any guidance on this would be appreciated and I'm interested to tackle the problem.

Test notebook is here: https://github.com/lrajlich/fastai/blob/CollabFilterLearner-is_reg-notebook/courses/dl1/lesson5-movielens_copy.ipynb

jph00:
I agree - that's a lot of places! Any thoughts on how to simplify?

And yes, we shouldn't use binary x-entropy here. We need to use softmax not sigmoid, and then categorical cross-entropy. We do use that in ConvLearner for single label multiclass classification, so you should be able to use a similar approach here. That will be a good exercise :) Feel free to ask on the forum if you get stuck.

jph00:
@yanneta did you have any thoughts to simplify this?

lrajlich:
Sorry, I should have responded earlier... I thought I'd have time to work on this last week but didn't. Basically I can simplify the `is_reg` parameter passing by modifying PassThruDataset have the `is_reg` property, eg, inherit `BaseDataset`. I'm not quite as sure how to address the tensor dimension issue with `F.cross_entropy` (which calls logsoftmax, then nllloss -- this appears to be the same as what ConvnetBuilder does when `is_multi` is false) - Intuitively it seems that the forward method in EmbeddingDotBias class needs to return a higher dimensional tensor than what it's currently returning... I'll look at the lesson6-rnn notebook and see if I can apply a comparable method as the flattening done there.

lrajlich:
I'm going to close this for now (it's been hanging out for a month) and touch on it in a couple of weeks when I get a chance to look into the tensor dimensionality issue with pytorch.

PR #52, opened 2017-11-30
added some docs for the WeightDrop class

apiltamang:
More docs to slowly help me complete the picture for the LSTM model we used for the NLP section. Let me know if any corrections need be made..

Thnx

### DIFF
--- a/fastai/rnn_reg.py
+++ b/fastai/rnn_reg.py
@@ -19,12 +19,46 @@ class LockedDropout(nn.Module):
 
 
 class WeightDrop(torch.nn.Module):
+    """A custom torch layer that serves as a wrapper on another torch layer.
+    Primarily responsible for updating the weights in the wrapped layer based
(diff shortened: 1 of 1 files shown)


jph00:
Thanks! A few comments:

- Rather than 'Method processes the weights in the wrapped module by essentially deleting them', you can always remove the word 'method' from the start, so it's just 'process the weights in the wrapped module by essentially deleting them'
- LockedDropout is doing variational dropout, so you should document that
- For methods where you're not sure what it's doing, don't add docs to them, but instead ask for help understanding them on the forum. I don't want us to have docs saying "I don't know what this is"!
- Ditto for even if you're not quite *sure* what's happening in a method - rather than adding a docstring, ask on the forum about it, and then document it after you're confident you understand it clearly

apiltamang:
closing this branch. I've opened a new branch where I've consolidated all the docs.

PR #55, opened 2017-12-02
Update column_data.py

AdityaSoni19031997:
Converted the rating to float..
But when i tried on my local machine the ratings were of float dtype by default

### DIFF
--- a/fastai/column_data.py
+++ b/fastai/column_data.py
@@ -139,11 +139,12 @@ class CollabFilterDataset(Dataset):
 
     @classmethod
     def from_data_frame(cls, path, df, user_name, item_name, rating_name):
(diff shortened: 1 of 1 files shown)


AdityaSoni19031997:
```
def from_data_frame(cls, path, val_idxs, df, **y**, cat_flds, bs, test_df=None):
        ((val_df, trn_df), (val_y, trn_y)) = split_by_idx(val_idxs, df, y)
        return cls.from_data_frames(path, trn_df, val_df, trn_y, val_y, cat_flds, bs, test_df=test_df)
```
Shouldn't we turn that y also as float

jph00:
I realized we shouldn't be doing this after all, since it breaks classification!

PR #57, opened 2017-12-04
Update column_data.py

AdityaSoni19031997:
added np.sqrt()


jph00:
Please test this change carefully and show whether it improves the accuracy on each of the lessons that uses it in class.

AdityaSoni19031997:
Will update in few days(2-3)...(having exams now..)

PR #58, opened 2017-12-05
docs for various methods for the NLP language-model

apiltamang:
consolidated all my docs contribution into this branch, and on top of the tip of origin/master. Followed instructions for a cleaner doc string format, including leaving out many uncertain statements.

Let me know.. thanks..

### DIFF
--- a/fastai/lm_rnn.py
+++ b/fastai/lm_rnn.py
@@ -21,10 +21,39 @@ def repackage_var(h):
 
 
 class RNN_Encoder(nn.Module):
+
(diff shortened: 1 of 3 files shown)


jph00:
Much improved!

BTW one thing that I think is missing still is an understanding of *why* we do these things - currently your docs are focused on "what" and not "why". It might be nice to mention in the docs a link to the relevant part of the AWD LSTM paper (or whichever paper is appropriate) to show which bit it's implementing. Feel free to ask if you're interested in doing this but aren't sure about something.

PR #63, opened 2017-12-08
Change dn161 act count

radekosmulski:
Either I am horribly wrong, or the creators of dn161 do not realize that they should be using powers of 2 for activation count to give their models superpowers :smile: 

The act count in conv_learner.py was 2048 but I think it should be 2208.


jph00:
I'm confused - the current counts for both dn161 and dn121 work for me. If you're getting an error, can you show me the details?

radekosmulski:
@jph00 terribly sorry, I didn't see your earlier comment. I must have done something wrong. Sorry for the trouble. Just checked it now again and not sure how I encountered this error.

PR #64, opened 2017-12-10
add an experimental folder to include your experimental notebooks and scripts

apiltamang:
Creating an experiments folder which should allow students to create notebooks and scripts in their local machine, without those content being version-controlled and pushed to remote.

Reason I think this would be a good idea is because I tend to create a new notebook and do a bunch of experiments all the time. However, whenever I want to update my code, I have to throw away those  notebooks. Also, when I create a new branch and want to make code/docs contributions, I have to again find a way to hide those experimental notebooks from git. I reckon I wouldn't want to loose my experimental content, and would also rather have them stay in a centralized folder.

Anyways, let me know if this is okay.. thnx!


yanneta:
What about making a folder outside of the library.

On Dec 10, 2017 3:11 PM, "apiltamang" <notifications@github.com> wrote:

Creating an experiments folder which should allow students to create
notebooks and scripts in their local machine, without those content being
version-controlled and pushed to remote.

Reason I think this would be a good idea is because I tend to create a new
notebook and do a bunch of experiments all the time. However, whenever I
want to update my code, I have to throw away those notebooks. Also, when I
create a new branch and want to make code/docs contributions, I have to
again find a way to hide those experimental notebooks from git. I reckon I
wouldn't want to loose my experimental content, and would also rather have
them stay in a centralized folder.

Anyways, let me know if this is okay.. thnx!
------------------------------
You can view, comment on, or merge this pull request online at:

  https://github.com/fastai/fastai/pull/64
Commit Summary

   - add an experimental folder to include your experimental notebooks and
   scripts

File Changes

   - *M* .gitignore <https://github.com/fastai/fastai/pull/64/files#diff-0>
   (2)
   - *A* courses/dl1/experiments/README.md
   <https://github.com/fastai/fastai/pull/64/files#diff-1> (2)

Patch Links:

   - https://github.com/fastai/fastai/pull/64.patch
   - https://github.com/fastai/fastai/pull/64.diff

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/fastai/fastai/pull/64>, or mute the thread

apiltamang:
you got it! Moved it to the root.. and outside of the library :)

yanneta:
What I mean is that you don't need to have it inside of the library. You
can make a directory outside and make a simlink inside that directory
pointing to the library.

If you are local l am happy to show you how to do it tomorrow.

apiltamang:
ahh I see! I've worked with sym-links a little, and I'm always a little _effy_ with them.

I figured I don't really need to push a 'experiments' folder, but just add it in .gitignore. From now on, I (or anyone) can always continue to create things in this folder without affecting version control.

So let me know if this sounds better (and less invasive :) ). If not, I'll try to work with sym-links, but do know they always manage to break down somewhere, somehow. I tried a bunch to create sym-links referencing the data via the default path in the fastai notebooks (since my data is always stored at a different root folder), but I gave up cause some line in some library always kept going belly-up!

jph00:
You can create a folder called 'tmp', or use 'tmp' as the first 3 chars of your filenames.

PR #68, opened 2017-12-13
Breaking down components of AdamW into configurable parts

anandsaha:
As per [this](http://forums.fast.ai/t/new-adamw-optimizer-now-available/8518/26?u=anandsaha) discussion, this PR allows user to determine which part of AdamW paper to apply.

(a) To only detach the weight regularization from gradient update:

```
fit(..., use_wd_sched=True)
```

(b) To normalize the wds with each cycle as well:

```
fit(..., use_wd_sched=True, norm_wds=True)
```

(c) To allow user to pass a custom function to further add a multiplicative factor to the wds:
```
f = lambda x: np.array(x.layer_opt.lrs) / x.init_lrs # An example function taking into account learnig rate
fit(..., use_wd_sched=True, norm_wds=True, wds_sched_mult=f)
```

I have updated `adamw-sgdw-demo.ipynb` with the above usages.

I found that using only (a) with wds=1e-5, and using all the three with wds=1e-3 gave me the same results. In the later, I accounted for the the fact that (b) and (c) reduces the wds further. (Graph of cell 20 in `adamw-sgdw-demo.ipynb`)

I also added a list which captures the history of the wds used: `learner.wd_sched.wds_history`

Also fixed a bug which happened in a recent check-in: [this](https://github.com/fastai/fastai/blob/master/fastai/model.py#L100) should be `for cb in callbacks: cb.on_train_end()` instead of `for cb in callbacks: cb.on_train_begin()`

I tested with  `lesson1.ipynb` and `lesson4-imdb.ipynb` to ensure nothing is broken.

In the coming days, I will experiment more with AdamW.


jph00:
Will be interested to hear how your experiments go!

anandsaha:
Sure thing @jph00! Currently a bit behind on this initiative for other work (NIPS paper challenge being one). But will be taking this up soon...

PR #69, opened 2017-12-14
torchtext UnicodeDecodeError: 'ascii' codec can't decode byte

Thamster:
Hi, 
I was getting a UnicodeDecodeError with the latest versions of python and torchtext which assume ascii text data in lesson4-imdb.ipynb. 
This fixes it.


jph00:
I think the issue is your locale - I don't think we should be forcing a decoder here.

PR #75, opened 2017-12-31
Update model.py

EKami:
Fixing the error:
```
Expected object of type Variable[torch.cuda.FloatTensor] but found type Variable[torch.cuda.LongTensor] for argument #1 ‘target’
```
for the rossman notebook. I don't know the impact of that change to the other notebooks nor do I know why it was casted to `long` in the first place. Maybe @jph00 could shed some lights on this.

### DIFF
--- a/fastai/model.py
+++ b/fastai/model.py
@@ -40,7 +40,7 @@ class Stepper():
         output = self.m(*xs)
         if isinstance(output,(tuple,list)): output,*xtra = output
         self.opt.zero_grad()
-        loss = raw_loss = self.crit(output, y.long())
+        loss = raw_loss = self.crit(output, y)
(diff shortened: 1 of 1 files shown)


Zohaggie:
Does this undo the change recently put in to fix the way the library has worked under Windows?

jph00:
Yup. That fix wasn't right 

--
Jeremy Howard


On Sun, Dec 31, 2017, at 11:34 AM, Zohaggie wrote:
> Does this undo the change recently put in to fix the way the library
> has worked under Windows?> — You are receiving this because you modified the open/close state.
> Reply to this email directly, view it on GitHub[1], or mute the
> thread[2].> 

Links:

  1. https://github.com/fastai/fastai/pull/75#issuecomment-354620442

PR #93, opened 2018-01-16
include epoch, train_loss, val_loss, and val_stats in print statements

apiltamang:
Added method for better printouts at the end of each iteration. The print method should automatically take care of variable number of validation-stats methods passed into it. 

Here're some screenshots of the results:

screenshot 2018-01-15 20 58 09

screenshot 2018-01-15 20 57 35

Sorry I don't have the entire notebooks. I just found today that my NVIDIA drivers have been totally corrupted, and now the GPUs just can't be found! Have a looong road ahead of me as I try to ameliorate this situation.


AdityaSoni19031997:
Nice one

PR #95, opened 2018-01-17
Updating model.py

irhumshafkat:
changing iter(dl) to dl, in predict_with_targs. using iter seems to load up all images at once (over 8GB of RAM + overloading 8GB swap on Ubuntu as well, when using test set for dogscats on resnet 50) and uses very high RAM, simply dl seems to function more as intended.


irhumshafkat:
After more tinkering around, turned out it was more of a PyCharm issue than the library itself, seems to run absolutely fine on Jupyter Notebooks. Closing.

PR #105, opened 2018-01-24
Default to 2 CPUs for multi-processing on OS X

sebastianruder:
`os.sched_getaffinity` is not supported on OS X. `multiprocessing.cpu_count()` is an alternative but seems to return unreliable numbers in some environments (see [here](https://github.com/python/mypy/issues/1196)). I've added a default of `2` cpus in case `os.sched_getaffinity` is not available.

Feel free to close this PR if you don't think this is necessary.

### DIFF
--- a/fastai/core.py
+++ b/fastai/core.py
@@ -78,10 +78,13 @@ def one_hot(a,c): return np.eye(c)[a]
 def partition(a, sz): return [a[i:i+sz] for i in range(0, len(a), sz)]
 
 def partition_by_cores(a):
+    return partition(a, len(a)//num_cpus() + 1)
+
+def num_cpus():
     try:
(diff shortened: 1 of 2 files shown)


apiltamang:
i think it might be better to define a method such as:
```
def get_num_cpus():
   try:
        // use os.sched_getaffinity method to try get cpu count
   except:
       // fall back to 2
  return num_cpu
```

and use this method wherever we desire to get optimal number of cpus. This would help prevent using try/catch at multiple places, as well as make the intention for using '2'  more evident.

Shouldn't take you long at all !! Also big fan of your blogs :)

sebastianruder:
Thanks for the advice! I've added a `num_cpus()` function now, which uses `os.cpu_count()` if `os.sched_getaffinity` is unavailable.

Thanks! Glad you like the blog. :)

PR #106, opened 2018-01-24
Found bug while running md.get_model in lesson4-imdb.ipynb. get_langu…

kevindewalt:
…age_model uses pad_token, not pad_idx parameter


jph00:
Looks like we've got a conflict here. Is the issue now resolved? If not, could you resolve the conflict?

PR #112, opened 2018-01-26
Print header for fit stats

MattKleinsmith:
**With changes:**

screenshot from 2018-01-26 12-53-32

---
**Before:**

screenshot from 2018-01-26 12-53-08

---

Inspired by @apiltamang's [changes](https://github.com/fastai/fastai/pull/93). With this change we'll get the clarity apiltamang provided with the succinctness of the original outputs.

Edit: I removed the 's' from 'epochs' in a second commit.


jph00:
Yes much better - I was hoping someone might do that! :)

PR #113, opened 2018-01-26
Add precompute option to set_data

MattKleinsmith:
This change allows users to get precomputed activations for new datasets after using set_data. This change won't affect user's current set_data calls, since its effect requires a new keyword argument.

### DIFF
--- a/fastai/conv_learner.py
+++ b/fastai/conv_learner.py
@@ -106,9 +106,15 @@ class ConvLearner(Learner):
     def create_empty_bcolz(self, n, name):
         return bcolz.carray(np.zeros((0,n), np.float32), chunklen=1, mode='w', rootdir=name)
 
-    def set_data(self, data):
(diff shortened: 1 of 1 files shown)


MattKleinsmith:
**A use case:**

I had three versions of a dataset. The first contained large images, the second contained center crops of size 512, and the third contained center crops of size 256. I fit with the third dataset using standard fastai usage (lesson1), which saved activations to the tmp folder. I then wanted to fit with the second dataset, which could also be sped up from saved activations, but this required me to enter the following after using set_data:

```python
learn.unfreeze()
learn.save_fc1()
learn.freeze()
learn.precompute = True
```

When I duplicated my dataset to include 128 and 64 versions, I had to use the above block two more times, and then decided to put it in the set_data method as an optional block.

jph00:
Many thanks for the clear explanation

PR #123, opened 2018-01-30
fix stretch_cv, zoom, RandomRotateZoom, and added padding function

erikgaas:
Changed im to x.
stretch_cv used to only take scale factors 1 and above. I added a function that will do reflection padding if the scale padding to maintain dimensions if the scale is between 0 and 1 on either axis.
Meanwhile, to keep RandomRotateZoom consistent to how it was working before, I added a 1+ to the random stretch.

### DIFF
--- a/fastai/transforms.py
+++ b/fastai/transforms.py
@@ -171,7 +171,9 @@ def coords2px(y, x):
     cols = np.rint([y[1], y[3], y[1], y[3]]).astype(int)
     r,c,*_ = x.shape
     Y = np.zeros((r, c))
-    Y[rows, cols] = 1
+    #Y[rows, cols] = 1
+    Y[rows[0]:rows[2], cols[0]:cols[1]] = 1
+    #Y[y[0]:y[2], y[1]:y[2]] = 1
     return Y
 
 
--- a/fastai/transforms.py
+++ b/fastai/transforms.py
@@ -16,7 +16,7 @@ def scale_min(im, targ):
 
 def zoom_cv(x,z):
     if z==0: return x
-    r,c,*_ = im.shape
+    r,c,*_ = x.shape
     M = cv2.getRotationMatrix2D((c/2,r/2),0,z+1.)
     return cv2.warpAffine(x,M,(c,r))
 
--- a/fastai/transforms.py
+++ b/fastai/transforms.py
@@ -20,13 +20,46 @@ def zoom_cv(x,z):
     M = cv2.getRotationMatrix2D((c/2,r/2),0,z+1.)
     return cv2.warpAffine(x,M,(c,r))
 
+def padding_resize_cv(x, new_r, new_c, mode=cv2.BORDER_REFLECT):
+    """Resizes an image through either padding or center cropping on each axis.
+    
+    Arguments:
+        x (array): image
+        new_r (int): target row size
+        new_c (int): target columns size"""
+    r,c,*_ = x.shape
+    r_side = (new_r - r) / 2
+    c_side = (new_c - c) / 2
+    
+    if r_side > 0:
+        r_pad = (math.floor(r_side), math.ceil(r_side))
+    else:
+        r_del = (abs(math.ceil(r_side)), abs(math.floor(r_side)))
+        x = x[ r_del[0]:r-r_del[1], :, :]
+        r_pad = (0,0)
+    if c_side > 0:
+        c_pad = (math.floor(c_side), math.ceil(c_side))
+    else:
+        c_del = (abs(math.ceil(c_side)), abs(math.floor(c_side)))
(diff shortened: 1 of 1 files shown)


jph00:
Thanks! Can you check all usages of these functions to ensure they still work OK?

jph00:
@erikgaas can you let me know whether you think you'll have a chance to check usages of these functions?

erikgaas:
@jph00 Sure! Sorry for the delay. I have looked for usages, and it looks like stretch_cv is called only by RandomRotatateZoom which is only called in planet.py, but not used in any notebooks. I've tested the functions on the sample fish image and it seems to work fine. Next I was going to go into the lesson 2 notebook and see if replacing the get_data with planet.get_data_zoom works just as well.

Ideally, I'd like to include some tests in the test_transforms.ipynb file for consistency, but I haven't found where to download the fish bounding boxes that are used. It looks like the notebook only uses one picture to do transform testing, so I started to make my own test_transform notebook that just uses the image that comes with fastai with my own manual annotation. I could add this to the end of the current test_transforms, but manual bounds for the image collides a bit with the current code which uses I think a json file.

So I'm trying to decide what would be best to add. Maybe unit tests on accuracy on some standard model, maybe dogsvscats? Or add a new notebook that tries to standardize transform testing on the sample image provided? Or if I can find the bounding boxes, I can add add a test for RandomRotateZoom there. (Along with a new padding zoom test.) I like the idea of a new notebook using the sample image, since it makes it faster for anyone to test.

What do you think?

jph00:
Thanks so much for looking at this. The bounding boxes for fish are discussed in the part1 v1 lesson 7 notebook. Here's the code to download them copied from that notebook:

```
def get_annotations():
    annot_urls = {
        '5458/bet_labels.json': 'bd20591439b650f44b36b72a98d3ce27',
        '5459/shark_labels.json': '94b1b3110ca58ff4788fb659eda7da90',
        '5460/dol_labels.json': '91a25d29a29b7e8b8d7a8770355993de',
        '5461/yft_labels.json': '9ef63caad8f076457d48a21986d81ddc',
        '5462/alb_labels.json': '731c74d347748b5272042f0661dad37c',
        '5463/lag_labels.json': '92d75d9218c3333ac31d74125f2b380a'
    }
    cache_subdir = os.path.abspath(os.path.join(path, 'annos'))
    url_prefix = 'https://kaggle2.blob.core.windows.net/forum-message-attachments/147157/'
    
    if not os.path.exists(cache_subdir):
        os.makedirs(cache_subdir)
    
    for url_suffix, md5_hash in annot_urls.iteritems():
        fname = url_suffix.rsplit('/', 1)[-1]
        get_file(fname, url_prefix + url_suffix, cache_subdir=cache_subdir, md5_hash=md5_hash)
```

erikgaas:
Perfect! Thanks. I added code to download annotations to test_transforms. RandomRotateZoom works fine, but it doesn't fit newer tfm system. I can take a peak at it. Shouldn't be too hard to get that working.

And stretch_cv and padding_resize_cv seem to be working fine. I added them as misc tests at the bottom of the notebook.

I think I found some bugs in cropping functions for bounding box though. These don't appear unless the fish has been found under the test run. Might be worth having tests that run transforms many times to uncover errors that happen randomly. Same goes for RandomRotateZoom which has a uniformly selected 4 way conditional.

jph00:
Can you tell me what you found re bugs in cropping functions for
bounding box? I'm working on this at the moment so it's a high priority
for me! :)

erikgaas:
If you try RandomCrop with a target of 800 a few times, there will be times where the bounding box is gone, but it doesn't say the fish is lost. Here is one example:

image

So I think there are some weird edge cases where the bounding box becomes incorrect. The same thing happens for CenterCrop at 900.  Everything works fine for dihedral, lighting, and flip I think because the box never gets cut off. But for any other more complicated transformations it will have trouble. 

So I'm think it is a problem wherever the conditional for "loss my fish" is.

erikgaas:
I looked over at coord2px. I think there was a problem either with the fish being off the screen or the crop taking place without two of the points. I'm actually surprised coord2px worked before because it just highlighted four points. I tried this strategy before on a different project and ran into problems when I had to do too much interpolation, causing the pixels to disapear. Instead of highlighting one point, I'm highlighting the whole rectangle. I verified it in the test_transforms notebook and it works fine on everything and now gets the crop above correct.

A problem I see with cropping though is this happens sometimes:

image

It is a correct crop, but it leaves space on the bottom. If that is the case, then it might make sense to have logic that doesn't add a bounding box if too much (if any) of the bounding box has been removed.

jph00:
I think this bug is fixed now - also all this code has been rewritten by another student to make it consistent with the more recent thread-safe approach. So I'll close this, but feel free to open a new PR or issue if anything comes up.

PR #142, opened 2018-02-11
Don't push the numericalized variable to the GPU if no GPU device exists

keithoffer:
This fixes a crash that occurs if you try and run the NLP module on a computer with no GPU device available. I noticed a USE_GPU flag in core.py that I'm ignoring, but it seems that flag is only used directly beneath it. Is there some plan for a global flag for not using GPU acceleration that I should be checking here as well?


jph00:
We're moving away from torchtext FYI (see fastai.text for early peek) so generally things will go thru `to_gpu` in the future, which checks that flag IIRC. Thanks for the PR!

Issue #139, opened 2018-02-09
CPU-only installation


levants:
It wold be better if you provide environment-cpu.yml (or something like this) without - cuda90 and - cudnn requirements
and update README.MD for CPU only installation

jph00:
Please use the forums.

PR #157, opened 2018-02-14
added a cpu only env

movefast:
#139 added a cpu only env with cuda90 and cudnn removed from the dependencies.


jph00:
Good idea :)

PR #161, opened 2018-02-16
Update imports.py

bkj:
clean up `import.py`; remove duplicate imports;


jph00:
This format takes up too much vertical space for my liking. For fastai, prefer formatting that allows more to be seen in one screen of text.

PR #164, opened 2018-02-19
Add create_sample and create_val

radekosmulski:
Adds functions for creating a sample and a validation set.

Usage: we point the function at the directory where a dataset lives and it will either append `_sample` to the path and copy files over, or for creating the validation set it will create a `val` directory and move files. For both operations files are randomly selected from a subset of specified size.

### DIFF
--- a/fastai/dataset.py
+++ b/fastai/dataset.py
@@ -71,7 +71,7 @@ def create_val(path, r):
     r (float): proportion of examples to use for validation, in the range from 0 to 1
 
     """
-    val_path = os.path.join(os.path.split(path)[0], 'val')
+    val_path = os.path.join(os.path.split(path)[0], 'valid')
     subdirs = [os.path.split(p)[1] for p in glob(os.path.join(path, '*'))]
     copy_or_move_with_subdirs(subdirs, path, val_path, r, move=True)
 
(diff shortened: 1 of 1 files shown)


fastai/dataset.py
```diff
@@ -47,6 +47,43 @@ def read_dirs(path, folder):
...
+
+    Note - this changes the dataset at <path> by moving files to the val set
+
+    Parameters:
+    -----------
+    path: dataset path
+    r (float): how many examples to use for validation, in the range from 0 to 1
+
+    """
+    val_path = os.path.join(os.path.split(path)[0], 'val')
```
mcintyre94: In `ImageClassifierData.from_paths` in this file we have a default `val_name='valid'` (and that's the convention used in the lessons), would it make sense to use the same default here? Perhaps `create_val` could take the same parameter with the same default so the methods closely mirror each other? :)

BTW, nice idea!

fastai/dataset.py
```diff
@@ -47,6 +47,43 @@ def read_dirs(path, folder):
             labels.append(label)
     return filenames, labels, all_labels
 
+def create_sample(path, r):
+    """ Takes a path to a dataset and creates a sample of specified size at <path>_sample
+
+    Parameters:
+    -----------
+    path: dataset path
+    r (float): how many examples to use as a sample, in the range from 0 to 1
```
mcintyre94: Minor wording suggestion: 'proportion of examples to use as a sample, in the range from 0 to 1'?

fastai/dataset.py
radekosmulski: Good thinking! Changes made

fastai/dataset.py
radekosmulski: changed

PR #168, opened 2018-02-21
pip editable install of fastai library

matttrent:
This PR installs the fastai library using pip's editable flag.  The library is installed into the conda environment like any other module.  However, instead of being as a static copy of the source like pip normally does, the installed package references the source folder in this directory.  

The advantage being that the package is available any time python is run in this environment (without concerning oneself with paths) AND running `git pull` will update the installed package without any additional steps.

After merging, run `conda env update` to install the editable package.

NOTE: The syntax involves a flag being passed to pip in `environment.yml`.  This support was added in conda 4.4 (released mid-2017).  Prior versions of conda will throw an (obtuse) error.


jph00:
This sounds like an interesting idea - could you help me understand it a bit better? Also please check whether it continues to support the existing approaches.

There are two ways we currently support using fastai:

- git clone, then use the notebooks (which have symlinks to fastai)
- pip install fastai

It looks like you've removed the ability to pip install fastai? Or at least you've removed it from the readme for some reason? We need to allow pip installation of the library for a number of reasons, including allowing it to be used in kaggle kernels.

Also you've removed the symlinks, which I think might stop git clone approach from working unless you're using the conda env? We don't want to *require* people to use that env.

matttrent:
Ah, apologies, I wasn't thinking beyond the class and using the environment.  Thinking about it a bit more, there's 3 cases I can imagine.

1) They want the most recent source and the whole fast.ai environment:
```
git clone ...
conda env update
```

2) They want the most recent source (updatable via git) in an environment of their choosing:
```
git clone ...
pip install -e .   # equivalent to python setup.py develop
```

3) They want the published package on PyPI:
```
pip install fastai
```

In all 3 cases, the package is installed into python like a normal pip package would.  You can then run python and `import fastai` will work from any folder.  This saves the need to symlink the package directory into each working directory.  The `git clone` approach no longer cares where the source directory is, and you can even `import fastai` in a separate repo.

In the case of approaches 1 & 2, the "installed" package is actually a symlink to the code in the cloned repo.  You can run `git pull` and the installed package is instantly updated.  Approach 1 just saves the extra step of typing `pip install -e .` into the CLI if you're creating the full environment.

If the basic approach (`pip install`) seems sound to you, I can update the README detailing the 3 approaches listed above and when they each make sense.

dwt:
I'd like to support suggesting `pip install -e .` instead of the symlinks, as it is much better practice than using symlinks. Some arguments:
* Magic Symlink no longer necessary
* Notebooks / Code can be put anywhere and still import fastai without problems
* Not needing to clutter up the repo with my own experiments means it's easier to create actual pull requests for fastai.
* It works more similar like any other python library, i.e. what people with python knowledge expect.

I think it's fine to keep the symlink if you want that for the included notebooks so they still work even if people miss the `pip install -e .` step (also good for backwards compatibility). But for moving forward I think recommending `pip install -e .` is definitely the way forward.

dwt:
Since it reads like you may not be familiar with what `pip install -e path/to/fastai/repo` does, it basically adds a symlink to the package to the virtual environment / site-packages. This means the library is importable like any installed python library, but still any update to the source code / git pull will immediately be used.

dwt:
You might also want to add the editable install of the library to the requirements.txt file and advise people to install the library that way:

```sh
python -m venv path/to/virtualenv
source path/to/virtualenv/bin/activate
pip install -r requirements.txt
```

or with pipenv (the new hot shit: <http://pipenv.org>)

```
cd path/to/repo
pipenv install -r requirements.txt
```

surmenok:
I use "pip install -e ." with fastai all the time. It's convenient to create separate projects and use fastai library there while still having ability to change fastai source code.

jph00:
Many thanks for these helpful notes.

We recommend using conda, not pip, in the course. The pip installer was created to allow kaggle kernels to work, and for those who prefer not to use anaconda. We're not going to be changing the course to be recommending pip, so these suggestions aren't really ideal (unless I'm missing something).

I'm going to close this PR for now, since it's not clear to me that it's solving a problem we really have right now (e.g. folks that want to use `pip -e` can still do so). If anyone has a suggestion for a better way to handle our conda environment, it would be useful to discuss it on the forum first rather than here so that students can participate in the discussion and add their thoughts.

PR #171, opened 2018-02-22
allow to pass a stepper class to fit()

cbarokk:
This PR is to allow implementing "selective negative sampling" as advocated in https://arxiv.org/abs/1703.07737. Basically, we want to:
- take a batch of genuine data (all implicitly labeled as 1 since this is real data)
- generate corrupted samples by copying some features from other samples within that batch (all implicitly labelled to 0 since this is corrupted data)
- do forward passes on the corrupted samples
- sort the corrupted according to the output (e.g. sigmoid)
- refactor the batch to now contain the genuine data and the most offending corrupted samples

I found that Stepper is probably the place where this should happen. I was unsure if I should pass the class or an instance to fit(). Passing the class allowed for minimal changes...

### DIFF
--- a/fastai/model.py
+++ b/fastai/model.py
@@ -61,7 +61,7 @@ def set_train_mode(m):
     else: m.train()
 
 
-def fit(model, data, epochs, opt, crit, metrics=None, callbacks=None, **kwargs):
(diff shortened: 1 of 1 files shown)


jph00:
Good idea. It actually used to be a param to fit() but seems I removed it some time!...

PR #172, opened 2018-02-22
Add tqdm refresh interval

radekosmulski:
With small batch sizes, with tunneling over ssh, my old laptop cannot handle the amount of refreshes of the Jupyter notebook. 

As far as I was able to see, there is no way of controlling this from jupyter. All you can do is ratelimit the amount of msgs per second but this leads to broken progress bar (refreshes for some initial portion of a second to then stop and jump again).

Not sure if this code is something we want in the framework but parking it here anyhow for your perusal.

### DIFF
--- a/fastai/model.py
+++ b/fastai/model.py
@@ -2,6 +2,7 @@ from .imports import *
 from .torch_imports import *
 from .core import *
 from .layer_optimizer import *
+import datetime as dt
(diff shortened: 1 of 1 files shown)


jph00:
I think this would be best solved by adding a refresh interval option to tqdm.

PR #175, opened 2018-02-23
dataset: Use csv import from standard library to read CSV files

smspillaz:
This is already in the standard library - there is no need to reimplement it and doing so could lead to some subtle bugs.

### DIFF
--- a/fastai/dataset.py
+++ b/fastai/dataset.py
@@ -63,6 +63,26 @@ def folder_source(path, folder):
     return fnames, label_arr, all_labels
 
 def parse_csv_labels(fn, skip_header=True):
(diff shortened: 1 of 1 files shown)


smspillaz:
I was also a bit surprised to see that labels were expected to be space-separated. This broke in a non-obvious way when I was working with a labelled dataset that had spaces in their labels.

I understand that this will break compatibility with existing datasets, but would it make sense to use the column separator for multiple labels as opposed to space separating them? Or at least provide the ability to set a label delimiter?

jph00:
I'm hoping to add the ability to read in a list-of-lists, so comma is the outermost delim, and space is the inner delim.

PR #187, opened 2018-03-03
Fixed lesson 4 model data loader "slice indices must be integers or None or have an __index__ method"

kwujc:


jph00:
How did you end up with a `bs` attr that's not an int?

kwujc:
Thank you for pointing that out.

I looked back and released that i did things like bs = 64/2 

Guess this was causing the issue.

PR #190, opened 2018-03-04
Add option to control shuffling of training data

aakashns:
Currently, the training the data in `ImageClassifierData` is always shuffled randomly using [`RandomSampler`](http://pytorch.org/docs/master/data.html#torch.utils.data.sampler.RandomSampler), and there is no way to control this behavior. This change introduces a new argument `shuf_trn` which allows the user to turn off shuffling.

This option can be useful in the following situations:
* You've already shuffled the CSV file and you want to train with the same deterministic set of batches in every epoch
* You want the predictions for training data in the right order (i.e. matching the rows in the CSV)
* You want to compute the activations of some intermediate layer and save them to disk

In all of the above cases, it helps to process the training data without shuffling.

### DIFF
--- a/fastai/dataset.py
+++ b/fastai/dataset.py
@@ -311,12 +311,12 @@ class ModelDataLoader():
     def dataset(self): return self.dl.dataset
 
 class ImageData(ModelData):
-    def __init__(self, path, datasets, bs, num_workers, classes):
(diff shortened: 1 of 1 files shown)


jph00:
I think you should just be able to use fix_dl in this case?

aakashns:
Shit, I totally missed that! Thanks for pointing it out.

Issue #196, opened 2018-03-07
Tornado Version Issues


coreymason:
From a fresh clone and `conda env update`, `jupyter notebook` gives me the error `'IOLoop' has no attribute 'initialized'`.

This is the same error as in https://github.com/saltstack/salt/issues/46340 so I believe this due to `conda env update` installing Tornado 5, which was very recently released (see issue).

Downgrading to Tornado **4.5.3** (with `conda install tornado=4.5.3`) **fixed** it -- the command works fine now.

Creating an issue so someone can look into this more and make any changes that might be needed -- like enforcing tornado version <5. I've seen the issue appearing on the forums today (where there was a potential fix that did **not** work for me personally -- [forums here](http://forums.fast.ai/search?q=%27IOLoop%27%20has%20no%20attribute%20%27initialized%27))

Full error:
```
Traceback (most recent call last):
  File "/home/corey/miniconda3/envs/fastai/bin/jupyter-notebook", line 7, in <module>
    from notebook.notebookapp import main
  File "/home/corey/miniconda3/envs/fastai/lib/python3.6/site-packages/notebook/notebookapp.py", line 45, in <module>
    ioloop.install()
  File "/home/corey/miniconda3/envs/fastai/lib/python3.6/site-packages/zmq/eventloop/ioloop.py", line 210, in install
    assert (not ioloop.IOLoop.initialized()) or \
AttributeError: type object 'IOLoop' has no attribute 'initialized'
```

PR #197, opened 2018-03-07
Tornado Version Issues #196

Gokkulnath:
'IOLoop' has no attribute 'initialized' Error Fix due to release of Tornado( Ver 5.0) Framework.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -88,7 +88,7 @@ dependencies:
   #- terminado
 - testpath
 - tk
-- tornado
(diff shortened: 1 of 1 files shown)


coreymason:
Looks like we had the same idea (my PR #198) but you missed a couple of the environment files.

Gokkulnath:
@coreymason  Oops! i forgot the other environment files. Thanks for fixing them as well :)

PR #208, opened 2018-03-12
Add doc string for each available model in fastai library.

PiotrCzapla:
The docs should help users to research details about the models.

I've used a decorator to keep Jeremy style and keep the model functions short.
The decorator works as it should, here is how the docs looks  like in jupyter notebook
untitled


jph00:
I had no idea that was possible - so cool! :)

PR #210, opened 2018-03-15
Generate documentation with Sphinx 

dmesquita:
The [scikit-learn documentation](http://scikit-learn.org/stable/documentation.html) is a great way to learn about Machine Learning, I think it would be cool if we made the fast.ai documentation a great way to learn about Deep Learning 🙂

This pull request just creates the first doc version from the docstrings, the goal is to expand it as we go


jph00:
Many thanks for working on this! It's a good idea to discuss ideas like this on the forum beforehand so you can work with other projects that are in place as appropriate. We've had some discussions there and have felt that we'd rather avoid auto-generating docs from the docstrings.

Could you please split this into multiple smaller PRs which we can discuss separately? Please remove the sphinx stuff and all auto-generated files. Thanks again!

PR #211, opened 2018-03-15
Add jupyterlab to fastai conda env

suvash:
- Tested to be compatbile with notebook and can switch between the two easily


jph00:
I'd rather not make this a requirement, since we don't use it in the lessons.

Issue #213, opened 2018-03-15
Lesson 4 spacy tokenizer is not initialized and throws NameError


bharadwaj6:
`NameError: name 'spacy_tok' is not defined` on line 8 in lesson4-imdb

pauloneves:
The problem was introduced in commit 3e4b5a9, till the commit before (98fad0e) everything was working fine. The code below reproduces the error:

```
%reload_ext autoreload
%autoreload 2
%matplotlib inline

from fastai.learner import *

import torchtext
from torchtext import vocab, data
from torchtext.datasets import language_modeling

from fastai.rnn_reg import *
from fastai.rnn_train import *
from fastai.nlp import *
from fastai.lm_rnn import *

import dill as pickle
print(spacy_tok)
```

it fails since spacy_tok is not defined.

xapss:
Yes, I noticed it has been refactored a bit. It's inside Tokenizer() class.
Try: `Tokenizer().spacy_tok`

urmaspitsi:
Try replacing spacy_tok with string "spacy", when initializing TEXT variable. Instead of module reference it seems to use string variable instead.

PR #214, opened 2018-03-15
fix spacy tokenizer import in lesson 4

bharadwaj6:
Fix for https://github.com/fastai/fastai/issues/213


MotorCityCobra:
I'm trying to fix this spacy problem to do lesson 4.  
I've looked through all these links and it's not clear what fixes it.  

What is this code in your 'fix spacy...' link above? I don't know what code that is? Is that run by me somehow?

PR #216, opened 2018-03-15
Custom categorical separator in csv mode

feribg:
ADDED: ability to change the separator for the categorical column when reading from CSV


feribg:
This is trivial, seems like most of the datasets I find always have spaces in the labels, which means we have to first normalize them and then again denormalize before making a submission, so maybe it's just easier to be a parameter.

PR #224, opened 2018-03-17
Change Model to eval mode before predict_array

sampathweb:
Ref: http://forums.fast.ai/t/different-probabilities-for-the-same-picture/13259

When using `learner.predict_array` the model is not switched to `eval` mode.  So this results in incorrect results as dropouts might still be active.  Fixed it by switching model to eval mode in `predict_array`.  Similar to how it's done in `predict_with_targs` -
 https://github.com/fastai/fastai/blob/master/fastai/model.py#L139


jph00:
FYI you should also call `reset` if it exists - see the other `predict*` functions for examples.

Issue #227, opened 2018-03-19
File download fails if directory does not exist


zerotosingularity:
The get_data method fails when the directory of the filename-parameter does not exist. I noticed the problem while taking "lesson6-rnn.ipynb" and downloading the following file:

```python
get_data("https://s3.amazonaws.com/text-datasets/nietzsche.txt", f'{PATH}nietzsche.txt')
```

which resulted in the error below:

```bash
---------------------------------------------------------------------------
FileNotFoundError                         Traceback (most recent call last)
<ipython-input-23-d19fa3f57c4e> in <module>()
----> 1 get_data("https://s3.amazonaws.com/text-datasets/nietzsche.txt", f'{PATH}nietzsche.txt')
      2 
      3 text = open(f'{PATH}nietzsche.txt').read()
      4 print('corpus length:', len(text))

~/fastai/courses/dl1/fastai/io.py in get_data(url, filename)
     15 
     16         with TqdmUpTo(unit='B', unit_scale=True, miniters=1, desc=url.split('/')[-1]) as t:
---> 17             urlretrieve(url, filename, reporthook=t.update_to)
     18 

~/anaconda3/envs/fastai/lib/python3.6/urllib/request.py in urlretrieve(url, filename, reporthook, data)
    256         # Handle temporary file setup.
    257         if filename:
--> 258             tfp = open(filename, 'wb')
    259         else:
    260             tfp = tempfile.NamedTemporaryFile(delete=False)

FileNotFoundError: [Errno 2] No such file or directory: 'data/nietzsche/nietzsche.txt'
```

I would assume the folder would need to be created before the file gets downloaded.

PR #228, opened 2018-03-19
[#227] Check if directory exists before download

zerotosingularity:
This should fix Issue #227 where the get_data method creates a missing folder before downloading the file.


jph00:
I'll merge it, but FYI easier is to skip the conditional and use ` os.makedirs(dirname, exists_ok=True)`

PR #230, opened 2018-03-19
Add Starting Doc String to ConvLearner

Holisticnature:
Starting small with basic edits to give some type of description of the ConvLearner class. This is my first commit to the repo, and the code of conduct is specific to standards of behavior rather than expectations of commits. I just expect we can have  a small discussion about the change, and is also why I am starting small.

### DIFF
--- a/fastai/conv_learner.py
+++ b/fastai/conv_learner.py
@@ -82,6 +82,14 @@ class ConvnetBuilder():
 
 
 class ConvLearner(Learner):
+    """
+    Class used to train a chosen supported covnet model. Eg. ResNet-34, etc.
+    Arguments:
+        data: training data for model
+        models: model architectures to base learner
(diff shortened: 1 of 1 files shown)


jph00:
I think you could provide more detail. Maybe easier to document Learner first. A learner is something that combines a ModelData object with a nn.Module object, such that you can train that module with that data.

Holisticnature:
Hi Jeremy,
I intended to add more detail to kwargs** potentially later (based on research into Learn() class). Do you want that done before this is committed to the main repo? 
I think this is better than leaving it blank, but I will add more detail as I get deeper.

jph00:
I don't think you need to add more detail to kwargs yet - once you've documented the relevant bit of `Learner` you can then include a brief summary here.

Holisticnature:
Understood. I will close this for now.

PR #232, opened 2018-03-19
Replace kaggle-cli with official kaggle API client

alup:
Since kaggle has released an official API client, we may replace `kaggle-cli` dependency with the official package. More details about the official client can be found at https://github.com/Kaggle/kaggle-api.


jph00:
I don't think we can do that without updating the lessons.

Issue #215, opened 2018-03-15
Transform is not thread safe if tfm_y=TfmType.PIXEL


PiotrCzapla:
Have a look at this line:
https://github.com/fastai/fastai/blob/13a8b6247c6278e1a56d5790beafcd6592639301/fastai/transforms.py#L172-L175

You assume that method `__call__` can be potentially accessed from multiple threads, hence you added the `RLock` there.  But as soon as `set_state` finishes the thread go to next line and the lock is freed, leaving the state open for further changes. Unfortunately the next line:
https://github.com/fastai/fastai/blob/13a8b6247c6278e1a56d5790beafcd6592639301/fastai/transforms.py#L181-L183

Assumes that the set_state stays unchanged during the whole execution, this is how RandomCrop is able to Crop both x and y in the same way. Unfortunately, the state can be changed since the `self.lock` was released.

To fix that  we should guard the whole execution like this:
```
 with self.lock:
      self.set_state()
      x,y = ((self.transform(x),y) if self.tfm_y==TfmType.NO
              else self.transform(x,y) if self.tfm_y==TfmType.PIXEL
              else self.transform_coord(x,y)) 
```

But that if I'm not mistaken that will kill any benefits that we have from having the transformations run in parallel.

The best would be to either pass state as paramters or use thread local variables (https://stackoverflow.com/questions/1408171/thread-local-storage-in-python).

I can propose a fix but since there is no unit test I'm not sure if I won't break anything. I'm trying to use the TfmType.PIXEL to create UNet implementation for the DataBowl, so once I have this somehow fixed I will add tests and then create PR.

jph00:
Thanks for posting this - we noticed this too and are working on it now! I agree with both your suggested fixes. Look for a fix shortly...

PR #234, opened 2018-03-21
Fixes for random transforms with multithreading

bearpelican:
Addressing issue: https://github.com/fastai/fastai/issues/215

Due to a race condition, images were not always being transformed with the intended random values.
This is apparent in the carvana notebook - https://github.com/fastai/fastai/blob/master/courses/dl2/carvana.ipynb
The car masks does not line up with the car images.

Debugged this with @sjdlloyd and @brettkoonce


brettkoonce:
🎉

PR #235, opened 2018-03-21
ADDED: support for both min and max zoom

feribg:
Allow transformations to accept the old singular value for max zoom amount or alternatively a tuple of (min, max) values.

### DIFF
--- a/fastai/transforms.py
+++ b/fastai/transforms.py
@@ -318,7 +318,12 @@ class RandomScale(CoordTransform):
     def set_state(self):
         self.new_sz = self.sz
         if random.random()<self.p:
-            self.mult = random.uniform(1., self.max_zoom)
+            max_z = self.max_zoom
+            min_z = 1.
+            if isinstance(self.max_zoom, tuple):
+                min_z = self.max_zoom[0]
+                max_z = self.max_zoom[1]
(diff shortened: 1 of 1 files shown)


fastai/transforms.py
```diff
@@ -318,7 +318,12 @@ def __init__(self, sz, max_zoom, p=0.75, tfm_y=TfmType.NO, sz_y=None):
     def set_state(self):
         self.new_sz = self.sz
         if random.random()<self.p:
-            self.mult = random.uniform(1., self.max_zoom)
+            max_z = self.max_zoom
+            min_z = 1.
+            if isinstance(self.max_zoom, tuple):
```
jph00: Generally best to allow other types of sequence, not just tuple. So you want:
```
if isinstance(self.max_zoom, collections.Iterable)
```

fastai/transforms.py
```diff
@@ -318,7 +318,12 @@ def __init__(self, sz, max_zoom, p=0.75, tfm_y=TfmType.NO, sz_y=None):
...
+                min_z = self.max_zoom[0]
```
jph00: I think this is more concise:
```
min_z,max_z = self.max_zoom
```

feribg:
Thanks for the feedback. Addressed it in the latest commit.

PR #237, opened 2018-03-22
Add classification capability to Structured model

vinodreddyg:
Adding support for classifications tasks using Column Data


fastai/column_data.py
```diff
@@ -5,40 +5,44 @@
 
 
 class PassthruDataset(Dataset):
-    def __init__(self,*args):
+    def __init__(self,*args, **kwargs):
```
jph00: why not say `is_reg=True` here, instead of `**kwargs`?

fromLittleAcorns:
Thanks for doing this Vinod, a very useful addition.

alew3:
Great addition! And to clarify how to use it .. Do I just have to pass is_reg=False to do data classification? Also, what kind of datatype does my y have to be ? (categorical?)

vinodreddyg:
Hey @alew3 ,

Yeah, just need to send is_reg = False when creating ColumnarModelData. If you are using proc_df to prepare the data to use with ColumnarModelData then proc_df sets the Y in the required format. Else the Y needs to be an integer-valued array representing the class. The shape of Y should be (n,)

PranY:
Thanks for adding this!

I tried using is_reg=False, out of the box, it told me my dependent variable was expected as CUDA.LONGTENSOR but it was CUDA.FLOAT32TENSOR. I tried by changing the type of my 'y' in ColumnarModelData and re-ran. It hit some cublas error at loss.backward()
screenshot-2018-6-14 exp_model

PR #239, opened 2018-03-23
fixing some broken transforms, and implementing some unimplemented tr…

sjdlloyd:
…ansforms

1. Tested that all the transforms worked with Coordinates - They all work :)
2. Implementing RandomRotateZoom. Added any relevant transforms that needed adding because of that


jph00:
Very nice. Remind me to talk to you about the 'zoom' bit - we need to think about how that interacts with cropping.

PR #243, opened 2018-03-24
Documentation

cuddle-cuddle:
Creating new documentation markup file for each fast.ai library files.


docs/transforms.md
```diff
@@ -0,0 +1,84 @@
+## [\[source\]](../fastai/transforms.js)
+
+## Classes
+
+* <a href="normalize">Normalize</a>
+* <a href="normalize">Denormalize</a>
+
+#### <a href="#normalize">class Normalize</a>
+Normalizes an image
+  * code
```
jph00: I don't think we should put the source in the docs - it's easy enough for the user to jump to the source themselves

docs/transforms.md
```diff
@@ -0,0 +1,84 @@
+## [\[source\]](../fastai/transforms.js)
+
+## Classes
+
+* <a href="normalize">Normalize</a>
```
jph00: Why not use markdown syntax links here? Also I don't think your anchor syntax is correct - have you tested it?

cuddle-cuddle: Ooops, that one is a typo. The whole thing is only 1 page long, so didn't notice that it didn't switch page. 

May I just use Sphinx? it's very very good, and easy to use. One can simply do make html and you have searchable docs etc, and links to source, etc etc. 

Rewriting everything is really really time consuming and tiring.

jph00:
Feel free to use whatever works best for you to create the skeleton! I
don't think we need to keep any of the output other than the basic
docstrings it copies over, right?

cuddle-cuddle:
I have updated documentation and made a folder sphinx_docs and complimentary readme files in docs folder. 
A shell script is included so you can just run it to generate all documentations.

docs/transforms.md
cuddle-cuddle: Fixed. 
Added sphinx_docs folder, to render this obsolete, so to say. 
Will remove in the future.

jph00:
Thanks for the updates! Just to clarify, my previous comment re sphinx is that I'm fine with you using it or anything else to create the *initial skeleton* - but it's not something I want in the ongoing code base.

What I'm hoping we can create now is a single role model documentation page for a single module, where we test out ideas for documenting fastai, until we are happy that we have something that works well. I've been thinking more about this since we initially discussed it and have some ideas about something that *might* work. :) Are you planning to drop by again, or is it best to have that discussion here in the PR?

cuddle-cuddle:
I'm gonna drop by again, hopefully on Monday. 
This is the spring break, so the little one is with me at home the whooole week... 
Starting next week I'll be a productive member of society again. 

Can't wait for some skeletons to build on! Kudos to @racheltho !

fastai/sgdr.py
```diff
@@ -122,6 +122,10 @@ def plot(self, n_skip=10, n_skip_end=5):
 
 
 class CosAnneal(LR_Updater):
+    """
+    cosine annilation. 
+    see paper: https://openreview.net/pdf?id=BJYwwY9ll
```
apiltamang: annealing :)
annilation almost sounds like annihilation, which generally brings images of large nuclear bombs. Think not generally in our business

fastai/core.py
```diff
@@ -79,7 +92,10 @@ def SGD_Momentum(momentum):
 
 def one_hot(a,c): return np.eye(c)[a]
 
-def partition(a, sz): return [a[i:i+sz] for i in range(0, len(a), sz)]
+def partition(a, sz): 
+    """splits iterables a in equal parts of size sz
+    """
```
apiltamang: Is the shape of 'a' generalizable? would be great to document if it can ...

apiltamang:
Thanks for single-handedly going this far for creating the docs, documenting etc...

For documentation, I think writing the general shape of the arrays/tensors that get passed in methods would be great. Really helps sort out what the method is about sometimes.

Also be generalizing, I mean, saying that an array is like:
(batch_size, n_channels, height, width). I realize that this is NOT always possible, so feel free to ignore otherwise.

jph00:
Closing this since new doc framework and docs are done now (see http://docs.fast.ai).

Issue #240, opened 2018-03-23
Inconsistency in dim ordering between ImageClassifierData.from_arrays and PyTorch models


raspstephan:
The issue is illustrated here: https://gist.github.com/raspstephan/df9bb81b954851e723938d74427c3ca5

In ImageClassifierData.from_arrays the attribute .sz is derived from x.shape[1], while PyTorch models need the channel dimension in the 1st axis. This prohibits the use of from_arrays with pretrained models as shown in the minimal nb.

If I am not missing something, I would suggest using [nsamples, channels, x, y] and therefore sz = x.shape[2] in from_arrays and adding documentation.

I would be happy to prepare a pull request, but wanted to make sure first I am not mistaken.

jph00:
Sounds reasonable to me. Thanks for the helpful gist.

PR #244, opened 2018-03-24
Fixed issue with dimension ordering in ImageClassifierData.from_arrays.

raspstephan:
Simple fix for the issue raised in https://github.com/fastai/fastai/issues/240 plus update in documentation. From arrays dataset now work with ConvLearner models.

### DIFF
--- a/fastai/dataset.py
+++ b/fastai/dataset.py
@@ -275,7 +275,7 @@ class ArraysDataset(BaseDataset):
     def get_x(self, i): return self.x[i]
     def get_y(self, i): return self.y[i]
     def get_n(self): return len(self.y)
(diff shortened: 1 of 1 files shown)


jph00:
Actually this isn't going to work. `from_arrays` is often used with fully connected layers that are only rank-2. We can't assume they're images. I'm going to close this for now since I'm not at all sure what the right solution is, or if there is one - I suspect there is a need to refactor the class hierarchy of ModelData somehow...

raspstephan:
Would an easy fix be to make the get_sz dependent on the shape? Something like 

sz = x.shape[2] if x.ndim == 4 else x.shape[]

PR #247, opened 2018-03-24
Doc String for Learner & ConvLearner

Holisticnature:
@jph00 I made a first attempt to document Learner & ConvLearner. Some of these I wish had a little more detail in their naming conventions. While this is [not C++](https://google.github.io/styleguide/cppguide.html#General_Naming_Rules), I do like the rule of thumb of keeping names descriptive where possible. However, this does follow the general style of other ML libraries, so maybe just getting clarification on the docstrings is enough. 
I did wanted to check on tmp_name. I was not sure exactly what this was storing? Is it activations/weights? Temporary files?

### DIFF
--- a/fastai/learner.py
+++ b/fastai/learner.py
@@ -14,6 +14,18 @@ import time
 
 class Learner():
     def __init__(self, data, models, opt_fn=None, tmp_name='tmp', models_name='models', metrics=None, clip=None):
+        """
+        Learner is a class that combines a ModelData object with a nn.Module object, such that you can train that
+        module.
+        data (ModelData): An instance of ModelData.
+        models: chosen neural architecture for solving a supported problem.
+                Eg. ResNet-34
+        opt_fn: optimizer function, uses SGD with Momentum of .9 if none.
+        tmp_name: output name of the directory containing temporary files from training process
+        models_name: output name of the directory containing the trained model
+        metrics(function): some function for evaluating a desired metric. Eg. accuracy.
+        clip: gradient clip chosen to limit the change in the gradient to prevent exploding gradients
+        """
         self.data_,self.models,self.metrics = data,models,metrics
         self.sched=None
         self.wd_sched = None
--- a/fastai/learner.py
+++ b/fastai/learner.py
@@ -15,16 +15,15 @@ import time
 class Learner():
     def __init__(self, data, models, opt_fn=None, tmp_name='tmp', models_name='models', metrics=None, clip=None):
         """
-        Learner is a class that combines a ModelData object with a nn.Module object, such that you can train that
+        Combines a ModelData object with a nn.Module object, such that you can train that
         module.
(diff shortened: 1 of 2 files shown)


fastai/learner.py
```diff
@@ -14,6 +14,18 @@
...
+        """
+        Learner is a class that combines a ModelData object with a nn.Module object, such that you can train that
+        module.
+        data (ModelData): An instance of ModelData.
+        models: chosen neural architecture for solving a supported problem.
+                Eg. ResNet-34
+        opt_fn: optimizer function, uses SGD with Momentum of .9 if none.
+        tmp_name: output name of the directory containing temporary files from training process
+        models_name: output name of the directory containing the trained model
+        metrics(function): some function for evaluating a desired metric. Eg. accuracy.
```
jph00: It's an array of functions.

fastai/learner.py
```diff
@@ -14,6 +14,18 @@
 
 class Learner():
     def __init__(self, data, models, opt_fn=None, tmp_name='tmp', models_name='models', metrics=None, clip=None):
+        """
+        Learner is a class that combines a ModelData object with a nn.Module object, such that you can train that
```
jph00: Remove "Learner is a class that"

fastai/learner.py
```diff
@@ -14,6 +14,18 @@
...
+        module.
+        data (ModelData): An instance of ModelData.
+        models: chosen neural architecture for solving a supported problem.
+                Eg. ResNet-34
```
jph00: 'ResNet-34' would not be a valid input here. If giving an example of param use, be sure it's a valid input for that param.

fastai/learner.py
```diff
@@ -14,6 +14,18 @@
...
+        opt_fn: optimizer function, uses SGD with Momentum of .9 if none.
+        tmp_name: output name of the directory containing temporary files from training process
+        models_name: output name of the directory containing the trained model
+        metrics(function): some function for evaluating a desired metric. Eg. accuracy.
+        clip: gradient clip chosen to limit the change in the gradient to prevent exploding gradients
```
jph00: It would be helpful to say what data type this is or give an example

fastai/learner.py
Holisticnature: Interesting. I was trying to mimic a similar param I saw in the library. I think this makes sense, I might have misread the example took from. Thanks for clarification.

fastai/learner.py
Holisticnature: Redundant. Understood.

fastai/learner.py
Holisticnature: Thanks!

fastai/learner.py
Holisticnature: As a point of clarification, models represents a nn.Module correct?

PR #249, opened 2018-03-24
Current proc_df changes distribution dispersion. Check examine_proc_df notebook for details.

jmhsi:


jph00:
This sounds interesting, but I'm unable to see what's going on in this PR. A PR should just contain a fix for a problem, ideally with docs and a brief test. To show the problem that's being solved, could you please create a gist with the notebook and a simple explanation?

PR #251, opened 2018-03-26
https://gist.github.com/jmhsi/7a357c56dd42f56ee7678340d30ec43b

jmhsi:


jmhsi:
gist at https://gist.github.com/jmhsi/7a357c56dd42f56ee7678340d30ec43b

jph00:
Thanks a lot for restructuring this - much more clear!

Sorry if this is a dumb question, but want to make sure I understand. The issue you're describing is that we're *first* replacing missing (with median), then *then* calculating the column stats. And your gist shows that the mapper contains the stdev and mean from after the replacement.

So.... isn't this what you actually want? I mean, the actual mean and stdev of the column really *are* now the mean and stdev with the missing vals replaced with median. It seems that your suggested way to scaling does not result in a column with a mean of zero and std of one, doesn't it? (AFAICT your gist never actually checks this, it only checks the mappers stored stats - although I may be missing something).

jmhsi:
My new proc_df does result in mean 0 standard deviation of 1, which is what cells [9] and [10] check. If you want I can run a df.describe() on them too and upload a new gist. Original proc_df is in cells [6] and [7] which shows that even though column num_2 is drawn from standard normal, the std_deviation is too low (.2295) because we've filled so much of the column with the mean/median before scaling. I would argue that changes our representation of the distribution in a way that we don't want by making it much more probability concentrated around the mean/median and with less tail probability.

I created every column from standard normal data, and in cell [12] we can see that old proc_df scaled values to be outlandish (-9.46 standard deviations away from mean 0 from a standard normal distribution at index 999998 on the validation set using old proc_df, compared to -2.17 standard deviations away from mean 0 with new proc_df). You would expect standardizing distributions that are already standard normal would hardly change their values. Old proc_df changes them greatly, new proc_df barely changes them. I can add a df.head() of the original train/valid so you can see original and scaled values if you want.

So with old proc_df, if we imagine the distribution, we scaled to a normal distribution that was much taller and skinnier than standard normal, and the StandardScaler transformed the data into a normal distribution that is much fatter and wider than standard normal. This is all knowing that the true data distribution was drawn from standard normal. New proc_df correctly fits a standard normal distribution, and transforms the data into a standard normal distribution.

PR #253, opened 2018-03-27
Convert Filenames to str - Pandas sets columns to int if it's all numerical filenames

sampathweb:
In my dataset, all filenames are named as "1.jpg", "2.jpg" etc.  So when I define the files as 
```
1,<label val>
2,<label val>
```
`ImageClassifierData.from_csv` breaks because pandas imports the filenames as integer and concatenation of `1+".jpg"` is an error.  To fix it, we just need to cast the filenames as str and that was the goal of this pull request.


jph00:
Ah yes good point

Issue #252, opened 2018-03-26
Fast.ai requires CUDA on infererence


nlothian:
I'd like to be able to run inference on a fast.ai model on a CPU. 

This _nearly_ works, but I get an error (note this stacktrace is from a pip installed version, but the issue appears to exist in latest):
File "[redacted]/venv/lib/python3.6/site-packages/fastai/learner.py", line 147, in predict_array
    def predict_array(self, arr): return to_np(self.model(V(T(arr).cuda())))

This can be fixed by removing the ".cuda()" from:

```
    def predict_array(self, arr):
        self.model.eval()
        return to_np(self.model(V(T(arr).cuda())))
```

https://github.com/fastai/fastai/blob/master/fastai/learner.py#L272

I'd submit a patch but I suspect that it would be good to be able to use CUDA when it is available.

jph00:
Just change it to use `to_gpu()` instead. That only uses the GPU if available. PR welcome!

nlothian:
That looks good. 

What version of Pytorch is .to_gpu() in? It doesn't appear to be in 0.3.1: http://pytorch.org/docs/0.3.1/tensors.html or in master: http://pytorch.org/docs/master/tensors.html

I can't work out where to look in the PyTorch source to find this, sorry.

nlothian:
That looks good. 

What version is to_gpu() available in? It doesn't appear to be in 0.3.1 and isn't in the master branch docs: http://pytorch.org/docs/master/tensors.html#torch.Tensor

I can't work out where to look in the PyTorch source tree to find it myself, sorry

PR #256, opened 2018-03-27
Allow inference on CPU

nlothian:
Fix for https://github.com/fastai/fastai/issues/252


jph00:
BTW (and not that it matters) but IIRC `T()` already does `to_gpu()` so you could probably just have removed the `.cuda()` entirely :)

PR #257, opened 2018-03-28
match upstream inception v4 changes

brettkoonce:
based on:
https://github.com/pytorch/vision/pull/43

modified final layer to restore

    nn.AdaptiveAvgPool2d((1,1))

logic from before, reenabled the pretrained=True flag.  Tested with cats/dogs dataset with basic retraining, appears to be working correctly.  Got a worse score than resnet 101, higher than resnet 34.  

Questions: 

Running produces this error:

FileNotFoundError: [Errno 2] No such file or directory: '/home/paperspace/fastai/courses/dl1/fastai/weights/inceptionv4-58153ba9.pth'

which can be solved by creating said directory and using

     wget https://s3.amazonaws.com/pytorch/models/inceptionv4-58153ba9.pth

to fetch the actual file.  Is there a better way to handle this case?

### DIFF
--- a/fastai/models/inceptionv4.py
+++ b/fastai/models/inceptionv4.py
@@ -5,7 +5,7 @@ import os
 import sys
 
 model_urls = {
-    'imagenet': 'http://webia.lip6.fr/~cadene/Downloads/inceptionv4-97ef9c30.pth'
+    'imagenet': 'https://s3.amazonaws.com/pytorch/models/inceptionv4-58153ba9.pth'
 }
 
 class BasicConv2d(nn.Module):
@@ -40,12 +40,12 @@ class Mixed_4a(nn.Module):
     def __init__(self):
         super(Mixed_4a, self).__init__()
 
-        self.branch0 = nn.Sequential(
+        self.block0 = nn.Sequential(
             BasicConv2d(160, 64, kernel_size=1, stride=1),
             BasicConv2d(64, 96, kernel_size=3, stride=1)
         )
 
-        self.branch1 = nn.Sequential(
+        self.block1 = nn.Sequential(
             BasicConv2d(160, 64, kernel_size=1, stride=1),
             BasicConv2d(64, 64, kernel_size=(1,7), stride=1, padding=(0,3)),
             BasicConv2d(64, 64, kernel_size=(7,1), stride=1, padding=(3,0)),
@@ -53,8 +53,8 @@ class Mixed_4a(nn.Module):
         )
 
     def forward(self, x):
-        x0 = self.branch0(x)
-        x1 = self.branch1(x)
+        x0 = self.block0(x)
+        x1 = self.block1(x)
         out = torch.cat((x0, x1), 1)
         return out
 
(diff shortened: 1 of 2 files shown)


jph00:
Thanks! How about using `fastai.io.get_data` to grab the weight file? If it's already there, that function will skip it.

fastai/torch_imports.py
```diff
@@ -25,6 +26,15 @@ def children(m): return m if isinstance(m, (list, tuple)) else list(m.children()
 def save_model(m, p): torch.save(m.state_dict(), p)
 def load_model(m, p): m.load_state_dict(torch.load(p, map_location=lambda storage, loc: storage))
 
+def load_pre_inception(pre, f, fn):
+    m = f()
+    path = os.path.dirname(__file__)
+    if not os.path.exists(f'{path}/weights/'):
+        os.makedirs(f'{path}/weights/')
+    if pre: get_data(m.url(), f'{path}/weights/{fn}.pth')
```
brettkoonce: @jph00 
1) all the models that use the load_pre loader (e.g. inception4, inceptionresnet_2, resnext50, resnext101, resnext101_64, wrn) will fail without pre-creating the weights directory, we can add that logic to load_pre to unify things if you would like.

2)if you like this method of passing the url can add that logic to the existing classes.  inceptionresnet_2 has a similar model structure inception4 but the other models (eg. resnext variants + wrn) will need to be customized slightly.  the torchvision models use the torch loader directly.

jph00: That sounds reasonable to me. Thanks for the suggestion.

fastai/models/inceptionv4.py
```diff
@@ -5,15 +5,15 @@
...
+    'inceptionv4': 'https://s3.amazonaws.com/pytorch/models/inceptionv4-58153ba9.pth'
 }
 
 class BasicConv2d(nn.Module):
 
     def __init__(self, in_planes, out_planes, kernel_size, stride, padding=0):
         super(BasicConv2d, self).__init__()
         self.conv = nn.Conv2d(in_planes, out_planes, kernel_size=kernel_size, stride=stride, padding=padding, bias=False) # verify bias false
-        self.bn = nn.BatchNorm2d(out_planes)
+        self.bn = nn.BatchNorm2d(out_planes, eps=0.001, momentum=0, affine=True)
```
jph00: Ouch just noticed this 0 momentum=0. That means that it *never* updates the bn statistics, right? I can't imagine why that would be a good thing - will mean you can't fine-tune as well on datasets where the new domain is very different to imagenet, like the Planet dataset.

Maybe set it to 0.9? And test fine-tuning on planet?

brettkoonce:
@jph00 
Took another pass at things.  Reverted the parameter changes to use the defaults (momentum is supposed to be .1, for what it's worth).  Went a bunch deeper into how the current models are working.  My thoughts:

1) the only other model in this directory that actually has a weights file/related logic is inceptionresnet_2.  resnext50, resnext101, resnext101_64, and wrn use the load_pre helper but do not actually have actual logic in these classes to download anything.
2) inceptionresnet_2 is not actually currently working (models file is 404).  I attempted to match upstream changes here: https://github.com/Cadene/pretrained-models.pytorch/blob/master/pretrainedmodels/models/inceptionresnetv2.py  but I need some feedback/guidance on how the conv_layer implements cut (e.g. it is not slicing the right layers)
3) if we moved inception_4 and inceptionresnet_2 to the model_zoo approach (use the torchvision loader) then the load_pre logic can be removed altogether (since the other models do not actually use it)
4) i wrote basic loaders for NASnet and inception 3, but they are not converging/training improperly (e.g. 60% accuracy on my training/validation data), both need conv_layer cut and model_features tweaks, I would like feedback/guidance there.
5) the inception_4 model fails on the planet dataset when I just blindly change the model in the dl1/course-2 notebook, don't want to hunt further right now.

jph00:
OK well I'll merge this, and we can chat about the other stuff when you're around in person if you like.

PR #258, opened 2018-03-28
Fixed bug in `model_summary` so it works for models with batchnorm

blakewest:
- Using 1 in the batch size dimension fails on models that include batch_norm (which is most). You get the error "ValueError: Expected more than 1 value per channel when training, got input size [1, 1024]"
- This was due to a change in PyTorch 0.3. See discussion here: https://github.com/pytorch/pytorch/issues/4534
- Short version is that because BatchNorm does (x - mean(x)) / std(x), the numerator will always be 0 when the batch size is only 1, and thus the output would be entirely zero, which doesn't really make sense, so PyTorch started throwing this error.
- Considering the prevalance of BatchNorm, this essentially means that batch sizes must always be greater than 1.
- I just changed it to 3, which was an arbitrary choice, but the output is correct again.

#### Before
image

#### After
image

Note the after photo is from me simply re-assigning the `model_summary` function to the new function reflected in this change. It runs and seems to print reasonable output

### DIFF
--- a/fastai/model.py
+++ b/fastai/model.py
@@ -184,8 +184,8 @@ def model_summary(m, input_size):
     m.apply(register_hook)
 
     if isinstance(input_size[0], (list, tuple)):
-        x = [to_gpu(Variable(torch.rand(1,*in_size))) for in_size in input_size]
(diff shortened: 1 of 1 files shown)


jph00:
Great detective work! :D

PR #263, opened 2018-03-29
Add docker support for fastai

Nikasa1889:


jph00:
Thanks for this useful contribution. I'd prefer to keep this in a separate repo however. Perhaps you could just create a 'fastai-docker' repo or something like that?

PR #269, opened 2018-03-29
Make to_np accept np.arrays

PiotrCzapla:
This makes fastai more consistent, otherwise few methods need tensors.
While they could be used with numpy arrays as well.
Here is an example that would break without this change:
```py
show_img(md.val_ds.denorm(md.trn_ds[0][0])[0])
```


jph00:
Good idea.

PR #270, opened 2018-03-30
Convnetbuilder with untrained resnet + fixes to transforms

sjdlloyd:
Conv_Learner change could be handy for Imagenet training


fastai/transforms.py
```diff
@@ -406,19 +406,20 @@ def do_transform(self, x, is_y):
         return x
 
 class RandomRotateZoom(CoordTransform):
-    def __init__(self, deg, zoom, stretch, mode=cv2.BORDER_REFLECT, tfm_y=TfmType.NO):
+    def __init__(self, deg, zoom, stretch, ps=(0.25,0.25,0.25,0.25), mode=cv2.BORDER_REFLECT, tfm_y=TfmType.NO):
```
jph00: I know that tuples aren't mutable, but I'm not 100% sure whether this tuple default argument is safe? See http://docs.python-guide.org/en/latest/writing/gotchas/

Also, maybe add an assertion that the prs sum to 1?

sjdlloyd:
Fair! Put a default arg in

PR #273, opened 2018-03-31
fix up `ds` vs `dl` suffixes in ModelData and ImageData

mcskinner:
ImageData takes Datasets as input, but wraps them in DataLoaders internally.


mcskinner:
Disregard, the `ds` stuff comes in from the `@property`s on the ModelData parent. Derp.

PR #274, opened 2018-04-01
fix Off-by-one error in bb_hw

pradhanbu:
- bb_hw function has a Off-by-one error
- fix and corresponding expected / actual output in courses/dl2/pascal.ipynb
- also add the fix to courses/dl2/pascal-multi.ipynb


courses/dl2/pascal.ipynb
```diff
@@ -442,11 +444,41 @@
...
+       "'expected: [155, 96, 196, 174], actual: [155  96 196 174]'"
+      ]
+     },
+     "execution_count": 25,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "'expected: {}, actual: {}'.format(bb_voc, bb_hw(bb_fastai))"
```
jph00: Thanks! This would be better done with a python 3.6 f-string for consistency with the rest of the lessons.

Issue #219, opened 2018-03-16
Implement "Averaging Weights Leads to Wider Optima and Better Generalization"


wdhorton:
I saw the Twitter thread on this arxiv paper today (Twitter: https://twitter.com/andrewgwils/status/974087637828071429, arxiv: https://arxiv.org/abs/1803.05407), and it was mentioned that it might be useful to include in the fastai lib. So I wanted to ask two questions: 1) @jph00 is this something that you'd be interested in including in the library, and 2) could I take a crack at making a making a PR with an implementation?

jph00:
Absolutely yes! :) Be sure to do plenty of testing to ensure it works as
well as it does in the paper on the same datasets, where possible.

PR #276, opened 2018-04-01
Add Stochastic Weight Averaging

wdhorton:
After filing #219, I've been working on implementing and testing Stochastic Weight Averaging, as described in the paper [Averaging Weights Leads to Wider Optima and Better Generalization](https://arxiv.org/abs/1803.05407)

@jph00 mentioned that I should do testing to ensure that I could achieve the same results seen in the paper, so I reproduced the experiments that the paper carried out on CIFAR-10 using VGG16 and PreResNet110 (reporting the mean and standard deviation of three runs for each experimental treatment). It was interesting to see that my results were all consistently higher than the original paper's, but I did observe the same improvements when comparing SWA with SGD.

Test accuracy (%) of SGD and SWA on CIFAR-10 for different training budgets.

Original paper:

| DNN (Budget)              |  SGD         | SWA 1 Budget | SWA 1.25 Budgets | SWA 1.5 Budgets |
| ------------------------- |:------------:|:------------:|:----------------:|:---------------:|
| VGG16 (200)               | 93.25 ± 0.16 | 93.59 ± 0.16 | 93.70 ± 0.22     | 93.64 ± 0.18    |
| PreResNet110 (150)        | 95.28 ± 0.10 | 95.56 ± 0.11 | 95.77 ± 0.04     | 95.83 ± 0.03    |

My results:

| DNN (Budget)              |  SGD         | SWA 1 Budget | SWA 1.25 Budgets | SWA 1.5 Budgets |
| ------------------------- |:------------:|:------------:|:----------------:|:---------------:|
| VGG16 (200)               | 93.42 ± 0.13 | 93.60 ± 0.14 | in progress    | 93.72 ± 0.08    |
| PreResNet110 (150)        | 95.53 ± 0.01 | 95.73 ± 0.14 | 96.04 ± 0.02     | 96.15 ± 0.06    |

I also wrote a notebook of functional tests to ensure that my code was doing what I intended, and to make sure it didn't cause errors when interacting with the rest of the existing library code.

The setup for the experiments and functional tests can be found in this gist: https://gist.github.com/wdhorton/447f9aecc209ee6fc6ddab3122f6b685

Edit: updated with VGG16 results as well


fastai/learner.py
```diff
@@ -167,9 +187,17 @@ def fit_gen(self, model, data, layer_opt, n_cycle, cycle_len=None, cycle_mult=1,
 
         if best_save_name is not None:
             callbacks+=[SaveBestModel(self, layer_opt, metrics, best_save_name)]
+            
+        if use_swa:
+            # make a copy of the model to track average weights
+            self.swa_model = copy.deepcopy(model)
```
jph00: I'm not sure about using `deepcopy`, in terms of either correctness or efficiency.  Would `state_dict()` be better? Or grab the params from layer_opt?

fastai/swa.py
```diff
@@ -0,0 +1,75 @@
...
+    
+def collect_bn_modules(module, bn_modules):
+    if isinstance(module, torch.nn.modules.batchnorm._BatchNorm):
+        bn_modules.append(module)
+
+def fix_batchnorm(swa_model, train_dl):
+    bn_modules = []
+    swa_model.apply(lambda module: collect_bn_modules(module, bn_modules))
+    
+    if not bn_modules:
```
jph00: This can be put on one line FYI

fastai/learner.py
jph00: I'm not sure about using `deepcopy`, in terms of either correctness or efficiency. Would `state_dict()` be better? Or grab the params from layer_opt?

fastai/swa.py
```diff
@@ -0,0 +1,75 @@
...
+        for model_param, swa_param in zip(model_params, swa_params):
+            swa_param.data *= self.swa_n
+            swa_param.data += model_param.data
+            swa_param.data /= (self.swa_n + 1)            
+    
+def collect_bn_modules(module, bn_modules):
+    if isinstance(module, torch.nn.modules.batchnorm._BatchNorm):
+        bn_modules.append(module)
+
+def fix_batchnorm(swa_model, train_dl):
```
jph00: A comment to explain what you're fixing and why would be helpful

fastai/learner.py
wdhorton: That's a good question, and I admit I wasn't totally sure if this was the right way to do it. I found a few things from the PyTorch maintainers that made me think that `copy.deepcopy(model)` was the suggested way to copy a PyTorch model:
* [this issue by soumith](https://github.com/pytorch/pytorch/issues/415)
* [this commit by apaszke](https://github.com/pytorch/pytorch/pull/454/commits/9c07f61b021b179e1974f9130cbd02848ef5a345)
* [this forum post](https://discuss.pytorch.org/t/are-there-any-recommended-methods-to-clone-a-model/483)

So then the question is, do we really need a copy of the model? I think it would be possible to perform the algorithm by keeping the average as a list of parameters, rather than an entire model (basically in [this line](https://github.com/fastai/fastai/pull/276/files#diff-6b218dc5c075483a5c9b7a5085e4570dR35) I could have `swa_params` be a list rather than getting them from `swa_model` each time). 

But as far as I've been able to figure it out, whenever we want to do inference, we still need a copy of the model to load that list of parameters into. And if we'd have to load those parameters in every time we want to calculate the validation score, then it seems to make more sense to just have a whole copy of the model from the start.

The authors' implementation similarly has two models around during training (the model being trained and the swa_model keeping track of the averages), but they avoid having to do `deepcopy` because they take in the model constructor function and a dict of kwargs to call it with, so they can just instantiate two models. Because `learn.model` is an already-instantiated model instance, we can't do that in this case.

On the efficiency question, I did `%timeit` for `deepcopy` on `preresnet110` and got: `32.1 ms ± 541 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)`. I don't think that's too bad compared to training time.

wdhorton:
I'll run some tests on the language model and NLP classifier stuff. I think if it doesn't work, we can document it like you suggested, and I can try and dig into it when I have more time later.

wdhorton:
@jph00 Finally had some time to do initial testing on the language model and NLP classifier stuff, got some out of memory issues that I'll have to dig into further (might just have to do with the size of that model and maintaining two copies of it). I think I should document that it's not quite working yet, and I can keep digging into it after this is merged—or maybe other students will be interested in investigating too once it's in the lib

jph00:
Yes that sounds great. Also got some conflicts now. If you're able to push a PR with those two little fixes we can try to use this for imagenet! :)

wdhorton:
Cool! I added a warning in the docstring about using it outside of image classification for now. Also fixed up the conflicts.

PR #284, opened 2018-04-04
Add 'device' argument to several method call to fix error without CUDA.

ForeverZyh:
Related pull: https://github.com/fastai/fastai/pull/142

### DIFF
--- a/courses/dl1/lang_model-arxiv.ipynb
+++ b/courses/dl1/lang_model-arxiv.ipynb
@@ -663,7 +663,7 @@
    "outputs": [],
    "source": [
     "def proc_str(s): return TEXT.preprocess(TEXT.tokenize(s))\n",
-    "def num_str(s): return TEXT.numericalize([proc_str(s)])"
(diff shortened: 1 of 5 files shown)


apiltamang:
Just curious, can you please document why these changes were necessary? If not in the .ipynb files, I see at least one .py file where the change was made.

A better method would be to create a simple utils method, and call that method from all those places. Then you could give that method a sensible name, and a good doc-string, so that everyone knows what your intentions are.

e.g.
```
def getDeviceStats():
''' this method is needed because of this and that, and so on .... '''
  return None if torch.cuda.is_available() else -1
```

More tip:
'None' and '-1' both appear to relate to the same thing, i.e. something been amiss. Would it make sense to return 0 and -1, or None and +1? This may not be possible, but just a thought.

Thanks ..

jph00:
Yup agree with the previous comments - especially since this is going to be seen by students, we should keep it simple. If possible, I'd like to make it even simpler - e.g. could we create an use an inherited class that does this automatically?
Also, you should check `fastai.core.USE_GPU` as well, since the user may have a GPU available but have chosen not to use it with this flag.

PR #285, opened 2018-04-04
lightweight terraforming of fastai.core, introduce `listy` idea

mcskinner:
I tend to do lightweight terraforming of code as a read it to fit with any suggested styles, as part of helping myself understand it.

I'm curious the extent to which this habit is useful. If it is, then I'm happy to continue firing off reviews like this. If not, I'm also fine keeping it to myself.

Also, this review contains a few small ideas instead of just one, as part of asking the terraforming question. Feel free to line item veto or ask for separate reviews.

### DIFF
--- a/fastai/core.py
+++ b/fastai/core.py
@@ -5,7 +5,7 @@ def sum_geom(a,r,n): return a*n if r==1 else math.ceil(a*(1-r**n)/(1-r))
 
 def is_listy(x): return isinstance(x, (list,tuple))
 def is_iter(x): return isinstance(x, collections.Iterable)
-def map_over(x, f): return [f(o) for o in x] if is_iter(x) else f(x)
+def map_over(x, f): return [f(o) for o in x] if is_listy(x) else f(x)
 
 conv_dict = {np.dtype('int8'): torch.LongTensor, np.dtype('int16'): torch.LongTensor,
     np.dtype('int32'): torch.LongTensor, np.dtype('int64'): torch.LongTensor,
--- a/fastai/core.py
+++ b/fastai/core.py
@@ -31,8 +31,8 @@ def create_variable(x, volatile, requires_grad=False):
 
 def V_(x, requires_grad=False, volatile=False): return create_variable(x, volatile, requires_grad)
 def VV_(x):                                     return create_variable(x, True)
-def V(x, requires_grad=False, volatile=False): return sapply_(x, lambda o: V_(o, requires_grad, volatile))
-def VV(x):                                     return sapply_(x, VV_)
+def V(x, requires_grad=False, volatile=False): return map_over(x, lambda o: V_(o, requires_grad, volatile))
+def VV(x):                                     return map_over(x, VV_)
 
 def to_np(v):
     if isinstance(v, (np.ndarray, np.generic)): return v
--- a/fastai/core.py
+++ b/fastai/core.py
@@ -3,6 +3,9 @@ from .torch_imports import *
 
 def sum_geom(a,r,n): return a*n if r==1 else math.ceil(a*(1-r**n)/(1-r))
 
+def listy(x): return isinstance(x, (list,tuple))
+def sapply_(x, f): return [f(o) for o in x] if listy(x) else f(x)
+
 conv_dict = {np.dtype('int8'): torch.LongTensor, np.dtype('int16'): torch.LongTensor,
     np.dtype('int32'): torch.LongTensor, np.dtype('int64'): torch.LongTensor,
     np.dtype('float32'): torch.FloatTensor, np.dtype('float64'): torch.FloatTensor}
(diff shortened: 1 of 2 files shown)


apiltamang:
I obviously see that you're trying to do the right thing, but to what gains.. Injecting changes as such, and as trivial as they seem, might introduce subtle bugs which may cause a lot of confusion.

While in retrospect, I agree that such refactoring definitely can also lead to catching, otherwise hidden,  bugs.

For your intent, I strongly suggest including some unit tests. There oughta be no refactoring for the pure reason of only refactoring (or terraforming) without some unit and integration tests. While machine-learning is not exactly amenable to traditional testing techniques, code-modifications as above seem like they could use some unit tests. Also, it would set a very good template for including more unit tests.

I hope you take the suggestion for good! I've met only very few developers who don't balk when asked to write test. Ultimately, I don't have write access to this repo, so it's really @jph00's call .

mcskinner:
The goal, which is good to question, is to bring the style of the code more consistently in line with the conventions in this project's [style guide](https://github.com/fastai/fastai/blob/master/docs/style.md). It happens to jive with my personal philosophy that, because code is read far more often than written, any investment in readability improvements tends to have high ROI. This is related to Chris Olah and Shan Carter's discussion of [Research Debt](https://distill.pub/2017/research-debt/) and the related concept of Interpretive Labor.

In this case I'm trying to keep those changes very small and visually verifiable. That seemed to be the norm for this project rather than tests, so it's what I did. I'm sure there is a line at which the changes become large enough in scope to warrant an automated check, and that might have happened here.

As you said, we can let @jph00 decide where that line is :) I don't mind writing tests, and in fact find it quite therapeutic, but if that's asked I wouldn't mind it if someone else set up the scaffolding first ;)

PiotrCzapla:
@mcskinner, I'm thinking about writing some test to flash out the API and how it is used so that I can propose such small changes to the code without the need to rerun everything manually. I've created an issue to discuss that:
https://github.com/fastai/fastai/issues/286

fastai/core.py
```diff
@@ -3,37 +3,35 @@
 
 def sum_geom(a,r,n): return a*n if r==1 else math.ceil(a*(1-r**n)/(1-r))
 
+def listy(x): return isinstance(x, (list,tuple))
```
jph00: Probably should be `is_listy`.

fastai/core.py
```diff
@@ -3,37 +3,35 @@
...
+def sapply_(x, f): return [f(o) for o in x] if listy(x) else f(x)
```
jph00: That can be more generally useful, so maybe remove the trailing underscore. But try to come up with a name that's more understandable to python users, since sapply is an R thing. :)

jph00: Also the conditional here should be `if isinstance(e, collections.Iterable)`, since you're not using the objects indexer so anything iterable will work.

jph00:
Yes agree with all the comments - little readability improvements are appreciated, but without tests it is a bit concerning to merge them. These particular changes all look fine, but we should probably invest in tests before doing much more refactoring or readability improvements.

fastai/core.py
mcskinner: For naming ideas, this feels a lot like the FP notion of `map` e.g. as in Haskell. That's a reserved keyword in Python though, and the Java'y naming of `transform` is overloaded with image preprocessing. So perhaps `map_over` or `apply`? The latter is kind of generic, so I picked `map_over` for this volley.

Done on the `isinstance` change, extracted to a function used here. Seems like something for `collections.Mapping` might be nice as well.

mcskinner:
Add me to the list of folks who appreciate a nice testing safety blanket. Let me know what I can do to help!

fastai/core.py
```diff
@@ -3,37 +3,36 @@
 
 def sum_geom(a,r,n): return a*n if r==1 else math.ceil(a*(1-r**n)/(1-r))
 
+def is_listy(x): return isinstance(x, (list,tuple))
+def is_iter(x): return isinstance(x, collections.Iterable)
+def map_over(x, f): return [f(o) for o in x] if is_listy(x) else f(x)
```
mcskinner: @jph00 I rolled this back to using `is_listy` because `collections.Iterable` also ends up catching `np.ndarray`, `torch.Tensor`, and `torch.autograd.Variable` in the net.

fastai/core.py
```diff
@@ -3,37 +3,36 @@
...
-def V(x, requires_grad=False, volatile=False):
-    return [V_(o, requires_grad=requires_grad, volatile=volatile)
-            for o in x] if isinstance(x,(list,tuple)) else V_(x, requires_grad=requires_grad, volatile=volatile)
-
-def VV_(x): return create_variable(x, True)
-def VV(x):  return [VV_(o) for o in x] if isinstance(x,list) else VV_(x)
+def V_(x, requires_grad=False, volatile=False): return create_variable(x, volatile, requires_grad)
+def VV_(x):                                     return create_variable(x, True)
+def V(x, requires_grad=False, volatile=False): return map_over(x, lambda o: V_(o, requires_grad, volatile))
+def VV(x):                                     return map_over(x, VV_)
```
mcskinner: This is a slight expansion to include `tuple` as well as `list` (which is all there was before).

Seems harmless, but I'm willing to revert.

fastai/core.py
PiotrCzapla: @jph00 If I get @mcskinner the whole point of this function is to apply a `f` on a pythonic structure of objects including `tensors` ,  `ndarrays`, `DataSets` (which are `Iterable`)

This seems to be very useful especially if we allow for nesting of structures, imagine:
```py
>>> ds = trn_ds, val_ds, tst_ds
>>> sapply(ds, DataLoader)    # -> (trn_dl, val_dl, tst_dl)

>>> sapply( (x, [mask, borders, weights]), crop)  #-> (corp(x), [crop(mask), crop(borders), crop(weights)])
```

Then the name `sapply` seems to  reflect the idea well as it can be read as "structure apply". Which explains why the function don't work on any`Iterable` and only on list tuples nad dicts and why `sapply(obj, f)`  is  `f(obj)` and not `map(obj, f)` (ie. because obj is not embedded in  a pythonic structure)

Such function can simplify  `get_tensor` from data loaders for example:
https://github.com/fastai/fastai/blob/master/fastai/dataloader.py#L10-L20

If we make sapply as follows:
```py
def sapply(x, f):
    if isinstance(x, collections.Mapping):
        return {k: sapply(sample, f) for k, sample in x.items()}
    elif isinstance(x, collections.List):
        return [sapply(sample, f) for sample in x]
    elif isinstance(x, collections.Tuple):
        return tuple(sapply(sample, f) for sample in x)
    else:
        return f(x)
```    
    
We can write:
```py
def pin_memory(batch):
    if isinstance(batch, (np.ndarray, np.generic)):
        batch = T(batch).contiguous()
        return batch.pin_memory()
    if isinstance(x, string_classes): return bach
    raise TypeError("Batch must contain be in structure of dicts, lists or tuples; found {}"
                     .format(type(batch)))

def get_tensor(batch, pin): return sapply(batch, pin_memory) if pin else batch
```

fastai/core.py
PiotrCzapla: @mcskinner  see the comments above, I think `sapply` represents better your idea than `map_over`, especially if we extend it to work with dicts and any depths of nested structures.

fastai/core.py
PiotrCzapla: I've just checked that `sapply` in `R` does not behave in the way described above, so it will be confusing for some. Maybe `struct_apply` then? It can be confused with https://docs.python.org/3/library/struct.html, but hopefully noone will mix DeepLearning with writing low level code to construct arbitrary `C` structures.

PR #289, opened 2018-04-04
Document 'fastai.models'

poppingtonic:
Based on [this question](http://forums.fast.ai/t/how-to-use-fastai-library-with-a-custom-pretrained-model/14439/4), I thought I'd document, at a high level, the `fastai/models` folder.


docs/anatomy.adoc
```diff
@@ -12,6 +12,7 @@ A substantial amount of the code in the fastai library is dedicated to reading,
...
 * link:../fastai/structured.py[fastai.structured] has methods that can massage `pd.DataFrame`s into friendlier forms for learning. This includes dealing with missing values, ensuring correct data types, normalizing numerical inputs and one-hot encoding categories, and some lightweight feature engineering on dates.
 * link:../fastai/text.py[fastai.text] has similarly intended utilities for reading and tokenizing text data.
+* link:../fastai/models[fastai.models] contains various tried and tested implementations of deep learning architectures that have been published out there. To test and attempt performance replication of a published model's weights, modules defined here can be combined with pre-trained weights by editing link:../torch_imports.py[torch_imports] to load the model and link to the published article(s) or paper(s) that describe(s) the architecture.
```
jph00: Replication isn't really a key reason for any of this. The key reason is so that we can use transfer learning to solve problems.

mcskinner: Nice to see this doc growing already :)

Should we start a new header section for modeling stuff, vs data and training? This would certainly not be the only thing there.

jph00: That might make sense...

jph00:
Closing this since new doc framework and docs are done now (see http://docs.fast.ai).

PR #297, opened 2018-04-06
Added some abbreviations

duckysherwood:
Added a few abbreviations which I came across.  This is by no means all the abbreviations it's just a start.


jph00:
I've merged these - but some of them only appear in the notebooks; I'm not sure either way if they should be in this doc, or maybe would be better to define them in the notebook where they appear?

duckysherwood:
Hmm.  I saw them in the notebooks and didn't comb through the libraries to see if they were also there.  Fundamentally, it comes down to the question of whether the notebooks are part of the source code or not.  I can see both sides of that.

I would be in favour of taking an inclusivist viewpoint.  If the abbreviation appears in the notebook and is for a word which is likely to be used in the future (like `wgt`), then having it in the abbreviation list lets a future contributor know what term they should be using.  It would cause extra work or confusion if contributor A used 'w' and contributor B used 'wt' and contributor C used 'wgt'.

PR #302, opened 2018-04-07
initial testing setup for LayerOptimizer

mcskinner:
I went with the builtin `unittest` framework, but I'm open to alternatives.

Run this guy like so:
```
(fastai) .../fastai/fastai$ python -m unittest tests.test_layer_optimizer
..........
----------------------------------------------------------------------
Ran 10 tests in 0.003s

OK
```

Of course the requirement to pinpoint specify the test is not ideal. I don't have a ton of expertise on how to resolve that, and am open to the idea of some testing framework alternatives to make the whole thing nicer.

It is also worth noting that two tests are verifying probably incorrect behavior (I commented them). I'm happy to fix that in a subsequent review, but wanted to separate it from the test.

### DIFF
--- a/fastai/torch_imports.py
+++ b/fastai/torch_imports.py
@@ -21,7 +21,7 @@ from .models.nasnet import nasnetalarge
 import warnings
 warnings.filterwarnings('ignore', message='Implicit dimension choice', category=UserWarning)
 
-def children(m): return m if isinstance(m, (list, tuple)) else list(m.children())
+def children(m): return m if is_listy(m) else list(m.children())
 def save_model(m, p): torch.save(m.state_dict(), p)
 def load_model(m, p): m.load_state_dict(torch.load(p, map_location=lambda storage, loc: storage))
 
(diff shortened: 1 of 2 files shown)


jph00:
Thanks! Fixing those 2 missing assertion issues would be great. BTW I found some of the tests a little unclear as to what you were testing and why. A comment or docstring on the test would be nice to explain tests where it's not pretty obvious from the code what's happening.

I think we should use pytest since it's more modern, concise, and [recommended by coders](https://www.slant.co/topics/2621/~python-unit-testing-frameworks).

We should also follow this suggestion from the [python testing guide](http://docs.python-guide.org/en/latest/writing/tests/):

> Use long and descriptive names for testing functions. The style guide here is slightly different than that of running code, where short names are often preferred. The reason is testing functions are never called explicitly. square() or even sqr() is ok in running code, but in testing code you would have names such as test_square_of_number_2(), test_square_negative_number(). These function names are displayed when a test fails, and should be as descriptive as possible.

I'll merge this PR now so that we have something to build on.

jph00:
Oh also it would be nice to start an asciidoc file in the docs/ dir where we can start describing fastai testing approaches/recommendations/etc.

PR #303, opened 2018-04-07
Added cutout transform

binga:
Implemented the cutout transform from this repository: https://github.com/uoguelph-mlrg/Cutout/

Will use this on a few datasets and show the results soon.

Input:
image

Output:
image

### DIFF
--- a/fastai/transforms.py
+++ b/fastai/transforms.py
@@ -62,7 +62,7 @@ def center_crop(im, min_sz=None):
     return crop(im, start_r, start_c, min_sz)
 
 def cutout(im, n_holes, length):
-    h, w, _ = im.shape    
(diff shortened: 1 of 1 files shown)


binga:
Ok, now I'm not sure about the `h, w, _ = im.shape`. Do we have numpy arrays or PyTorch tensors here? :|

jph00:
I'll merge this now so folks can play with it - but I think you should revert the shape bug fix. It seems you added your transform right to the end of the list, after the channel axis rotation transform, but that's a really unusual place to put a transform. Normally we put them earlier in the pipeline.

And don't forget, if you have a question about what's happening in the code, just add a breakpoint and take a look in the debugger! :)

PR #305, opened 2018-04-07
Add multi-label classification to columnar model

ronlut:
1 thing worth mentioning:
I extracted `is_multi` property from `ImageClassifierData` to the base `ModelData` class as now both `ColumnarModelData` and `ImageClassifierData` need this property.

### DIFF
--- a/fastai/column_data.py
+++ b/fastai/column_data.py
@@ -5,22 +5,23 @@ from .learner import *
 
 
 class PassthruDataset(Dataset):
-    def __init__(self,*args, is_reg=True):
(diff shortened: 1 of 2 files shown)


jph00:
Great thanks! One thing to think about if you're interested - can we refactor this is_reg and is_multi stuff in all the places it's used so we don't have to have it as a param to so many functions and can simplify the code a bit?

ronlut:
Will think about it. 
Currently working on making `StructuredLearner.summary()` work :)

PR #308, opened 2018-04-07
correctness assertions for LayerOptimizer `set_lrs` and `set_wds`

mcskinner:


jph00:
Thanks for this.

Just an FYI: your earlier `is_listy` change to torch_imports didn't work since fastai.core isn't imported there - since we don't as yet have the protection of unit tests or CI, please double-check that something like the lesson 1 notebook works before sending in a PR (or create a simple little integration test script to check).

mcskinner:
Roger that, sorry for the mess. I'll go ahead and add a script and maybe we can consider it an interim best practice in the testing doc? (PR out for that now).

PR #309, opened 2018-04-07
start documenting testing style and patterns

mcskinner:
The first section is culled from the discussion on #302, the second section is a few example points from my own philosophy that I'm using to bootstrap the discussion. As usual, feel free to line item veto.

### DIFF
--- a/docs/testing.adoc
+++ b/docs/testing.adoc
@@ -12,7 +12,7 @@ More generally, aim to write tests that also explain the code they are testing. 
 
 == Testing patterns
 
-* Mocking or faking objects is fine for getting started, but real versions of the required objects are preferred when they can be used without prohibitive overhead.
+* Do not use mock or fake objects. The library is nice enough that real versions of required objects can be used without prohibitive overhead.
 * Keep test methods small and tidy, just like any other code.
 * Aim to add a regression test as part of any bug fix PR.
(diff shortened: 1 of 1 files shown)


docs/testing.adoc
```diff
@@ -0,0 +1,18 @@
...
+
+____
+Use long and descriptive names for testing functions. The style guide here is slightly different than that of running code, where short names are often preferred. The reason is testing functions are never called explicitly. square() or even sqr() is ok in running code, but in testing code you would have names such as test_square_of_number_2(), test_square_negative_number(). These function names are displayed when a test fails, and should be as descriptive as possible.
+____
+
+More generally, aim to write tests that also explain the code they are testing. A really good test suite can also serve as really good documentation.
+
+== Testing patterns
+
+* Mocking or faking objects is fine for getting started, but real versions of the required objects are preferred when they can be used without prohibitive overhead.
```
jph00: I deeply deeply dislike faking or mocking objects and have managed to get this far in my life without ever needing to do this - that includes testing every part of a production email system written from scratch!

mcskinner:
The from scratch stuff isn't too bad, it's legacy and external code where mocks can save you oodles of effort tracing through byzantine construction dependencies.

That said, fastai doesn't have that problem, and tests can provide a nice pressure to keep it that way.

Updated the point to just say no.

PR #311, opened 2018-04-07
Add ObjectDetectionData

radekosmulski:
My intent was to create a data class that could replace the trick that allows us to use ImageClassifierData for detection.

Things were going relatively okay until I encountered transforms. I have taken only a cursory glance at how they work however I am not sure what would be a nice way of getting them to work with this.

Also, I am not really sure this is the architectural direction we want to take. Might be I have dabbled in something that doesn't make sense for me to look into at this point. Might be this effort is completely unneeded and the only feasible way of getting a class for data for detection would be via a bigger refactoring that would include transforms.

Happy for this to be closed, only thought I'd share this in case there was value in working on this further. Don't suspect that this is the case but if it is then happy to continue with the effort but would require guidance how to proceed.

### DIFF
--- a/fastai/dataset.py
+++ b/fastai/dataset.py
@@ -1,4 +1,5 @@
 import csv
+from functools import lru_cache
 
 from .imports import *
 from .torch_imports import *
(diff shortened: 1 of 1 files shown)


jph00:
Frankly I'm not sure - I think the best thing to do would be to pop this all into a notebook along with examples showing how to use, and ask for feedback/help on the forum. I'll close it for now until we've had a chance to get more eyeballs on it. It's not a bad idea actually for more major changes to get shared as a gist on the forum first.

Issue #245, opened 2018-03-24
Problems with Random transformation not respecting Y. Especially in context of Y being a Mask


PiotrCzapla:
I had huge issues getting the transformation to work with for current Data bowl competition. First it had the concurrency issues, now once this is fixed I've noticed that there are more issues. I will list them one by one in this thread and follow with a pull request that solves the issues.

✅  `RandomCrop` -  works fine
<img width="814" alt="screen shot 2018-03-24 at 18 52 15" src="https://user-images.githubusercontent.com/340180/37867145-7fbe4b5a-2f94-11e8-851f-0415bd32d4d9.png"> 

❌  `RandomRotate(10)` - does not repect masks see image below
<img width="931" alt="screen shot 2018-03-24 at 18 43 45" src="https://user-images.githubusercontent.com/340180/37867056-6e53e038-2f93-11e8-931d-6fa203e7714d.png">

❌ `RandomLighting(0.05, 0.05)` - changes light of a mask

❌ `RandomDihedral` - does not respect masks
<img width="873" alt="screen shot 2018-03-24 at 18 50 17" src="https://user-images.githubusercontent.com/340180/37867112-3722dd16-2f94-11e8-9b2b-ce19a751b5f2.png">

❌  `RandomFlip` -same issue
<img width="903" alt="screen shot 2018-03-24 at 18 51 03" src="https://user-images.githubusercontent.com/340180/37867123-5daa1666-2f94-11e8-90c5-7acb0e166a65.png">

❌  Normalize - tries to normalize binary Masks

PiotrCzapla:
I think I see where the issue is. The `tfm_y` is not being set along when the  `tfms_from_stats` is called with `tfm_y=TfmType.MASK`. (see screenshot), which makes it impossible to use `aug_tfms=transforms_top_down` along with `tfm_y`. 

The issue isn't that easy to correct, as assign tfm_y to all aug_tfms will cause problems later as most of the time aug_tfms are global objects, that are reused between calls. But It could be a good temorary fix
<img width="908" alt="screen shot 2018-03-24 at 18 58 02" src="https://user-images.githubusercontent.com/340180/37867294-4d946fae-2f96-11e8-9af5-8e3ded1771c0.png">

jph00:
I've just pushed some changes showing how to do this. See the transforms etc in this notebook: https://github.com/fastai/fastai/blob/master/courses/dl2/carvana.ipynb

jph00:
[...]

PR #312, opened 2018-04-07
Let Transforms handle complex structures and fix the tfm_y issue #245

PiotrCzapla:
To train a mask rcnn using Fastai we need to return  Mask and Bounding
boxes for each sample, this currently cannot be achieved using fastai.

The proposed changes:
- Fix the above issue, as it gives user ability to define arbitrary
complex structures in x and y.

- Allow users to use transform_on_side even if  tfm_y is PXIEL, MASK or
COORD. (#245)

- Moreover, the introduction of .determ() will let us remove the Thread
local storage from set_state in a long run, probably speeding up the
transformations, and making them less dependant on side effects.

The propose change works as follows:

We remove the assumption that we have only x and y returned from a dataset.
Instead, we allow the user to specify a custom function that is responsible
for applying a deterministic transformation to the structure of x and
y.

See TestTransforms.test_it_is_possible_to_transform_mask_n_coords for a
concrete example.

### DIFF
--- a/courses/dl2/cifar10-dawn.ipynb
+++ b/courses/dl2/cifar10-dawn.ipynb
@@ -1,0 +1,199 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## CIFAR 10"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "%matplotlib inline\n",
+    "%reload_ext autoreload\n",
+    "%autoreload 2"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from fastai.conv_learner import *\n",
+    "from fastai.models.cifar10.wideresnet import wrn_22\n",
+    "torch.backends.cudnn.benchmark = True\n",
+    "PATH = Path(\"data/cifar10/\")\n",
+    "os.makedirs(PATH,exist_ok=True)"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 3,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "classes = ('plane', 'car', 'bird', 'cat', 'deer', 'dog', 'frog', 'horse', 'ship', 'truck')\n",
+    "stats = (np.array([ 0.4914 ,  0.48216,  0.44653]), np.array([ 0.24703,  0.24349,  0.26159]))\n",
+    "\n",
+    "bs=512\n",
+    "sz=32"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 4,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "tfms = tfms_from_stats(stats, sz, aug_tfms=[RandomCrop(sz), RandomFlip()], pad=sz//8)\n",
+    "data = ImageClassifierData.from_paths(PATH, val_name='test', tfms=tfms, bs=bs)"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 5,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "m = wrn_22()"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 6,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "learn = ConvLearner.from_model_data(m, data)\n",
+    "learn.crit = nn.CrossEntropyLoss()\n",
+    "learn.metrics = [accuracy]\n",
+    "wd=1e-4\n",
+    "lr=1.5"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 7,
+   "metadata": {},
+   "outputs": [
+    {
+     "data": {
+      "application/vnd.jupyter.widget-view+json": {
+       "model_id": "0f6fa780bc0b4b1e978fb7572e718ce0",
+       "version_major": 2,
+       "version_minor": 0
+      },
+      "text/plain": [
+       "HBox(children=(IntProgress(value=0, description='Epoch', max=30), HTML(value='')))"
+      ]
+     },
+     "metadata": {},
+     "output_type": "display_data"
+    },
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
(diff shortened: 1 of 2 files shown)


PiotrCzapla:
@jph00 The PR is not intended to be merged yet. I would add more tests to make sure that the changes are indeed working as they should, but I'd like to hear your opinion before invest time in writing the tests.

To make the code readable I've added only the absolutely necessary changes and I left the now redundant code in place so that the commit is easier to read.

Writing the test for the transform_coord (awesome idea btw.) I've found an issue there, it seems that each time a transformation is applied a bounding box is shrunk by one pixel. So if you have 7 transformations your bounding box will be 7 pixels short. I will fix that once I have the green light to go with this design.

Below are few ideas that I haven't implemented for simplicity, and I would love to hear your opinion.

- Move `apply_transform` to the `Dataset` as it seems a more natural place for such function.
- Remove tfm_y in favor of tfmtype passed to each transformation
- Remove is_y completely and we can replace it with tfm.INPUT_PIXEL
- Remove set_state & store completely in favor of creating a copy of the transformation either with fixed random state. 
- Add ability to handle dictionaries (in np_collate)
- Add ability to handle namedtuples (in np_collate) so that batch of namedtuples would become a namedtuple (that is for future)

PiotrCzapla:
Ohh, one more thing, the idea of creating new "transformation" once `determ()` is called will let us implement TTA for mask & bounding box problems. Because the list of deterministic transformations  created in Transforms.determ can be saved and then reused to undo the transformation on a result from the model.

PiotrCzapla:
Here is an example how Transformations could look like in future:


```py
class RandomRotate(CoordTransform):
    def __init__(self, deg, p=0.75, mode=cv2.BORDER_REFLECT):
        super().__init__()
        self.deg,self.p = deg,p

    def new_state(self):
        return dict(rdeg = rand0(self.deg),
                    rp = random.random() < self.p)

    def do(self, x, tfmtype, is_y, rdeg, rp):
        interpolation = cv2.INTER_NEAREST if is_y else cv2.INTER_AREA
        mode = cv2.BORDER_CONSTANT if tfmtype in (TfmType.COORD, TfmType.CLASS) else self.mode
        if rp: x = rotate_cv(x, rdeg, mode=mode, interpolation=interpolation)
        return x

    def undo(self, *kwargs): pass

    ### default implementation of determ that would be moved to the based type
    def determ(self):
        rdeg = rand0(self.deg)
        rp = random.random() < self.p
        return CoordAction(self, self.new_state())

class CenterCrop(CoordTransform):
    def __init__(self, sz, sz_y=None):
        super().__init__()
        self.min_sz,self.sz_y = sz,sz_y

    def do(self, x, tfmtype, is_y):
        return center_crop(x, self.sz_y if is_y else self.min_sz)

    def undo(self, x, tfmtype, is_y):
        pass
```
And the action class would look like this
```py
class CoordAction:
    def __init__(self, trans, state={}):
        self.trans,self.state = trans,state

    def __call__(self, x, tfmtype, is_y=False):
        if not is_y: self.shape = x.shape  # used in subsequent transform_coord
        if tfmtype == TfmType.COORD: return self.trans.do_coord(x)
        if tfmtype == TfmType.NO:    return x
        return self.trans.do(x, tfmtype, is_y, **self.state)

    def map_y(self, y0):
        if self.shape is None:
            raise ValueError(
                "You cannot run TfmType.COORD transformation without running first transformation of x (is_y=False)")
        y = make_square(y0, self.shape)
        y_tr = self,trans.do(y, TfmType.PIXEL, True, **self.state)
        return to_bb(y_tr, y)

    def do_coord(self, y):
        yp = partition(y, 4)
        y2 = [self.map_y(y) for y in yp]
        return np.concatenate(y2)
    def __str__(self):
        return f"CoordAction({self.trans.__name__}, {self.state}))"
```

jph00:
I don't have time to think about this deeply right now, but the direction sounds hopeful. Certainly I do want to have transforms_side_on et all work with tfm_y.

It's a bit inconvenient for me to have PRs sitting here that aren't actually ready to be merged. I also think it's nice to have some discussion more widely about significant user-visible changes. Therefore, would you be comfortable with me closing this and instead creating a gist showing the proposed new approach, and posting a thread on the forum that introduces it for discussion?

If that's not ideal for you, then feel free to say no!

PiotrCzapla:
I see your point having a long live PR is awful from the maintainers perspective. It is as bad from the contributor perspective having to deal with the upcomming changes on an active repository. 
That is why I opted for full compatibility in the first rewrite and I wanted to have it out and merge ready in matters of days. But If you don't have the time to review it in the next couple of days I understand.

I will throw the gist to the forum to discuss the API v2 (the one showed in comment). Although I know that discussion can take ages, and I want to have something that works for the Data Science Bowl before it is finshed :).

I see 2 ways to achive that:
A. Have a quick discussion and merge a fully compatible, tested version providing it is ready in 3 days, the change won't be much longer than what is proposed above. If you won't have time, I understand.
B. Create transforms2.py and put there a proptotype version of the API that isn't 100% compatible (with redundancy removed). So that contributors can think about appling the changes to the second file as well.  I'm not sure If I manage to write the unit test in a way that they can be used for both transforms2.py and transforms.py but I will try.
C. I can keep everything in my fork for time being / or in a branch, but then we won't benefit from tests.

Given that you don't have time probably B. or C. would be a good way to go. Let me know and if you select B or C, close the PR.

PiotrCzapla:
I've created the thread to discuss the new API as @jph00  requested.  If you have comments/suggestions please go there: http://forums.fast.ai/t/proposal-for-transforms-py-v2/14633

This PR will be either closed or a small backward compatible change would be merged to master, so it isn't best place to discuss the full proposal.

PiotrCzapla:
Update. Small change of plans as it seems that Datascience bowl deadline is in 2 days not in 7 as I initially though so I will first focus on finishing some ideas before I can jump on this again. So to avoid having this PR hanging around  much longer I'm going for the B option, and create transforms2.py. 

I will use a bit of python magic to enable the new API reusing the existing classes from transforms.py as much as it is possible, so that we can benefit from fixes changes to the transforms.py and keep the updated API.

PiotrCzapla:
@jph00 I've extracted the relevant changes to transforms as mixins and placed them into transforms2.py. It works, tests are passing, but this is a bit too magical for my taste. 

I would rather speed up discussion regarding the new API and go for plan A ie add backwards compatible changes to transforms.py.

PiotrCzapla:
I've reverted back to the single transforms.py and rebased the branch on recent master so that the changes are clearly visible, the unet notebook is coming in a separate repository.

PiotrCzapla:
@jph00 I've extracted the test to a new PR #433, once it is merged this PR is going to be one commit.

PR #313, opened 2018-04-08
Added GoogleNetResize augmentation

supreethmanyam:
Settings for GoogleNetResize can be passed through a parameter gnet_resize_params as a dictionary to both tfms_from_model and tfms_from_stats.

Default parameters are as follows:
```gnet_resize_params={'min_area_frac':0.08, 'min_aspect_ratio':0.75, 'max_aspect_ratio':1.333, 'flip_p':0.5, 'random_state':None}```

Output from CropType.NO only:
screen shot 2018-04-08 at 5 53 42 am

Output from CropType.GOOGLENET only with default parameters:
screen shot 2018-04-08 at 5 53 57 am

### DIFF
--- a/fastai/transforms.py
+++ b/fastai/transforms.py
@@ -61,6 +61,26 @@ def center_crop(im, min_sz=None):
     start_c = math.ceil((c-min_sz)/2)
     return crop(im, start_r, start_c, min_sz)
 
+def googlenet_resize(im, targ, min_area_frac, min_aspect_ratio, max_aspect_ratio, flip):
+    h,w,*_ = im.shape
+    area = h*w
+    for _ in range(10):
+        targetArea = random.uniform(min_area_frac, 1.0) * area
+        aspectR = random.uniform(min_aspect_ratio, max_aspect_ratio)
+        ww = int(np.sqrt(targetArea * aspectR) + 0.5)
+        hh = int(np.sqrt(targetArea / aspectR) + 0.5)
+        if flip:
+            ww, hh = hh, ww
+        if hh <= h and ww <= w:
+            x1 = 0 if w == ww else random.randint(0, w - ww)
+            y1 = 0 if h == hh else random.randint(0, h - hh)
+            out = im[y1:y1 + hh, x1:x1 + ww]
+            out = cv2.resize(out, (targ, targ), interpolation=cv2.INTER_CUBIC)
+            return out
+    out = scale_min(im, targ, interpolation=cv2.INTER_CUBIC)
+    out = center_crop(out)
+    return out
+
 def scale_to(x, ratio, targ): return max(math.floor(x*ratio), targ)
 
 def crop(im, r, c, sz): return im[r:r+sz, c:c+sz]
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: fastai/transforms.py [hunk 1]
```diff
@@ -61,6 +61,26 @@ def center_crop(im, min_sz=None):
     start_c = math.ceil((c-min_sz)/2)
     return crop(im, start_r, start_c, min_sz)
 
+def googlenet_resize(im, targ, min_area_frac, min_aspect_ratio, max_aspect_ratio, flip_p):
```
jph00: Add a docstring with a 1-line summary and link to the paper

fastai/transforms.py
```diff
@@ -61,6 +61,26 @@ def center_crop(im, min_sz=None):
...
+    h,w,*_ = im.shape
+    area = h*w
+    for _ in range(10):
+        targetArea = random.uniform(min_area_frac, 1.0) * area
```
jph00: Are the some assumptions here that the original image is big enough? What happens if it's really small? What happens if it's really big? Could you show a couple of examples - maybe link from the PR to a gist that gives this transform a really good work-out?

fastai/transforms.py
```diff
@@ -61,6 +61,26 @@ def center_crop(im, min_sz=None):
...
+        aspectR = random.uniform(min_aspect_ratio, max_aspect_ratio)
+        ww = int(np.sqrt(targetArea * aspectR) + 0.5)
+        hh = int(np.sqrt(targetArea / aspectR) + 0.5)
+        if flip_p:
+            ww, hh = hh, ww
+        if hh <= h and ww <= w:
+            x1 = 0 if w == ww else random.randint(0, w - ww)
+            y1 = 0 if h == hh else random.randint(0, h - hh)
+            out = im[y1:y1 + hh, x1:x1 + ww]
+            out = cv2.resize(out, (targ, targ), interpolation=cv2.INTER_CUBIC)
```
jph00: cv2 docs recommend INTER_AREA for making smaller. But for is_y=True transforms we need INTER_NEAREST if its COORD or CLASS. There's a partially working version in other transforms such as `Scale` you can borrow from, although I don't think they handle PIXEL correctly yet (should use INTER_AREA in that case)

fastai/transforms.py
```diff
@@ -61,6 +61,26 @@ def center_crop(im, min_sz=None):
...
+            return out
+    out = scale_min(im, targ, interpolation=cv2.INTER_CUBIC)
```
jph00: See above re interpolation

fastai/transforms.py
```diff
@@ -518,6 +538,30 @@ def __init__(self, n_holes, length, tfm_y=TfmType.NO):
     def do_transform(self, img, is_y):
         return cutout(img, self.n_holes, self.length)
 
+class GoogleNetResize(CoordTransform):
+    
+    def _init(self, params=None):
+        if params:
+            for k, v in params.items():
+                if k != 'self':
```
jph00: Can you add a comment explaining what's happening here, and why?

fastai/transforms.py
```diff
@@ -518,6 +538,30 @@ def __init__(self, n_holes, length, tfm_y=TfmType.NO):
...
+                    setattr(self, k, v)
+    
+    def __init__(self, targ_sz, 
+                 gnet_resize_params={'min_area_frac':0.08, 'min_aspect_ratio':0.75, 'max_aspect_ratio':1.333, 'flip_p':0.5, 'random_state':None},
```
jph00: Just make these regular args, not a dict

fastai/transforms.py
```diff
@@ -533,20 +577,27 @@ class CropType(IntEnum):
...
     NO = 3
+    GOOGLENET = 4
 
-crop_fn_lu = {CropType.RANDOM: RandomCrop, CropType.CENTER: CenterCrop, CropType.NO: NoCrop}
+crop_fn_lu = {CropType.RANDOM: RandomCrop, CropType.CENTER: CenterCrop, CropType.NO: NoCrop, CropType.GOOGLENET: GoogleNetResize}
 
 class Transforms():
-    def __init__(self, sz, tfms, normalizer, denorm, crop_type=CropType.CENTER, tfm_y=TfmType.NO, sz_y=None):
+    def __init__(self, sz, tfms, normalizer, denorm, crop_type=CropType.CENTER,
+                 gnet_resize_params={'min_area_frac':0.08, 'min_aspect_ratio':0.75, 'max_aspect_ratio':1.333, 'flip_p':0.5, 'random_state':None},
```
jph00: Remove gnet_resize_params from everywhere you've included them - if the user wants different params, they can just create the transform list manually. These shortcut functions are really just designed for convenience when using defaults.

fastai/transforms.py
supreethmanyam: I have used cifar-10 images for small size images, dogs vs cats for medium size images and downloaded few images of around 1920*1080 images from the web for large size images.
Link for gist: https://gist.github.com/supreethmanyam/11d4daadd799c6a1a3234745f65430b6

fastai/transforms.py
supreethmanyam: _init method was used to convert each element in the params dict to attribute of GoogleNetResize class without having to explicitly write for every argument. I have removed this as the parameters are now regular arguments.

jph00:
Great job!

supreethmanyam:
Thank you Jeremy :)

I have added Test_GoogleNetResize.ipynb on the fastai repo. I have sent another PR to remove it and bring the repo to ideal state. Sorry for the mistake.

PR #314, opened 2018-04-08
Corrected docstring for RNN_Encoder.forward to reflect correct tensor shape

hamelsmu:
The input for the method `forward` is actually (sl, bs) not (bs, sl).  

I was trying to follow the code, and this confused me until I realized this was just a mistake in the docstring.  Thanks


jph00:
I could see how that could have been confusing!

PR #318, opened 2018-04-08
Plots + minor fixes

sgugger:
(had forgotten to remove my test files in the previous one + I'll post examples on how to use these functions when/if you accept the PR)

Plots:
- introduces a lr_find2 that computes the validation loss + metrics after each batch in the training loop for the enxt batch of the validation set.
- validation losses and metrics are saved each time they are computed (whether in normal training or LR find) so we can plot them after if we want.

Fixes:
- use_clr_beta changed the momentum only in the last group of layers
- use_clr_beta changes beta1 instead of momentum if Adam is used
- can pass all the arguments that goes to fit in lr_find or lr_find2


fastai/model.py
```diff
@@ -87,7 +87,7 @@ def set_train_mode(m):
     else: m.train()
 
 
-def fit(model, data, epochs, opt, crit, metrics=None, callbacks=None, stepper=Stepper, **kwargs):
+def fit(model, data, epochs, opt, crit, metrics=None, callbacks=None, stepper=Stepper, all_val=False, **kwargs):
```
jph00: This all_val thing seems a bit messy and odd - but I don't understand the problem you're solving so perhaps there isn't a better way. Would be interested to hear more.

fastai/model.py
```diff
@@ -116,6 +116,7 @@ def fit(model, data, epochs, opt, crit, metrics=None, callbacks=None, stepper=St
         stepper.reset(True)
         t = tqdm(iter(data.trn_dl), leave=False, total=num_batch)
         i = 0
+        if all_val: val_iter = iter_batch(data.val_dl)
```
jph00: Why isn't this a python iterator? BTW since it's a class I'd expect to see CamelCase

fastai/model.py
sgugger: all_val is just to say the fit function we want to validate after each batch.

PR #322, opened 2018-04-08
Changed nt to n_tok, to better conform with style guidelines

duckysherwood:
The style guidelines say that the count of something should be `n_` and not `n`; the abbreviations say that `t` is tensor, so I called it `tok` instead (which is used elsewhere).  Also fixed a comment.

Also updated `abbr.md` to add `tok`.

### DIFF
--- a/docs/abbr.md
+++ b/docs/abbr.md
@@ -152,3 +152,4 @@ Note that there are always exceptions, especially when we try to comply with the
 |                  | activation                          | actn           |                                                  |
 |                  |                                     |                |                                                  |
 | **Text**         | string                              | s              |                                                  |
+|                  | token                               | tok            |                                                  |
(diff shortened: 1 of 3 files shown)


docs/abbr.md
```diff
@@ -152,3 +152,4 @@ Note that there are always exceptions, especially when we try to comply with the
 |                  | activation                          | actn           |                                                  |
 |                  |                                     |                |                                                  |
 | **Text**         | string                              | s              |                                                  |
+|                  | token                               | tok            |                                                  |
```
apiltamang: did you mean this to be 'n_tok' instead of 'tok'? I could be mistaken ...

jph00: No `tok` is correct here

duckysherwood: No, there is an entry on line 30 defining the prefix `n_`.  So this like defines `tok` for `token` and `n_tok` is the number of tokens.

jph00:
Much clearer - thanks!

PR #324, opened 2018-04-09
Added docstrings for SortishSampler and for LanguageModelLoader

EvenOldridge:
Added docstrings for SortishSampler and for LanguageModelLoader


jph00:
I've fixed the docs to reflect that it's not +/- 5 stdevs of bptt, but N(bptt,5). I noticed in the forum that you made the same mistake, so looks like you may have misunderstood the code her. Just an FYI

PR #325, opened 2018-04-09
Dynamic Unet future work

KeremTurgutlu:
I've added comments about only supporting size change of H, W -> H/2, W/2 in encoder part and a TODO in order to make this adaptive for custom encoders that might have a different size decay.

### DIFF
--- a/fastai/models/unet.py
+++ b/fastai/models/unet.py
@@ -27,9 +27,10 @@ def conv_bn_relu(in_c, out_c, kernel_size, stride, padding):
 
 
 class UnetBlock(nn.Module):
+    #TODO: ADOPT KERNEL SIZE, STRIDE AND PADDING SO THAT ANY SIZE DECAY WILL BE SUPPORTED
     def __init__(self, up_in_c, x_in_c):
         super().__init__()
-        self.upconv = nn.ConvTranspose2d(up_in_c, up_in_c // 2, 2, 2)
+        self.upconv = nn.ConvTranspose2d(up_in_c, up_in_c // 2, 2, 2) # H, W -> 2H, 2W
         self.conv1 = nn.Conv2d(x_in_c + up_in_c // 2, (x_in_c + up_in_c // 2) // 2, 3, 1, 1)
         self.conv2 = nn.Conv2d((x_in_c + up_in_c // 2) // 2, (x_in_c + up_in_c // 2) // 2, 3, 1, 1)
         self.bn = nn.BatchNorm2d((x_in_c + up_in_c // 2) // 2)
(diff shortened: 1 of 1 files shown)


fastai/models/unet.py
```diff
@@ -27,9 +27,10 @@ def conv_bn_relu(in_c, out_c, kernel_size, stride, padding):
 
 
 class UnetBlock(nn.Module):
+    #TODO: ADAPT KERNEL SIZE, STRIDE AND PADDING SO THAT ANY SIZE DECAY WILL BE SUPPORTED
     def __init__(self, up_in_c, x_in_c):
         super().__init__()
-        self.upconv = nn.ConvTranspose2d(up_in_c, up_in_c // 2, 2, 2)
+        self.upconv = nn.ConvTranspose2d(up_in_c, up_in_c // 2, 2, 2) # H, W -> 2H, 2W
```
apiltamang: Only if you'd be interested ...
I briefly brought this issue up in the forums as well, and I think @jph00 is likely warm towards this: would you mind sharing more info on what _**up_in_c, x_in_c**_ are? Are they tensors, arrays, floats, ints ? And if applicable, what would the generalized shape of these vars  be?

I look at them and I know nothing, nada ... and I assume that being in the thick of things, you're pretty fresh on the material.

Much thanks!

KeremTurgutlu: up_in_c: channel size of the tensor that is been upsampled.
x_in_c: channel size of the tensor that is been concatenated to the upsampled tensor.

You may go through CaravanaUnet notebook in lectures and get a better intuition about the overall process as well as reading Unet paper might help for the decoding part.

Thanks!

fastai/models/unet.py
```diff
@@ -27,9 +27,10 @@ def conv_bn_relu(in_c, out_c, kernel_size, stride, padding):
 
 
 class UnetBlock(nn.Module):
+    #TODO: ADAPT KERNEL SIZE, STRIDE AND PADDING SO THAT ANY SIZE DECAY WILL BE SUPPORTED
```
jph00: What's with THE YELLING? ;)

KeremTurgutlu: Sorry :)

fastai/models/unet.py
apiltamang: True ...
I figured if I said it enough, the word would spread around.
But then, I think it's also by habit. I deal with too many PRs at work where these sorta things keep getting repeated over and over ...

I apologize I haven't been leading by example, though. I really would like to. For the most part, I'm just trying to catch up with the lessons, transitioning to a new job, and just those sorta things. I feel like the PR comments are the least  I can still do, at least for the moment.

Cheers .... and good night!

PR #326, opened 2018-04-09
[documentation]: Adds some existing abbreviations from the codebase

NEIA20:
Hello! 👋 

I was reading through the code and came across some abbreviations that weren't in the `abbr.md` doc. 

I know that there's a push to add documentation to `fastai` right now, and I'm happy to help with that, but is there anything in terms of implementation that needs work or needs to be added? I'd love to contribute something

### DIFF
--- a/docs/abbr.md
+++ b/docs/abbr.md
@@ -59,6 +59,7 @@ Note that there are always exceptions, especially when we try to comply with the
 |                  | directory                           | dir            |                                                  |
 |                  | percentage                          | p              |                                                  |
 |                  | ratio, proportion of something      | r              |                                                  |
+|                  | count                               | cnt            |                                                  |
 |                  |                                     |                |                                                  |
 |                  | figure                              | fig            |                                                  |
 |                  | image                               | im             |                                                  |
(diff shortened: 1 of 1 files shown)


apiltamang:
+1 .
Thanks ...

NEIA20:
@jph00 I have some comments/questions about the sphinx implementation: 

Sphinx takes text files written in [restructuredText (rst)](https://www.southampton.ac.uk/~fangohr/computing/rst/rst.txt) and transforms them into html (although it can generate other outputs as well, like pdfs). In `fastai/docs` there are files written in markdown and asiidoc. Sphinx can work with docs that are written in both markdown and rst, but because markdown and rst are so similar, and because sphinx was built to work with rst, I think that it would probably be simpler to just rewrite the existing docs in rst.

There are other documentation generators, and some that use markdown as their basis, but I think that sphinx seems pretty powerful and useful. I don't think it would be too much trouble to rewrite the existing docs - there aren't too many yet. Sphinx is also easy to integrate with [readTheDocs](https://docs.readthedocs.io/en/latest/) if that's something fastai is interested in as well

If we do want to use sphinx and re-write the docs into rst, that's probably the first thing I'd do

What do you think?

jph00:
Don't worry about Sphinx for now - I haven't done a good job of explaining things and probably won't until I've finished teaching part 2.

NEIA20:
ok no worries 👍

Issue #319, opened 2018-04-08
LanguageModelData.from_dataframes requires test_df


kevindewalt:
## Works
`md = LanguageModelData.from_dataframes(path=data_path, field=TEXT, col='all_text',
                                      train_df=train_df, val_df=val_df, test_df=test_df, bs=bs, bptt=bptt)`

works fine.

## Error
Replacing `test_df=test_df` with `test_df=None` returns error: 

`md = LanguageModelData.from_dataframes(path=data_path, field=TEXT, col='all_text',
                                      train_df=train_df, val_df=val_df, test_df=None, bs=bs, bptt=bptt)`
returns:

`ValueError                                Traceback (most recent call last)
<ipython-input-104-5813a72e373e> in <module>()
      1 md = LanguageModelData.from_dataframes(path=data_path, field=TEXT, col='all_text',
----> 2                                       train_df=train_df, val_df=val_df, test_df=None, bs=bs, bptt=bptt)

~/fastai/fastai/nlp.py in from_dataframes(cls, path, field, col, train_df, val_df, test_df, bs, bptt, **kwargs)
    275     def from_dataframes(cls, path, field, col, train_df, val_df, test_df=None, bs=64, bptt=70, **kwargs):
    276         trn_ds, val_ds, test_ds = ConcatTextDatasetFromDataFrames.splits(text_field=field, col=col,
--> 277                                     train_df=train_df, val_df=val_df, test_df=test_df)
    278 
    279         return cls(path, field, trn_ds, val_ds, test_ds, bs, bptt, **kwargs)

ValueError: not enough values to unpack (expected 3, got 2)`

## investigation
Same error when replacing val_df with None.

mcskinner:
@jph00 any reason for this `None` filtering in `ConcatTextDatasetFromDataFrames` ([code here])(https://github.com/fastai/fastai/blob/16313f83fef7f47b6008643e9eb8f5397fbabe54/fastai/nlp.py#L196):
`return tuple(d for d in (train_data, val_data, test_data) if d is not None)`

It seems easy to remove, and my grepping says the only caller is the one in this stack trace.

jph00:
Yeah it's for use in notebooks where `splits` is often called without a test set.

dalupus:
I've added a pull to fix this.  #347

PR #334, opened 2018-04-10
proposed fix for #319 re language model train/val/test handling

mcskinner:
Fix for #319 
- Add a `map_none` convenience similar to `map_over` (which, should we rename that `map_listy`?)
- Add a flag param to let `None` pass through `ConcatTextDatasetFromDataFrames.splits`.
- Make `LanguageModelData.__init__` `None`-safe.

On that last point, @jph00 does downstream code know how to handle `None` for the datasets and/or dataloaders? If not then this is only a partial fix.

@kevindewalt mind posting a bit more about the code you were trying to run so it can be converted into a regression test? I'll do my honest best in the meantime.

### DIFF
--- a/fastai/nlp.py
+++ b/fastai/nlp.py
@@ -190,9 +190,9 @@ class ConcatTextDatasetFromDataFrames(torchtext.data.Dataset):
     @classmethod
     def splits(cls, train_df=None, val_df=None, test_df=None, keep_nones=False, **kwargs):
         res = (
-            map_none(train_df, partial(cls, **kwargs)),
-            map_none(  val_df, partial(cls, **kwargs)),
-            map_none( test_df, partial(cls, **kwargs)))
+            cls(train_df, **kwargs),
+            cls(val_df, **kwargs),
+            map_none(test_df, partial(cls, **kwargs)))  # not required
         return res if keep_nones else tuple(d for d in res if d is not None)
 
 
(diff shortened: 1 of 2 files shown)


fastai/nlp.py
```diff
@@ -188,12 +188,12 @@ def __init__(self, df, text_field, col, newline_eos=True, **kwargs):
...
-        train_data = None if train_df is None else cls(train_df, **kwargs)
-        val_data = None if val_df is None else cls(val_df, **kwargs)
-        test_data = None if test_df is None else cls(test_df, **kwargs)
-
-        return tuple(d for d in (train_data, val_data, test_data) if d is not None)
+    def splits(cls, train_df=None, val_df=None, test_df=None, keep_nones=False, **kwargs):
+        res = (
+            map_none(train_df, partial(cls, **kwargs)),
+            map_none(  val_df, partial(cls, **kwargs)),
+            map_none( test_df, partial(cls, **kwargs)))
```
jph00: I think only use `map_none` for `test_df` since I don't think we handle the others being None.

fastai/nlp.py
```diff
@@ -250,8 +250,10 @@ def __init__(self, path, field, trn_ds, val_ds, test_ds, bs, bptt, backwards=Fal
         self.pad_idx = field.vocab.stoi[field.pad_token]
         self.nt = len(field.vocab)
 
-        self.trn_dl, self.val_dl, self.test_dl = [LanguageModelLoader(ds, bs, bptt, backwards=backwards)
-                                                  for ds in (self.trn_ds, self.val_ds, self.test_ds) ]
+        factory = lambda ds: LanguageModelLoader(ds, bs, bptt, backwards=backwards)
+        self.trn_dl = map_none(self.trn_ds, factory)
```
jph00: See above

fastai/nlp.py
mcskinner: ha yep, that makes a lot of sense.

PR #335, opened 2018-04-10
Add val_ratio to ImageClassifierData.from_csv

radekosmulski:
A tiny convenience argument allowing for specifying val_ratio when creating an ImagaClassifierData object from_csv vs having to generate the val_idxs and pass that.


jph00:
I don't think this is worth an extra param, compared to just passing in val_idxs.

PR #336, opened 2018-04-11
Update lm_rnn.py

ppleskov:
PoolingLinearClassifier returns log probabilities


jph00:
This isn't normally what is wanted in a module.

PR #337, opened 2018-04-11
added SimpleSampler

ppleskov:
SimpleSampler is a sampler without any sorting, used for a test dataset. Native python Sampler does not work properly.


jph00:
Can you provide a gist or similar showing what problem this solves, and how the existing class is problematic?

PR #338, opened 2018-04-11
A simple convenience method for getting int2tok and tok2int converters

blakewest:
- This is based on the two examples we have of this in the notebooks.

Note that the converters come back identical between the notebook method and the method for this PR.
image


blakewest:
And a shot of the error handler.
image

PR #340, opened 2018-04-12
Adding assertions for fp16 loss scaling

bearpelican:
Also moving sampler to kwargs


bearpelican:
Need to do some testing

PR #341, opened 2018-04-12
Metrics expects target to be a tensor now

bearpelican:
Fixes metrics for pytorch 0.3


bearpelican:
Debugged this with @jph00

PR #342, opened 2018-04-12
Update README.md

anubrata:
the documentation says 
`conda activate fastai` 
instead of 
`source activate fastai`


jph00:
The new version of conda expects the latter.

PR #348, opened 2018-04-15
Defaults num_workers to 0

steveorsomethin:
DataLoader currently defaults num_workers to None, but assumes numeric values only in \_\_iter\_\_. Defaulting to 0 satisfies the assumptions and probable intent in \_\_iter\_\_.


jph00:
The intent is that `None` should result in using `num_cpus()`. If that's broken, a fix would be appreciated!

PR #353, opened 2018-04-16
Proposition to refactor sgdr

sgugger:
Separates the different phases of training with TrainingPhases. Each phase can have its own parameters, including the optimizer.

For now it's independent of the existing classes to not break anything while we debug but ultimately it can be used to do all the existing schedulers.

To use, build a list of TrainingPhase learn.fit_opt_sched(phases).

To change the dataset used for different phases (it will bug if there is a change of batch_size, don't know if it's a feature we want to add as well) use the additional argument list_data (must be the same length as phases).


jph00:
Cool! Might be nice to add *poly* decay too - seems popular for imagenet.

    lr = init_lr*(1 - iter/max_iter)**power

sgugger:
Will put it in the next PR, I've also changed COSINE so that it goes from start to end when an end is specified, and to zero otherwise.
Writing the tutorial notebook, I'm hoping to have it ready today.

PR #354, opened 2018-04-17
[random forests]: Adds `get_average_tree_depth` function

NEIA20:
When trying to reduce overfitting of a random forest by using the `max_depth` parameter, I wanted to know what the average depth of the trees was - that way I'd know which values of `max_depth` to try

This function came in handy, and I thought it might be useful to have it in fastai


jph00:
It's a nice idea, but probably not widely useful enough for adding to the library - thanks for the PR anyway! :)

NEIA20:
@jph00 ok! no worries :) 

Did you see my comment on #326? I can make an issue for the sphinx implementation if that's easier to track

PR #356, opened 2018-04-17
Remove extra cuda call

cheeseblubber:
This is an extra cuda call and will cause some of your model to use cuda even
when specifying USE_GPU=False

### DIFF
--- a/fastai/core.py
+++ b/fastai/core.py
@@ -20,7 +20,7 @@ def T(a, half=False, cuda=True):
         if a.dtype in (np.int8, np.int16, np.int32, np.int64):
             a = torch.LongTensor(a.astype(np.int64))
         elif a.dtype in (np.float32, np.float64):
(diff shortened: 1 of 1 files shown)


jph00:
That's intentional - you should never use HalfTensor on CPU.

cheeseblubber:
Would you have to then convert the Half tensor to float tensor when bringing a model trained on GPU to CPU? If that is the case would this make sense:
```
def T(a, half=False, cuda=True):
    if not torch.is_tensor(a):
        a = np.array(np.ascontiguousarray(a))
        if a.dtype in (np.int8, np.int16, np.int32, np.int64):
            a = torch.LongTensor(a.astype(np.int64))
        elif a.dtype in (np.float32, np.float64):
            a = torch.HalfTensor(a) if half and cuda else torch.FloatTensor(a) # Change is here
        else: raise NotImplementedError(a.dtype)
    if cuda: a = to_gpu(a, async=True)
    return a
```

PR #358, opened 2018-04-17
Added a custom loss func option and some refactoring

ronlut:
This makes setting a custom loss function possible by passing the crit argument to any learner initialization.
For example, now this is possible:

    loss = torch.nn.NLLLoss(weight=class_weights)
    m = md.get_learner(... crit=loss ...)

to get a learner with a weighted loss function.

I removed the `hasattr` check for `is_multi` in `conv_learner.py` as I think it's always true.
Correct me if i'm wrong, as I mainly use `column_data.py` currently.

### DIFF
--- a/fastai/column_data.py
+++ b/fastai/column_data.py
@@ -137,12 +137,14 @@ class MixedInputModel(nn.Module):
 class StructuredLearner(Learner):
     def __init__(self, data, models, **kwargs):
         super().__init__(data, models, **kwargs)
+
+    def _determine_loss_func(self, data):
         if data.is_reg:
-            self.crit = F.mse_loss
+            return F.mse_loss
         elif data.is_multi:
-            self.crit = F.binary_cross_entropy
+            return F.binary_cross_entropy
         else:
-            self.crit = F.nll_loss
+            return F.nll_loss
 
     def summary(self):
         return model_summary(self.model, [(self.data.trn_ds.cats.shape[1], ), (self.data.trn_ds.conts.shape[1], )])
@@ -197,6 +199,7 @@ def get_emb(ni,nf):
     e.weight.data.uniform_(-0.05,0.05)
     return e
 
+
 class EmbeddingDotBias(nn.Module):
     def __init__(self, n_factors, n_users, n_items, min_score, max_score):
         super().__init__()
@@ -210,10 +213,14 @@ class EmbeddingDotBias(nn.Module):
         res = um.sum(1) + self.ub(users).squeeze() + self.ib(items).squeeze()
         return F.sigmoid(res) * (self.max_score-self.min_score) + self.min_score
 
+
 class CollabFilterLearner(Learner):
     def __init__(self, data, models, **kwargs):
         super().__init__(data, models, **kwargs)
-        self.crit = F.mse_loss
+
+    def _determine_loss_func(self, data):
+        return F.mse_loss
+
 
 class CollabFilterModel(BasicModel):
     def get_layer_groups(self): return self.model
--- a/fastai/column_data.py
+++ b/fastai/column_data.py
@@ -138,16 +138,9 @@ class StructuredLearner(Learner):
     def __init__(self, data, models, **kwargs):
         super().__init__(data, models, **kwargs)
 
-    def _determine_loss_func(self, data):
-        if data.is_reg:
-            return F.mse_loss
-        elif data.is_multi:
-            return F.binary_cross_entropy
-        else:
-            return F.nll_loss
+    def _get_crit(self, data): return F.mse_loss if data.is_reg else F.binary_cross_entropy if data.is_multi else F.nll_loss
 
-    def summary(self):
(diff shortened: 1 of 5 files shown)


ronlut:
I removed the `hasattr` check for `is_multi` in `conv_learner.py` as I think it's always true.
Correct me if i'm wrong, as I mainly use `column_data.py` currently.

fastai/text.py
```diff
@@ -204,7 +204,9 @@ def get_model(self, opt_fn, emb_sz, n_hid, n_layers, **kwargs):
 class RNN_Learner(Learner):
     def __init__(self, data, models, **kwargs):
         super().__init__(data, models, **kwargs)
-        self.crit = F.cross_entropy
+
+    def _determine_loss_func(self, data):
```
jph00: You can put these on one line.

fastai/column_data.py
```diff
@@ -137,12 +137,14 @@ def forward(self, x_cat, x_cont):
 class StructuredLearner(Learner):
     def __init__(self, data, models, **kwargs):
         super().__init__(data, models, **kwargs)
+
+    def _determine_loss_func(self, data):
         if data.is_reg:
-            self.crit = F.mse_loss
+            return F.mse_loss
```
jph00: Maybe better to say

    return F.mse_loss if data.is_reg else F.binary_cross_entropy if data.is_multi else F.nll_loss

fastai/conv_learner.py
```diff
@@ -95,15 +95,20 @@ class ConvLearner(Learner):
     def __init__(self, data, models, precompute=False, **kwargs):
         self.precompute = False
         super().__init__(data, models, **kwargs)
-        if hasattr(data, 'is_multi'):
```
jph00: This isn't always true - many users use their own custom ModelData classes. So need to put this back

fastai/conv_learner.py
```diff
@@ -95,15 +95,20 @@ class ConvLearner(Learner):
...
-            elif self.metrics is None:
-                self.metrics = [accuracy_thresh(0.5)] if self.data.is_multi else [accuracy]
+        if not data.is_reg and self.metrics is None:
+            self.metrics = [accuracy_thresh(0.5)] if self.data.is_multi else [accuracy]
         if precompute: self.save_fc1()
         self.freeze()
         self.precompute = precompute
 
+    def _determine_loss_func(self, data):
+        if data.is_reg:
```
jph00: You can use the ternary operator here too

fastai/learner.py
```diff
@@ -383,6 +384,7 @@ def fit_opt_sched(self, phases, cycle_save_name=None, best_save_name=None, use_w
         if len(data_list)==0: data_list = [self.data]
         return fit(self.model, data_list, n_epochs,layer_opt, self.crit,
             metrics=metrics, callbacks=callbacks, reg_fn=self.reg_fn, clip=self.clip, fp16=self.fp16, **kwargs)
-    
 
+    def _determine_loss_func(self, data):
```
jph00: Rather than raising error here, just set it to some default (mse seems fine). They might set `crit` later so no need for an exception now.

jph00: Also maybe a better name is `_get_crit` since we normally use `get` prefix in this situation, and use `crit` instead of loss.

fastai/column_data.py
ronlut: I think readability here is more important than making it short...
>  Readability Counts 

(PEP20)

Also, when and if someone will implement multi-output multi-class (or multi-output regression), things will get even more complicated.

If you're sure about this one liner, tell me. Honestly I don't really see the benefit in this case :)

jph00:
I don't feel strongly about it, but the reasoning is in the style guide
if you're interested:
https://github.com/fastai/fastai/blob/master/docs/style.md
--
Jeremy Howard


On Tue, Apr 17, 2018, at 2:39 PM, Rony Lutsky wrote:
> *@ronlut* commented on this pull request.


> 
> In fastai/column_data.py[1]:


> >          if data.is_reg: -            self.crit = F.mse_loss +
> >          return F.mse_loss
>> I think readability here is more important than making it short...


>> Readability Counts


> (PEP20)


> Also, when and if someone will implement multi-output multi-class (or
> multi-output regression), things will get even more complicated.> If you're sure about this one liner, tell me. Honestly I don't really
> see the benefit in this case :)> — You are receiving this because you commented. Reply to this email
> directly, view it on GitHub[2], or mute the thread[3].> 


Links:

  1. https://github.com/fastai/fastai/pull/358#discussion_r182242052
  2. https://github.com/fastai/fastai/pull/358#discussion_r182242052

ronlut:
Made the changes you requested 👍 
Tell me if something is missing.

> but the reasoning is in the style guide

In that case, I don't really agree with this section of the style guide (sorry, didn't notice it until you mentioned), but let's save that discussion for another occasion :)

PR #359, opened 2018-04-17
Added Polynomial annealing

sgugger:
This should be the last PR for the new API.
I wrote [this tutorial](https://github.com/sgugger/Deep-Learning/blob/master/Understanding%20the%20new%20fastai%20API%20for%20scheduling%20training.ipynb) and would love to have your thoughts before publishing it on the forum.

### DIFF
--- a/fastai/sgdr.py
+++ b/fastai/sgdr.py
@@ -418,12 +418,14 @@ class DecayType(IntEnum):
     LINEAR = 2
     COSINE = 3
     EXPONENTIAL = 4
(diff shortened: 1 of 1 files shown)


sgugger:
Understood, will remove the math and add more code ;-). Of course you can include it in the next lesson if you think that could be useful.

I'll add some examples for the polynomial decay as well and post the notebook on the forum once I'm done.

binga:
This notebook is awesome!

PR #364, opened 2018-04-17
Add module decisions file

anshbansal:
This is based on Jeremy's comment in Lesson 8.

### DIFF
--- a/docs/module-decisions.md
+++ b/docs/module-decisions.md
@@ -1,0 +1,13 @@
+# Module Decisions
+
+## Introduction
+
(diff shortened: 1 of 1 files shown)


rohitgeo:
@anshbansal for bounding boxes, the format that fastai uses is [ymin, xmin, ymax, xmax].

Pascal VOC uses xmin, ymin, width, height, not (coordinates top right, (height, width--)

... so the following should be updated from
Will use (coordinates top _right_, coordinates bottom right) instead of (coordinates top _right_, (_height_, _width_))
to
Fastai uses (coordinates top _left_, coordinates bottom right) instead of (coordinates top _left_, (_width_, _height_))
or simply:
Fastai uses bounding boxes defined as [ymin, xmin, ymax, xmax]

PR #365, opened 2018-04-17
Add mean_f_score

radekosmulski:
Add a mean f score as used by Kaggle

Kaggle wiki seems to be down atm but it is documented here: https://www.kaggle.com/wiki/MeanFScore


radekosmulski:
New version pushed

PR #369, opened 2018-04-18
Bug fix

sgugger:
On_phase_begin always needs to be called **before** passing the optimizer to the stepper (otherwise the stepper has an optimizer that isn't updated).

Also, changing the optimizer causes to zero all the buffers in it (which is kind of logical) so we lose all the previous moving averages. For now I only create a new one if there is a change of name (so if we keep SGD with momentum for instance we don't reinitialize the moving average of the gradients), but perhaps it could be interesting to 'match' the buffers in some way?

### DIFF
--- a/fastai/layer_optimizer.py
+++ b/fastai/layer_optimizer.py
@@ -57,7 +57,8 @@ class LayerOptimizer():
             for pg in self.opt.param_groups: pg['alpha'] = beta
 
     def set_opt_fn(self, opt_fn):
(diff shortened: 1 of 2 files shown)


jph00:
Yeah I don't know if that's fixable...

sgugger:
Oh, there's definitely a possibility since all the buffed values are stored in a dictionary. Wonder if it could work somehow, to match it (aka momentum with beta1 for instance) and pass it along when we change optimizer.

PR #370, opened 2018-04-19
Add data loader documentation

thedch:
Jeremy — 

As you have mentioned in lecture the past few weeks, I am going to try to contribute to the documentation. As I have worked on this document, I have found it an excellent way to understand the library better, so hopefully this is beneficial to you (and general users of the library) as well. 

I have no doubt that this will require some revisions, this is my first draft. I have tried to adhere to your template as best I can. Despite plenty of `pdb.set_trace()`, I still do not fully understand some of the methods, so my descriptions may need modifications or rewriting. Please let me know of parts you would like me to revisit, and I will do my best. 

Thank you,
Daniel


jph00:
Many thanks! I'd suggest studying the pytorch DataLoader, that has almost identical API to this, and then in the docs only mention the differences (and point to the pytorch DataLoader docs for the details of the rest).

jph00:
Yup you should add them too.

Also note that you should write your docs in the form shown in `transform_tmpl.adoc`. The stuff in {{...}} will be expanded using suitable templates automatically.

jph00:
Not sure how we ended up with so many commits in one PR! I'll close this, and each change can get added to a separate PR - hope that's OK.

(If it's not - please just let me know, and I can reopen this and we can discuss how to merge it in a way that preserves my sanity...)

PR #377, opened 2018-04-22
Added write_dir kwarg to allow specification of a custom writing path

JannesKlaas:
Writing to the data path is not always possible. In Kaggle Kernels for example, the data input path is protected.

This change adds the option to pass a write_dir to the Learner which will then be used to write the tmp and model files. 

Usage:
```Python
WDIR = 'write_dir'
kwargs = {'write_dir':WDIR}
learn = ConvLearner.pretrained(arch, data, precompute=True, **kwargs)
```

### DIFF
--- a/fastai/learner.py
+++ b/fastai/learner.py
@@ -16,7 +16,7 @@ import time
 
 
 class Learner():
-    def __init__(self, data, models, opt_fn=None, tmp_name='tmp', models_name='models', metrics=None, clip=None, crit=None):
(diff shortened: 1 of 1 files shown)


jph00:
I'd rather not added more params - instead, perhaps make it so if the tmp and models paths are absolute paths then it doesn't do `path.join`?

JannesKlaas:
That would be cleaner indeed, sending a new PR

PR #378, opened 2018-04-22
Fix for texts_labels_from_folders

gradzio:
It did not allow to load some of the data sets due to utf8 error.


jph00:
Best to fix this in your environment instead.

PR #382, opened 2018-04-22
Fix shell command in README.md

ashtonsix:


jph00:
It's called `conda` in the new version.

PR #383, opened 2018-04-22
Add support for data loaders w/ multiple inputs

ashtonsix:
Hi, thanks for your brilliant course & this library!

I'm working on the [SNLI dataset](https://nlp.stanford.edu/projects/snli/), & created this `DataLoader` to consume my iterators:

```py
class DLBridge():
    def __init__(self, dataloader):
        self.dataloader = dataloader
        self.emb_onehot = nn.Embedding(config.d_out, config.d_out)
        self.emb_onehot.weight.data = torch.eye(config.d_out)
    def onehot(self, t):
        return self.emb_onehot(t.type(torch.LongTensor)).cuda()
    def __len__(self):
        return len(self.dataloader)
    def __iter__(self):
        for i, batch in enumerate(self.dataloader):
            yield (batch.premise, batch.hypothesis), self.onehot(batch.label)

baseline_data = ModelData(f'{PATH}/models/baseline', DLBridge(train_iter), DLBridge(dev_iter))
```

Here, `x` is a tuple, which gets destructured in the forward pass through my model (from one of [PyTorch's awesome examples](https://github.com/pytorch/examples/tree/master/snli)):

```py
class BaselineModel(nn.Module):
    def __init__(self):
        # ...
    def forward(self, batch):
        prem_embed = self.embed(batch[0])
        hypo_embed = self.embed(batch[1])
        if self.config.fix_emb:
            prem_embed = Variable(prem_embed.data)
            hypo_embed = Variable(hypo_embed.data)
        if self.config.projection:
            prem_embed = self.relu(self.projection(prem_embed))
            hypo_embed = self.relu(self.projection(hypo_embed))
        premise = self.encoder(prem_embed)
        hypothesis = self.encoder(hypo_embed)
        scores = self.out(torch.cat([premise, hypothesis], 1))
        return scores
```

I had to modify `core.py` to get this working properly.

### DIFF
--- a/fastai/core.py
+++ b/fastai/core.py
@@ -32,9 +32,9 @@ def create_variable(x, volatile, requires_grad=False):
     return x
 
 def V_(x, requires_grad=False, volatile=False): return create_variable(x, volatile, requires_grad)
-def V (x, requires_grad=False, volatile=False): return map_over(x, lambda o: V_(o, requires_grad, volatile))
+def V (x, requires_grad=False, volatile=False): return map_over(x, lambda y: map_over(y, lambda o: V_(o, requires_grad, volatile)))
 def VV_(x): return create_variable(x, True)
-def VV (x): return map_over(x, VV_)
+def VV (x): return map_over(x, lambda y: map_over(y, VV_))
 
 def to_np(v):
     if isinstance(v, (np.ndarray, np.generic)): return v


ashtonsix:
Hmm... I should probably get this working for inputs that are dicts too. What do you think?

jph00:
Sorry I'm not following either the code added or your description. Could you try to show a really simple example of what is calling this new code, and why? Are there any approaches to solving this problem that wouldn't involve changing these functions? (I'd like to keep them as simple as possible since they're so widely used.)

ashtonsix:
Some networks (like mine) need to accept multiple inputs in `.forward` & do something like this:

```py
class Model(nn.Module):
    def __init__(self):
        # ...
    def forward(self, batch):
        firstInput = batch.firstInput
        secondInput = batch.secondInput
        # ...
```

Right now `fastai` assumes every `DataLoader` will yield a 2-tuple of tensors `(x, y)` & crashes if `x` is an arbitrary object that contains tensors because some functions in `model.py` (`fit`, `validate`, `validate_next`) try to convert that arbitrary object into a `torch.Variable` (by calling `V`, leading to `NotImplementedError` from `T`).

My code stops `V` crashing on lists/tuples with a depth of 2, (depth-2 because functions like `fit` wrap `x` in a list).

| Code                | Works now | Works after my change |
|---------------------|-----------|-----------------------|
| V(torch.Tensor)     | :heavy_check_mark:         | :heavy_check_mark:                     |
| V([torch.Tensor])   | :heavy_check_mark:         | :heavy_check_mark:                     |
| V([(torch.Tensor)]) | ❌         | :heavy_check_mark:                     |

Supporting this requires either one modification in `core.py` or several in `model.py`.

Instead of making `V` find tensors in the input and convert them to variables, we could rename `create_variable` to `maybe_create_variable` & make it behave like the identity function when `T` throws errors.

arvind-cp:
Hi @ashtonsix ,

I am working with the quora duplicates kaggle dataset which takes an input pair, just like SNLI.
Please see how I made a custom pair dataset & dataloader in https://github.com/arvind-cp/fastai/blob/arvind-cp-LM-eval/courses/dl2/FiTLAM%20-%20Quora%20Duplicate%20Pairs%20-%20STS%20task.ipynb

It still breaks the stepper methods. So, in model.py you can edit the step and evaluate methods like this:

from:
output = self.m(*xs)

to:

if len(xs) > 1: output = self.m(*xs)
else: output = self.m([*xs])

Please let me know if this solution will work for you.

PR #387, opened 2018-04-24
Adoc

cuddle-cuddle:
start to document fastai library: 
-- one-liners for .py files
-- more detailed explanations in asciidoc files


.gitignore
```diff
@@ -2,6 +2,7 @@
 .pypirc
 ~*
 tmp*
+sample_data/
```
jph00: Could you rename to `tmp_sample_data` - that way it'll be ignored already, and we won't need anything added to .gitignore

fastai/dataset.py
```diff
@@ -36,6 +39,12 @@ def resize_img(fname, targ, path, new_path):
     im.resize(sz, Image.LINEAR).save(dest)
 
 def resize_imgs(fnames, targ, path, new_path):
+    """
+    Enlarge or shrink a set of images in the same directory to scale, such that the smaller of the height or width dimension is equal to targ.
+    Note:
```
jph00: Let's move the notes into adoc - not sure these are urgent enough to be here

fastai/dataset.py
```diff
@@ -90,6 +102,10 @@ def create_val(path, r):
     copy_or_move_with_subdirs(subdirs, path, val_path, r, move=True)
 
 def copy_or_move_with_subdirs(subdir_lst, src, dst, r, move=False):
+    '''
```
jph00: Is this used anywhere? I don't know what it is. I don't think your current docs are correct - probably best not to add a doc if you don't know what something is. Maybe instead just add `TODO: doc` as the docstring if you don't know?

(For this one though, if it's not used anywhere in fastai we can remove it - but be sure to check both fastai and notebook dirs)

fastai/transforms.py
```diff
@@ -29,10 +31,13 @@ def stretch_cv(x,sr,sc,interpolation=cv2.INTER_AREA):
     return x[cr:r+cr, cc:c+cc]
 
 def dihedral(x, dih):
+    '''performs any of 8 90 rotations or flips for image x.
+    '''
     x = np.rot90(x, dih%4)
     return x if dih<4 else np.fliplr(x)
 
 def lighting(im, b, c):
+    ''' adjusts image's balance and contrast'''
```
jph00: This should be more specific. Might be easier to add the docstrings *after* the asciidoc sometimes, since for the asciidoc you can do sample images etc -  which will mean you'll have more info about the details.

Also, note that for fastai.transforms a lot of the methods just support one transform (e.g. `lighting` supports `RandomLighting`) - so maybe just add the doc to the transform class, and in the method say 'implementation for RandomLighting' or something

fastai/transforms.py
```diff
@@ -88,6 +93,7 @@ def googlenet_resize(im, targ, min_area_frac, min_aspect_ratio, max_aspect_ratio
     return out
 
 def cutout(im, n_holes, length):
+    ''' cuts out n_holes number of square holes of size length in image at random locations. holes may be overlapping. '''
```
jph00: Probably more useful to say 'Implements *cutout* augmentation as described in <URL>' and link to the paper

PR #388, opened 2018-04-24
port LSUV initializer example into a pytest

mcskinner:
I plan to wrap this logic in a class and get rid of the globals, clean it up, etc. This quick test helped me understand enough about the goals to make a go at refining as needed, and it'll also keep me from breaking things again as I refactor.

The `pytest.fixture` stuff is worth noting. It seems really cool for loading data once and then reusing it across tests, but I'm not if sure there's a policy on putting images into the repository. I took the first 3 from the original repo's example here, but am happy to adjust.


jph00:
Thanks! There's already an image in fastai/images - could you either use that image, or if you need some specific image put yours there? Give it a name that doesn't include 'lsuv' - so that other people can use it in their sample code too. (Also we should rename the image that's already in there to be more descriptive!)

mcskinner:
Done. I went with "industrial_fishing" for the name, but open to alternatives. This is extra nice because now that `image_data` fixture can be reused across any image processing tests! We can do something similar for each style of data loading.

PR #390, opened 2018-04-24
Learner write to tmp_name / models_name absolute paths.

JannesKlaas:
As discussed in #377 , the learner now writes to the absolute path specified in `tmp_name` or `models_name` if they are absolute paths. Otherwise it creates sub directories in the data folder with `path.join`


jph00:
Great!

Issue #392, opened 2018-04-25
AttributeError: ‘list’ object has no attribute ‘data’ in pascal notebook


bharadwaj6:
happens in single object detection section.

More details here: http://forums.fast.ai/t/attributeerror-list-object-has-no-attribute-data/15301

gsgbills:
I had that error, and fixed it by replacing in model.py:212
res.append([f(preds.data,y.data) for f in metrics]) 
by 
res.append([f(preds.data,y) for f in metrics]) 

It may be related to the change to Pytorch 0.4....

hiromis:
I can confirm it's not related to Pytorch 0.4 - I tested with both 0.3 and 0.4 and got the same error.

PR #393, opened 2018-04-25
fix for metric calculation in model validate

bharadwaj6:
for #392


jph00:
Hmmm... I'm pretty sure I added this for some apparently good reason, but now not sure what that was! I'll merge this PR but let's keep an eye out for regressions, especially on pytorch 0.4.

PR #395, opened 2018-04-25
Ensure pytorch version is less than 0.4 in environment-cpu.yaml file

mattmcclean:
Now that PyTorch 0.4 is released, the environment file needs to be updated to prevent pytorch version 0.4  being installed as there are breaking changes in this version. Matches the pytorch entry in the file ```environment.yaml``` which keeps version 0.3.1.

### DIFF
--- a/environment-cpu.yml
+++ b/environment-cpu.yml
@@ -94,7 +94,7 @@ dependencies:
 - widgetsnbextension
 - xz
 - zeromq
-- pytorch>=0.2.0
+- pytorch<0.4
 - bcolz
 - prompt_toolkit
 - pytest


Zohaggie:
Can you please outline the "breaking changes"? 

Please note that with the latest 0.4 release CUDA 5.0 capability is being restored, See https://discuss.pytorch.org/t/pytorch-no-longer-supports-this-gpu-because-it-is-too-old/13803/21 and https://discuss.pytorch.org/t/pytorch-no-longer-supports-this-gpu-because-it-is-too-old/13803/26

mattmcclean:
I re-tested my application (based on lesson1 notebook code) and it is actually works ok however I do get the following warning when initializing my model from a persisted local copy: 
```
/opt/program/fastai/initializers.py:6: UserWarning: nn.init.kaiming_normal is now deprecated in favor of nn.init.kaiming_normal_.
  if hasattr(m, 'weight'): init_fn(m.weight)
```

jph00:
I believe everything should work correctly on 0.4. I'll close this, and instead if there are breakages in 0.4 please opens issues/PRs for each one.

PR #396, opened 2018-04-25
added LogSoftmax layer; modified import syntax

WNoxchi:
See [Gist](https://gist.github.com/WNoxchi/17ed83c3a0450cb4939b264d94c74054) for testing. 

`from .layers import *` --> `from fastai.layers import *`  can now import darknet normally via: `from fastai.models import darknet`.
The LogSoftmax layer works with the NLL loss criterion set by fastai's `conv_learner.ConvLearner` to produce Cross Entropy loss -- this is done automatically through fastai's `ConvnetBuilder` when loading pretrained models, but darknet isn't pretrained yet.

See [fastai thread](http://forums.fast.ai/t/project-implement-yolo-v3-backbone-and-preact-resnet/14613/50?u=borz) for how this started.

### DIFF
--- a/fastai/models/darknet.py
+++ b/fastai/models/darknet.py
@@ -1,7 +1,7 @@
 import torch
 import torch.nn as nn
 import torch.nn.functional as F
-from .layers import *
+from fastai.layers import *
 
 class ConvBN(nn.Module):
     "convolutional layer then batchnorm"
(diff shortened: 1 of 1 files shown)


jph00:
I'd rather we moved towards using the loss functions with integrated softmax, rather than having them in the models.

PR #403, opened 2018-04-27
Adoc: new documentation, and example under /doc/ folder

cuddle-cuddle:
adding a folder with some small sample data (~10s of images), and ipython notebooks to explain all fastai classes.


jph00:
I'd rather not auto-add all the skeletons. Let's only add them once we have something we're happy with. I'll close this PR - instead could you please just add stuff that's not auto-generated?

PR #407, opened 2018-04-29
[docs]: Adds comment to `read_dir`

NEIA20:


mcskinner:
@NEIA20 @jph00 this PR and others to the same file have whitespace inconsistency.

Running the tests `python -m pytest -s tests` fails during startup.

Proposed fix in #410.

PR #408, opened 2018-04-29
[docs]: Adds comment to `folder_source` and updates a couple variable names

NEIA20:


fastai/dataset.py
```diff
@@ -82,8 +82,18 @@ def n_hot(ids, c):
...
+  
+  Returns:
+  -------
+  fnames: a list of the filenames within `folder`
+  all_lbls: a list of all of the labels in `folder`, where the # of labels is determined by the # of directories within `folder`
+  lbl_arr: a numpy array of the label indices in `all_lbls`
+  
+  """
     fnames, lbls, all_lbls = read_dirs(path, folder)
-    lbl2idx = {v:k for k,v in enumerate(all_lbls)}
```
jph00: I don't mind this change - but FYI I pretty much always use `k,v` for key and value in enumerations and dict comprehensions. Sorry if it's missing from the abbrev file!

NEIA20: oh i see my bad, i'll remember that 👍I was thinking that `k` meant `key` for objects specifically

And don't worry `key` and `value` are in the abbrev file!

PR #409, opened 2018-04-29
add test for `text.py` sampler implementations

mcskinner:
I noticed some periodic lag while using `SortishSampler`, and found a bug that might help explain it. This test covers the current behavior, with a commented bit for the bugfix. Next PR will cover the bug fix.


mcskinner:
Actually a second PR is silly. I've just done the fix as a second commit on this PR.

jph00:
Thanks! Some online comments as to what that new code is doing would be helpful I think - it's not obvious to me at least.

PR #412, opened 2018-04-29
fix typo in Learner.__init__ path management

mcskinner:
@JannesKlaas one of the new incantations was spelled `data_path` instead of `data.path`.

PR #414, opened 2018-04-30
Fix misnamed data.path access in Learner library

nickrobinson:
A recent change to learner.py resulted in a misnamed reference to self.data_path instead of self.data.path.


nickrobinson:
Resolved in #412

PR #416, opened 2018-04-30
Dataloader doc

thedch:
Jeremy —

(This PR is a followup from #370)

As you noticed, I messed up my fastai fork (`git rebase` is dangerous!) earlier today when working on the docs. I have revised the formatting to better fit the template doc, and added method arguments.

### DIFF
--- a/docs/dataloader.adoc
+++ b/docs/dataloader.adoc
@@ -4,7 +4,7 @@
 
 *Note:* the fastai DataLoader has a similar API to the PyTorch DataLoader. Please see http://pytorch.org/docs/master/data.html#torch.utils.data.DataLoader[the PyTorch documentation] for usage and details. The documentation presented here focuses on the differences between the two classes.
 
-== {{class DataLoader, dataset, batch_size=1, shuffle=False, sampler=None, batch_sampler=None, pad_idx=0, num_workers=None, pin_memory=False, drop_last=False, pre_pad=True, half=False, transpose=False, transpose_y=False}}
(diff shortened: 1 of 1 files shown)


docs/dataloader.adoc
```diff
@@ -0,0 +1,59 @@
...
+
+transpose (type bool, default `False`)::
+    If `True`, each batch will have its inputs transposed.
+
+transpose_y (type bool, default `False`)::
+    If `True`, each batch will have its outputs (labels) transposed.
+
+=== {{methods}}
+
+{{method jag_stack(b),}}
```
jph00: These should be in the format: `{{method jag_stack,b}}`. i.e. the list of params comes after the comma. It's possible the template script isn't handling this correctly - in which case feel free to have a go at fixing it! :)

docs/dataloader.adoc
```diff
@@ -0,0 +1,59 @@
...
+shuffle::
+
+sampler::
+
+batch_sampler::
+
+pad_idx (type int, default 0)::
+    A padding index representing how many zeros to add to each batch. See: `pre_pad`
+
+num_workers (type int, default `None`)::
```
jph00: The arguments should be a template, eg {{arg pad_idx,int,None}}. Not sure if that template has been created yet or not...

thedch:
Jeremy -- 

Thanks for a great course, I had a great time. Per our convo last night, I've updated the formatting.

PR #425, opened 2018-04-30
change using self.data_path to self.data.path

talhadar:
using self.data_path in line 37 causes an error: AttributeError: ‘ConvLearner’ object has no attribute ‘data_path’


adai183:
I had the same bug and tested the proposed solution. 
This does the job!

PR #426, opened 2018-04-30
Added conda-forge channel for jupyter_contrib_nbextensions

sudarshan85:
Currently, jupyter_contrib_nbextensions are installed using pip in the environment files. I noticed that due to this, all the extensions does not show up. According to the official [repo](https://github.com/ipython-contrib/jupyter_contrib_nbextensions), installing through the conda-forge channel is the preferred method. Furthermore, there were duplicate entries of this package in the env files under pip. I removed the duplicate  entries and added the conda-forge channel and the package under that.

### DIFF
--- a/courses/dl2/imdb.ipynb
+++ b/courses/dl2/imdb.ipynb
@@ -760,7 +760,7 @@
     "    dropouti=drops[0], dropout=drops[1], wdrop=drops[2], dropoute=drops[3], dropouth=drops[4])\n",
     "\n",
     "learner.metrics = [accuracy]\n",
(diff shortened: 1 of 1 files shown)


jph00:
I don't feel comfortable adding conda-forge - and certainly not at such high priority. It'll impact a lot of packages. Is there a way to fix this one package such that conda-forge is only used for this, and nothing else?

sudarshan85:
Do you see conda-forge potentially creating conflicts in the future?

jph00:
Yes I do.

PR #427, opened 2018-05-01
Fixes transpose bug in dataloader

RudrakshTuwani:
The current implementation of transpose assumes that there is only one feature array. In many situations, the DataSet object returns multiple features, in which case the first feature ends up being transposed whereas the rest is left as it is. This update makes the transpose implementation generic and lets it handle multiple features.

### DIFF
--- a/fastai/dataloader.py
+++ b/fastai/dataloader.py
@@ -73,8 +73,8 @@ class DataLoader(object):
 
     def get_batch(self, indices):
         res = self.np_collate([self.dataset[i] for i in indices])
-        if self.transpose:   res[0] = res[0].T
-        if self.transpose_y: res[1] = res[1].T
+        if self.transpose:   res[:-1] = [r.T for r in res[:-1]]
+        if self.transpose_y: res[-1] = res[-1].T
         return res
 
     def __iter__(self):


jph00:
I'm not sure this solution is quite right - seems that it will create a list of indep vars, even if there's just one. And if the indep vars is a tensor, it'll turn it into a list.

Might be a good idea to write a few tests of various edge cases as part of fixing this PR...

fastai/dataloader.py
```diff
@@ -73,8 +73,8 @@ def np_collate(self, batch):
 
     def get_batch(self, indices):
         res = self.np_collate([self.dataset[i] for i in indices])
```
RudrakshTuwani: res is itself a list of length equal to the number of outputs of __getitem__ method of DataSet. The old dataloader assumed this length to be 2 and transposed **only** the first array if transpose was set to True. The update makes the assumption that the last array in res list contains our target label, whereas all the preceding arrays are features.  It thus, contrary to the old dl, transposes all the arrays up to the last array in the res list if transpose is set to True.

jph00: Thanks - so it works fine with DataLoader - but looks like it is indeed listifying the inputs, even if they're not originally lists. That isn't the expected behaviour of a DataSet, and might break some code.

jph00: DataSets are not only used with DataLoader. They're also used directly. If you have just a single tensor as x, you're returning a list [x], rather than leaving it as-is. Does that make sense?

RudrakshTuwani: Oh, I haven't made any changes to the DataSets, just the get_batch method of DataLoaders in case of transpose being set True. Also, the new dataloader doesn't return a list.
```
*x,y = next(iter(trn_dl))
```
The asterisk (*) is actually the reason why x ends up becoming a list. If you instead use the following, then each xi is a Tensor.
```
x1,....,xn, y = next(iter(trn_dl)) 
```

PR #428, opened 2018-05-01
Asserts for ImageClassifierData.from_csv

benediktbrandt:
Added two asserts to make sure that ImageClassifierData.from_csv is correctly invoked. 

The argument "folder" needs to be a relative path (relative to the argument called path). If it is given as an absolute path there is no error and the code still mostly works, but it breaks some functionality like data = ImageClassifierData.from_csv(..); data.resize(..);

The argument tfms of course should not be none (), since we might need to resize images, etc.

### DIFF
--- a/fastai/dataset.py
+++ b/fastai/dataset.py
@@ -451,6 +451,8 @@ class ImageClassifierData(ImageData):
         Returns:
             ImageClassifierData
         """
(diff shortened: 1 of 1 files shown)


jph00:
Makes sense - might be nice in the future however to make abs paths work correctly.

PR #430, opened 2018-05-01
modify LR_Finder.plot to use n_skip_end=0

JanLauGe:
I would like to suggest modifying the learning rate finder method LR_Finder.plot so that `n_skip_end=0` plots all values. Previously, one had to use `n_skip_end=1` to get all values, while n_skip_end=0 would return an empty plot. The new behavior makes `n_skip_end` consistent with the behavior of `n_skip`


jph00:
Good thinking

PR #433, opened 2018-05-01
Extract transforms tests & some fixes as separate PR from #312

PiotrCzapla:
I think we can get the test for transforms.py merged first, then the #312 will be easier to read as it will show only the changes done to allow for custom_transformations.

I will add more test showing how the transformation API works right now and once we are happy that the coverage is acceptable I'm going to update the changes in #312 to show that they are backward compatible.

### DIFF
--- a/tests/test_transform.py
+++ b/tests/test_transform.py
@@ -1,0 +1,288 @@
+import numpy as np
+import pytest
+
+from fastai.transforms import *
+
+t_rand_img128x128x1 = np.random.uniform(size=[128,128,1])
+t_rand_img128x128x3 = np.random.uniform(size=[128,128,3])
+#
+# # as per https://stackoverflow.com/questions/7100242/python-numpy-first-occurrence-of-subarray
+# def rolling_window(a, size):
+#     shape = a.shape[:-1] + (a.shape[-1] - size + 1, size)
+#     strides = a.strides + (a. strides[-1],)
+#     return np.lib.stride_tricks.as_strided(a, shape=shape, strides=strides)
+
+
+def test_scale_min_works_with_masks():
+    mask = np.ones([128, 256], dtype=np.float32)
+    mask[0:64,0:128] = 20
+
+    em = np.array([[20., 20., 1., 1.],
+                   [1., 1., 1., 1.]], dtype=np.float32)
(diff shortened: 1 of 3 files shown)


jph00:
This is great!

tests/test_transform.py
```diff
@@ -0,0 +1,313 @@
...
+    im = np.array([[0, 0, 0, 0, 0],
+                   [0, 1, 1, 1, 0],
+                   [0, 1, 1, 1, 0],
+                   [0, 1, 1, 1, 0],
+                   [0, 0, 0, 0, 0], ], dtype=np.float32)
+    expect = [1,1,3,3]
+    np.testing.assert_array_equal(to_bb(im, "not used"), expect)
+
+### tests for transformation objects
+def test_RandomCrop():
```
mcskinner: (1) Sorry for leaving my last comments on the commit and not here...
(2) One last thought is that this violates method naming conventions in general, not sure if we care about that vs the class name.

PiotrCzapla: Sticking to PEP8 around naming test functions is bad idea as the functions names are used in different context.
The most important thing is that you know what the tests are about when you read the function names.
I like bdd specification style of test (wiki)[https://en.wikipedia.org/wiki/Behavior-driven_development#Specialized_tooling_support], (ScalaTest as example)[http://www.scalatest.org], where you write a spec in plain English, if you know any good way of doing such thing in pytest please share.

In py.test I tend to follow this convention:

test_<object_of_test>__<what function is being tested> 

But I'm not sure consistent around that (I've forgotten to put double underscore to separate object from a feature that is being tested), I'm fine to adopt some good convention.

mcskinner: We're trying to follow the convention from the Python testing guide:

>Use long and descriptive names for testing functions. The style guide here is slightly different than that of running code, where short names are often preferred. The reason is testing functions are never called explicitly. square() or even sqr() is ok in running code, but in testing code you would have names such as test_square_of_number_2(), test_square_negative_number(). These function names are displayed when a test fails, and should be as descriptive as possible.

See also: https://github.com/fastai/fastai/blob/master/docs/testing.adoc

PR #441, opened 2018-05-03
Adds test file for `core.py` and tests for `partition` fn

NEIA20:


tests/test_core.py
```diff
@@ -0,0 +1,39 @@
...
+
+def test_partition_functionality():
+  sz = 2
+  a = [1,2,3,4,5]
+  ex = [[1,2],[3,4],[5]]
+  result = partition(a, sz)
+  assert len(result) == len(ex)
+  assert all([a == b for a, b in zip(result, ex)])
+
+  sz = 3
```
mcskinner: I might suggest that we split these up into separate, nicely named tests. When something fails, it makes it easier to tell what went wrong. I defer to @jph00 on the patterns he prefers.

tests/test_core.py
```diff
@@ -0,0 +1,39 @@
+from fastai.core import partition
+import pytest
```
mcskinner: I've put this import as the first line by convention, but I'm not gonna insist on that vs just mention it :)

tests/test_core.py
NEIA20: sure i don't mind breaking the test up into smaller pieces. The primary difference among the examples in `test_partition_functionality` is the `sz` param. So breaking `test_partition_functionality` up into sub-tests would look something like:

```
def test_partition_with_sz_2():
  sz = 2
  a = [1,2,3,4,5]
  ex = [[1,2],[3,4],[5]]
  result = partition(a, sz)
  assert len(result) == len(ex)
  assert all([a == b for a, b in zip(result, ex)])

```

My main motivation behind so many `sz` variations was to provide several examples so that someone wanting to use this function would know how it would behave.

Maybe I should just have one testing example here and then list the rest in a doc string where the function is defined. What do you think?

tests/test_core.py
NEIA20: no worries i don't mind fixing that

tests/test_core.py
mcskinner: Go ahead and leave as is for now. I don't want to pretend I have any authority to be prescriptive :)

One suggestion I have elsewhere is to consider [pytest parameterizing](https://docs.pytest.org/en/latest/example/parametrize.html), but we can wait for @jph00 to decide the preference.

PR #444, opened 2018-05-04
Adoc: documentation for core.py and sgdr.py

cuddle-cuddle:
Nothing new added. Purely documentation.


fastai/sgdr.py
```diff
@@ -490,6 +542,9 @@ def update(self):
 
 
 class OptimScheduler(LossRecorder):
+    '''
```
jph00: Could you put all the one-liner docs on a single line? (i.e. put the quotes on the same line)

fastai/sgdr.py
```diff
@@ -248,6 +291,10 @@ def calc_mom(self):
         return res
 
 class CircularLR_beta(LR_Updater):
+    '''
```
jph00: Probably best to remove any docs you're not sure of

jph00:
Many thanks!

PR #445, opened 2018-05-04
Added conda-forge channel only for jup extensions

sudarshan85:
According to the issue listed [here](https://github.com/conda/conda/issues/2800) in the conda repo there is an option to add a channel per dependency. Now that issue is not closed yet but there is no activity since Nov 2017. I changed the env files to use conda-forge only for the jupyter extensions and it worked. I was able to see all the extensions listed. I have changed all the env files to use conda-forge for jup extensions.


jph00:
Good find!

Issue #386, opened 2018-04-23
ImageClassifierData.from_csv does not select the rows corresponding to val_idxs even if explicitly passed.


sidchat04:
The method parse_csv_labels returns a sorted list of file names, hence the indices passed using val_idxs select the wrong rows when creating the validation data.

jph00:
Could you provide a simple failing test case?

sidchat04:
Hi @jph00, I can provide a test case in a sample notebook, or even try to put in a pull request for a test case in code (though I don't have a lot of experience there). What would be the preferred way?
I am actually also working on a simple fix for this issue, and will try to put in a pull request for that soon.

jph00:
Ah well in that case, putting a test in the PR would be perfect. We have
a 'tests' directory now, so you could put it there, if not inconvenient.
Otherwise a notebook as a gist with a link in the PR would be fine.
--
Jeremy Howard


On Tue, Apr 24, 2018, at 8:05 AM, Sidharth Chaturvedi wrote:
> Hi @jph00[1], I can provide a test case in a sample notebook, or even
> try to put in a pull request for a test case in code (though I don't
> have a lot of experience there). What would be the preferred way?>  I am actually also working on a simple fix for this issue, and will
>  try to put in a pull request for that soon.> — You are receiving this because you were mentioned. Reply to this
> email directly, view it on GitHub[2], or mute the thread[3].> 


Links:

  1. https://github.com/jph00
  2. https://github.com/fastai/fastai/issues/386#issuecomment-383967183

radekosmulski:
I think I ran into the same issue. By the time our filenames and targets arrive at `split_by_idx` they are sorted. They are split based on idxs that correspond to the original order in the csv and this produces incorrect results.

radekosmulski:
This happens in dataset.py here:

```
   115  def parse_csv_labels(fn, skip_header=True, cat_separator = ' '):
   116      """Parse filenames and label sets from a CSV file.
   117
   118      This method expects that the csv file at path :fn: has two columns. If it
   119      has a header, :skip_header: should be set to True. The labels in the
   120      label set are expected to be space separated.
   121
   122      Arguments:
   123          fn: Path to a CSV file.
   124          skip_header: A boolean flag indicating whether to skip the header.
   125
   126      Returns:
   127          a four-tuple of (
   128              sorted image filenames,
   129              a dictionary of filenames and corresponding labels,
   130              a sorted set of unique labels,
[...]

PR #447, opened 2018-05-05
Proposed fix for  #386 ImageClassifierData.from_csv does not select the rows corresponding to val_idxs even if explicitly passed.

sidchat04:
Fix for #386 

- First commit adds a regression test that can be helpful in demonstrating the issue. I have tried following the test guidelines, but it's still my first time writing tests in Python using pytest, so please let me know if anything needs to be changed.
 - Second commit attempts to fix the issue by introducing another parameter (sort_fnames) in methods used by ImageClassifierData.from_csv, that indicate whether the returned file names need to be sorted. The parameter is passed in as False, if the list of validation indexes has been provided. It is True by default.

Please let me know if any changes are required. Thanks.

### DIFF
--- a/fastai/dataset.py
+++ b/fastai/dataset.py
@@ -112,7 +112,7 @@ def parse_csv_labels(fn, skip_header=True, cat_separator = ' '):
 
     Returns:
         a four-tuple of (
-            sorted image filenames,
+            image filenames,
             a dictionary of filenames and corresponding labels,
             a sorted set of unique labels,
             a dictionary of labels to their corresponding index, which will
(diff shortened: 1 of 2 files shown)


PiotrCzapla:
@sidchat04 I'm having a hard time understanding the issue. Do you mean that the order of the CSV file is not being kept?

sidchat04:
@PiotrCzapla The problem is this: If I pass [2, 3] as the val_idxs parameter in from_csv method, I want rows with index 2 and 3 from the passed CSV to be chosen as my validation data. But since the list of file names has already been sorted by the time validation data is picked, the rows chosen are different.

This does not create a problem in cases when we are randomly generating the validation indexes. But in some cases, the validation data is also provided along with the problem (example is the Kaggle competition https://www.kaggle.com/c/imaterialist-challenge-furniture-2018/data).

PiotrCzapla:
I see.  what I dont understand is why are we sorting the CSV file in the first place as the CSV has order that does not change. Maybe this is something coming from code where files are being read from file system and the order can change from time to time, @jph00 ?

jph00:
Thanks this is very helpful! I must admit I'm not wondering why we're sorting the csv at all. It does seem like leaving it in the original order is most likely to be what's desired. Does anyone know of a reason that sorting is a good idea?

sidchat04:
Possibly to be able to view the file names conveniently, but I am not so sure. I can easily undo my changes in the PR, and then remove the sorting.

sidchat04:
@jph00 My apologies for not following up on this. I have now updated this PR to remove sorting of the filenames in CSV altogether. Please let me know if any further changes are required.

PR #450, opened 2018-05-06
Redefine transfroms as set of partially applied functions

PiotrCzapla:
This is refined version of the previous proposal #312. I've managed to get rid of the TfmType and is_y by converting the transformations to set of applied functions whose parameters can be changed during execution.

We keep the concept of `determ`, `apply_transform` from #312. but we make`tfms.determ()` returning a partially applied function (almost as we return wrapper to make the printout nicer).  
The change is **backward compatible** to large extend.

This let us express the `sz_y` (the reason for`is_y` existance) as follows:
     
```py
def ap_t(t, x, y):
    nx = t(x)
    ny = t(y, sz=sz_y) # here the default value is set to sz_y
    return nx,ny
```

The coords transformation (one of reasons for `TfmType`) looks like this:

```py
def ap_t(t, x, y): # this is how we can express coords transformation for y
    nx = t(x)
    ny = transform_coords(t, y, x.shape)
    return nx,ny
```

The mask transformation (another reason for `TfmType`) can be expressed as follows:
```py
def ap_t(t, x, y): # this is how we can express coords transformation for y
    nx = t(x)
    ny = t(y, **TfmParams.CLASS)
    return nx,ny
```

I've also add an inspection mechanism to transforms to make it easier to reason about random transformations, here is how it looks like

```py
>>> tfms[0]
Trasnforms(...)
# with ComposedTransform(tfms=[
  RandomScale.do(_, sz=16, interpolation=3, zoom=?),
  AddPadding.do(_, pad=2, pad_mode=2),
  RandomRotate.do(_, pad_mode=2, interpolation=3, rp=?, rdeg=?),
  RandomLighting.do(_, b=?, c=?),
  RandomDihedral.do(_, rot_times=?, do_flip=?),
  NoCrop.do(_, sz=16, interpolation=3),
  Normalize.do(_, mean=[0.5 0.5 0.5], stddev=[0.5 0.5 0.5]),
  ChannelOrder.do(_)
])

>>> tfms[0].determ()
ComposedTransform(tfms=[
  RandomScale.do(_, sz=16, interpolation=3, zoom=1.0),
  AddPadding.do(_, pad=2, pad_mode=2),
  RandomRotate.do(_, pad_mode=2, interpolation=3, rp=False, rdeg=1.3287164035703682),
  RandomLighting.do(_, b=-0.02255752415182746, c=0.008637519508592914),
  RandomDihedral.do(_, rot_times=2, do_flip=True),
  NoCrop.do(_, sz=16, interpolation=3),
  Normalize.do(_, mean=[0.5 0.5 0.5], stddev=[0.5 0.5 0.5]),
  ChannelOrder.do(_)
])
```

### DIFF
--- a/fastai/transforms.py
+++ b/fastai/transforms.py
@@ -2,24 +2,24 @@ from .imports import *
 from .layer_optimizer import *
 from enum import IntEnum
 
-def scale_min(im, targ, interpolation=cv2.INTER_AREA):
+def scale_min(im, sz, interpolation=cv2.INTER_AREA):
     """ Scales the image so that the smallest axis is of size targ.
 
     Arguments:
         im (array): image
-        targ (int): target size
+        sz (int): target size
     """
     r,c,*_ = im.shape
-    ratio = targ/min(r,c)
-    sz = (scale_to(c, ratio, targ), scale_to(r, ratio, targ))
-    return cv2.resize(im, sz, interpolation=interpolation)
+    ratio = sz/min(r,c)
+    dsize = (scale_to(c, ratio, sz), scale_to(r, ratio, sz))
+    return cv2.resize(im, dsize, interpolation=interpolation)
 
-def zoom_cv(x,z):
+def zoom_cv(im, zoom):
     '''zooms the center of image x, by a factor of z+1 while retaining the origal image size and proportion. '''
-    if z==0: return x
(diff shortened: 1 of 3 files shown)


jph00:
There's a new transform pipeline now available in fastai_v1

PR #451, opened 2018-05-06
Adds tests for `split_by_idxs` and adds `test_core` folder

NEIA20:
This PR: 
- adds test for `split_by_idxs`
- adds index validation for `split_by_idxs`
- "dry"s up the `partition` tests
- creates a `test_core` folder for `core.py` testing files (the one testing file was already getting long with just the tests from 2 functions)


fastai/core.py
```diff
@@ -53,6 +53,8 @@ def noop(*args, **kwargs): return
 def split_by_idxs(seq, idxs):
     last = 0
     for idx in idxs:
+        if idx >= len(seq) or idx < len(seq) * -1:
```
jph00: For future reference, you can simply this to (IIRC):

```
not (-len(seq) <= idx < len(seq))
```

jph00:
Are you sure we want to split the tests up into separate files? Seems to me like that's going to add complexity in finding tests and understanding modules - what's the downside of having large test files? Pretty much all modern editors support folding and/or visual file maps, so navigating large files is pretty convenient.

fastai/core.py
NEIA20: you're right i like that better - the condition is more obvious

NEIA20:
That's funny cuz the reason why I broke them into separate files is because I thought it would be quicker to find the test(s) that you were looking for 😂 

I might be biased here because I think I personally find it less overwhelming when code is broken up into separate files. The compartmentalization is also helpful - e.g. I know that in `test_split_by_idxs.py` I'm only looking at code that pertains to tests for a function called `split_by_idxs`

It seems like it might be subjective, and given that I should probably stick with the existing patterns in the codebase? (i.e. keeping them in one file)

jph00:
Many thanks for the consideration - yes I think keeping in one file would be great. FYI if you're not familiar with code folding you should check it out and see if your editor can have it turned on by default (and also try to learn the keyboard shortcuts for increasing and decreasing folding level for a file) - hopefully things aren't too overwhelming when you can easily collapse the file into a handle of lines.

NEIA20:
No problem at all, I put the core tests back in one file and updated the if condition in `split_by_idxs`

I'm familiar with code folding but I don't use it enough - the keyboard shortcuts sound like a really good way to go 👍

PR #458, opened 2018-05-10
import fixes for pix2pix part of cgan

tensoralex:
fixes to aligned_dataset.py, single_dataset.py, pix2pix_model.py, test_model.py
to allow to utilize cyclegan notebook in pix2pix mode


jph00:
Many thanks @tensoralex ! It would be great to do a PR to the original cyclegan project that includes all the relative module path fixes that allow the module to be used from a notebook - that way we could remove the cgan folder entirely and use the source repo directly (e.g as a submodule). Any interest in giving that a go?

tensoralex:
Sure @jph00 , let me take a look.

PR #466, opened 2018-05-11
Deleted the extra return keyword from line 45 of core.py

anthonyjclark:
I pulled the fastai repo and noticed an extra "return" keyword in one function (caused an import error).

PR #469, opened 2018-05-11
core.py syntax bug fix

phillynch7:


Eric-Arellano:
Hi, looks like someone already addressed this with PR #466.

PR #476, opened 2018-05-16
Optimize wd_loss

sgugger:
1. wd_loss is now done without waving a copy of the model, inside the stepper, applying the weight_decay after computing the gradient but before doing the step in the optimizer.
2. The training API now supports datasets of different lengths/batchsizes.

### DIFF
--- a/fastai/layer_optimizer.py
+++ b/fastai/layer_optimizer.py
@@ -38,10 +38,18 @@ class LayerOptimizer():
         set_lrs(self.opt, lrs)
         self.lrs=lrs
 
+    def set_wds_out(self, wds):
+        if not isinstance(wds, Iterable): wds=[wds]
+        if len(wds)==1: wds=wds*len(self.layer_groups)
+        set_wds_out(self.opt, wds)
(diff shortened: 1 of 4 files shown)


fastai/layer_optimizer.py
```diff
@@ -38,10 +38,18 @@ def set_lrs(self, lrs):
         set_lrs(self.opt, lrs)
         self.lrs=lrs
 
+    def set_wds_out(self, wds):
+        if not isinstance(wds, Iterable): wds=[wds]
```
jph00: Maybe instead of

```
if not isinstance(wds, Iterable): wds=[wds]
if len(wds)==1: wds=wds*len(self.layer_groups)
```

you could make a little function:

```
wds = listify(wds, self.layer_groups)
```

...where the second param is used directly if it's an int, or we use its `len` if it's listy? I think we've used this pattern in quite a few places.

sgugger: Yeah, it's used quite a few times and you're right, a helper function would make the code cleaner.

PR #493, opened 2018-05-23
Clean-up

sgugger:
- refactors code with listify
- removes the warning for kaiming init in pytorch 0.4

### DIFF
--- a/fastai/sgdr.py
+++ b/fastai/sgdr.py
@@ -514,8 +514,6 @@ class TrainingPhase():
         self.layer_opt.set_mom(start_mom)
         if self.beta is not None: self.layer_opt.set_beta(self.beta)
         if self.wds is not None:
-            if not isinstance(self.wds, Iterable): self.wds=[self.wds]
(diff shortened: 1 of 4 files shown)


fastai/sgdr.py
```diff
@@ -514,8 +514,6 @@ def phase_begin(self, layer_opt, nb_batches):
         self.layer_opt.set_mom(start_mom)
         if self.beta is not None: self.layer_opt.set_beta(self.beta)
         if self.wds is not None:
-            if not isinstance(self.wds, Iterable): self.wds=[self.wds]
-            if len(self.wds)==1: self.wds=self.wds*len(self.layer_opt.layer_groups)
```
antishok: Is this not missing the replacement line? `self.wds = listify(self.wds, self.layer_opt.layer_groups)`

sgugger: There are listified inside layer_opt with the next lines so I didn't bother doing it here.

PR #501, opened 2018-05-25
 fix pytorch import error in CPU only environment 

lchen23:
After installing the CPU only version of fastai library by running `conda env update -f environment-cpu.yml`. The pytorch library still fails to import and reports this error: 
`C:\Program Files\Anaconda3\envs\fastai\lib\site-packages\torch\__init__.py in <module>()
     74     pass
     75 
---> 76 from torch._C import *
     77 
     78 __all__ += [name for name in dir(_C)
ImportError: DLL load failed: The specified module could not be found.`

I found a simple solution by running the command `conda install -c peterjc123 pytorch-cpu`, which installs a cpu-version of pytorch. The original solution is found here: https://github.com/peterjc123/pytorch-scripts. 

I changed the readme file to report this big-fix and hope it will be helpful to others who wish to test cpu-only version of the fastai library.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,8 +34,7 @@ Use this if you do not have an NVidia GPU. Note you are encouraged to use Papers
 
 `conda env update -f environment-cpu.yml`
 
-* Anytime the instructions say to activate the Python environment, run `conda activate fastai-cpu` or `source activate fastai-cpu`.
-
+- Anytime the instructions say to activate the Python environment, run `conda activate fastai-cpu` or `source activate fastai-cpu`.
 
 ## To update
 1. Update code: `git pull`
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ Use this if you do not have an NVidia GPU. Note you are encouraged to use Papers
 
 `conda env update -f environment-cpu.yml`
 
-- Anytime the instructions say to activate the Python environment, run `conda activate fastai-cpu` or `source activate fastai-cpu`.
+Anytime the instructions say to activate the Python environment, run `conda activate fastai-cpu` or `source activate fastai-cpu`.
 
 ## To update
 1. Update code: `git pull`
--- a/README.md
+++ b/README.md
@@ -35,9 +35,7 @@ Use this if you do not have an NVidia GPU. Note you are encouraged to use Papers
 `conda env update -f environment-cpu.yml`
 
 * Anytime the instructions say to activate the Python environment, run `conda activate fastai-cpu` or `source activate fastai-cpu`.
-* If the fastai library fails to import and reports this error: "ImportError: DLL load failed: The specified module could not be found", run this commond in conda:
 
-`conda install -c peterjc123 pytorch-cpu`
 
 ## To update
 1. Update code: `git pull`
--- a/environment-cpu.yml
+++ b/environment-cpu.yml
@@ -22,6 +22,7 @@ dependencies:
 - expat
 - html5lib
 - icu
+- defaults::intel-openmp
 - ipykernel
 - ipython
 - ipython_genutils


jph00:
I believe that's just for Windows - in which case perhaps mention that too?

lchen23:
Yes, I believe so. I'm using Windows 7 and having the pytorch import problem. 
The original author also mentioned the purpose of his package is  "It is a repo that contains scripts that makes using PyTorch on Windows easier."  So, it's probably a windows specific problem.

peterjc123:
@lchen23 Is there a reason that you could not use the official one, which is `conda install -c pytorch pytorch-cpu`?

lchen23:
@peterjc123 I just tried `conda install -c pytorch pytorch-cpu`, which installed pytorch-cpu 0.4.0.  And I got a new error: `ImportError: DLL load failed: The operating system cannot run %1`. My OS is Windows 7 pro. Maybe pytorch-cpu 0.4.0 could work for windows 10, I'm not sure. The pytorch-cpu 0.3.1 from `conda install -c peterjc123 pytorch-cpu` still works for me though.

peterjc123:
@lchen23 Would you please provide us with some logs to help us, such as the logs with `python -v -c 'import torch'` and the result of the following code:
```python
# pip install psutil first
import psutil, os
p = psutil.Process( os.getpid() )
for dll in p.memory_maps():
  print(dll.path)
```
And you could try `conda install mkl mkl-fft intel-openmp numpy`.
I've tested 0.4.0 CPU on Windows Server 2008 R2, so it should work fine on Windows 7. I just don't know what's wrong there.

lchen23:
screenshot-2018-5-29 untitled
screenshot-2018-5-29 untitled 1
screenshot-2018-5-29 untitled 2

peterjc123:
@lchen23 Thanks! Looks like `numpy` is correctly loaded but the others things are not.
Since I don't see intel-openmp there, I guess it may be the reason. Could you please try `conda install intel-openmp`?

peterjc123:
@lchen23 Hi, would you please try `conda update conda`?

lchen23:
@peterjc123 Here's the list of packages I installed in conda https://github.com/lchen23/pytorch/blob/master/import_test.ipynb . I find I already have all the packages you mentioned and they are up to date. Just the pytorch-cpu doesn't work.

peterjc123:
@lchen23 Thanks for your list. I'll try to build a new environment and try to reproduce your issue.

lchen23:
@peterjc123 Thanks for your help. To create my env, I basically just downloaded the fastai lib, `cd fastai` and ran `conda env update -f environment-cpu.yml`. My OS is windows 7 pro. Hope this info is helpful for you to reproduce the issue.

peterjc123:
@lchen23 Hi, I've figured out the solution. The problem is on `intel-openmp`. You have to install the one from defaults channel, so the command to fix is `conda install -c defaults intel-openmp -f`.

jph00:
OK so perhaps we need an update to environment-cpu and also to these docs? PRs welcome! :)

lchen23:
@peterjc123 Your fix works 100%! @jph00 I'm trying to create a sharable env.yml file by `conda env export > environment-cpu.yml`. But it gives me this file:
`name: fastai-cpu
channels:
  - defaults
prefix: C:\ProgramData\Anaconda3\envs\fastai-cpu\envs\fastai-cpu`
The file points to a specific location in my PC for env settings. How can I make it to export all installed package info instead of pointing to a folder in my pc?

jph00:
Maybe easiest to just manually edit the yaml file that's already there?

lchen23:
I manually added `- defaults::intel-openmp` and changed `- pytorch<0.4` to `- pytorch=0.4` in environment-cpu.yml.  I used the new file to recreate a new env and it worked OK. See the new environment-cpu.yml attached.
@jph00 if you are OK with the change, I can close this PR and open a new PR to update the  environment-cpu.yml file
[environment-cpu.txt](https://github.com/fastai/fastai/files/2060523/environment-cpu.txt)

jph00:
We aren't supporting pytorch 0.4 yet, so that shouldn't be changed

PR #509, opened 2018-05-27
Adding F1 score to fastai/courses/dl1/planet.py

MicPie:
Hello,

I recently needed the F1 score for a project and implemented it like the F2 score in the planet.py file including optimal threshold finding.

Maybe this is also interesting for others.

Best regards
Michael


jph00:
Thanks for the PR - not sure it's something I want to add however.

PR #510, opened 2018-05-27
Make it easy to plot training & validation loss on same figure

ashtonsix:
Record epochs to align plots.

plot showing training & validation loss on same figure


ashtonsix:
Oh, & I just added times because it seemed like a good idea

PR #511, opened 2018-05-27
Change `nc` to `cl`

bkj:
Is this correct?  Seems probably, but not obvious from the variable names.


sebastianruder:
Yep, you're right. There are some errors in the documentation. We'll fix those over the next days.

PR #520, opened 2018-05-31
Reformat code to look more simple.

claforte:
Add a bit of whitespace to make the first pattern obvious.
Initialize cls() arguments outside, to make the second pattern obvious.


claforte:
Hi Jeremy, two small changes that I think make the code easier to understand for newcomers.

jph00:
Yup the consistency is nice :)

PR #521, opened 2018-05-31
Docstring for ModelData

claforte:


jph00:
The key thing about ModelData is that it encapsulates training, validation, and (optionally) test dataloaders, so that a Learner has all the data it needs. Might be worth mentioning that.

PR #522, opened 2018-05-31
Fix docstring typos/grammar

claforte:
Also standardize triple-quotes to what seems like the most commonly-used pattern.


jph00:
Many thanks

PR #546, opened 2018-06-06
Fix __init__

slee5777:
def __init__ is missing "bias=False" after added qrnn.


slee5777:
The following codes cannot be run without modified the __init__.  I tested it on my language model.

learner = md.get_model(opt_fn, em_sz, nh, nl, dropouti=drops[0], dropout=drops[1], 
                      wdrop=drops[2], dropoute=drops[3], dropouth=drops[4], qrnn=True)

PR #550, opened 2018-06-07
Added the spacy library to the requirements files.

peristeri:


dmh43:
Why isn't this also in `setup.py` under `install_requires`?

peristeri:
An oversight on my part.

PR #553, opened 2018-06-09
planet.py - modify def f2, so as to convert preds and targs to bumpy array before using sklearn

datafool:
When running on GPU, both preds and targs will be on GPU and are tensor. They need to be brought to CPU and converted to numpy before using any sklearn function. This pull request fixes the bug in planet/f2


datafool:
This error is getting resolved after updating condo environment. So, closing this pull request!

PR #565, opened 2018-06-16
fastai/model.py (fit): enable global metrics on test dataloader

quocpeyrot:
(e.g. whole segmentation DICE every n-epoch)

This is especially useful when training FCN networks with only partial sampling in training/validation set. In such cases, you'd still want to get real (whole/full subject) metrics from time to time.


sgugger:
Sorry it took us so long to reply! Jeremy has been very busy with personal events and I've just joined the team.
I like the way you coded this... but using the test set too much is generally a bad practice so we'd like to avoid having inside the library. I'd definitely recommend posting a notebook on the [forum](http://forums.fast.ai/) with an example of use, since there are some situations where it can be useful. As long as you create your own fit function by copy-pasting the code you suggested, it will be the one used by any learner object.

PR #567, opened 2018-06-18
Removed hard-coded paths and unused arguments, cleaned up code, updated README

sebastianruder:
Removed hard-coded paths and unused arguments and cleaned up code.

I also renamed files to make their use clearer and updated the README with the new filenames and explanations of all arguments used in the scripts.

### DIFF
--- a/courses/dl2/imdb_scripts/README.md
+++ b/courses/dl2/imdb_scripts/README.md
@@ -19,7 +19,7 @@ and `val.csv` is 10% the size of `train.csv`.
 
 Run `create_toks.py` to tokenize the input texts.
 
-Example command: `python create_toks.py imdb`
+Example command: `python create_toks.py data/imdb`
 
 Usage:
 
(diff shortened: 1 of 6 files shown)


courses/dl2/imdb_scripts/README.md
```diff
@@ -55,65 +55,96 @@ tok2id.py --prefix PREFIX [--max-vocab MAX_VOCAB] [--min-freq MIN_FREQ]
 
 ### 3. Fine-tune the LM
 
-First run `train_tri_wt.py` to create a pre-trained language model using WikiText-103 (or whatever base corpus you prefer)
-or use one of the pre-trained models.
+Before fine-tuning the language model, you can run `pretrain_lm.py` to create a 
+pre-trained language model using WikiText-103 (or whatever base corpus you prefer).
 
 Example command: `python train_tri_wt.py data/wiki/de/ 0 --lr 1e-3 --cl 12`
```
andehr: I think this line should have `python pretrain_lm.py` instead of`python train_tri_wt.py`

sebastianruder: Thanks for pointing that out! Missed that! :)

PR #574, opened 2018-06-20
Add jupyterlab and official kaggle API

KaiLicht:
JupyterLab is now [ready for daily use](https://blog.jupyter.org/jupyterlab-is-ready-for-users-5a6f039b8906) and will eventually replace Jupyter Notebooks in the future. It has full support for *.ipynb files and gives the alternative to work in a more REPL-Style workflow.
The official kaggle API is very convenient and stable in comparison to the unofficial kaggle-cli. Both packages together are adding less than 20 MB.

### DIFF
--- a/environment-cpu.yml
+++ b/environment-cpu.yml
@@ -31,6 +31,7 @@ dependencies:
 - jinja2
 - jsonschema
 - jupyter
+- jupyterlab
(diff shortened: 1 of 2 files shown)


jph00:
We'll likely add these as part of the next course - but not yet.

PR #589, opened 2018-06-28
Flatten target values array to feed into model

eisenjulian:
RNN_Learner users cross-entropy loss that expects a single label for each sample, so this cannot work for multi-label classification out of the box, furthermore, we need to flatten the data to prevent mismatched shape errors.


eisenjulian:
FYI @sebastianruder

PR #607, opened 2018-07-05
Add Loss Visualization

WiraDKP:
This feature helps to provide a runtime evaluation of the `trn_loss` and `val_loss`.
It is activated by stating `visualize=True` when using the fit function. For example:
>`learn.fit(lr, 1, cycle_len=15, use_clr=(20,10), visualize=True)`

Here's how it looks.
image

The notation is quite intuitive:
▼ means the loss is lower than previous epoch
△ means the loss is higher than previous epoch


reshamas:
I love this!  👍

PR #611, opened 2018-07-06
align with new API for proc_df which is now 4 args

stas00:
fix the notebooks to use the latest API


stas00:
Thank you for merging, @sgugger

Issue #618, opened 2018-07-12
Making resizing of Image more efficient


karanchahal:
The code for Image Resizing as of right now is as follows:

```python
if is_y: return scale_min(x, self.sz_y, cv2.INTER_AREA if self.tfm_y == TfmType.PIXEL else cv2.INTER_NEAREST)
else   : return scale_min(x, self.sz,   cv2.INTER_AREA   )
```

I was reading up on Open CV documentation [here](https://docs.opencv.org/2.4/modules/imgproc/doc/geometric_transformations.html?highlight=resize#resize) , more specifically this line.

_To shrink an image, it will generally look best with CV_INTER_AREA interpolation, whereas to enlarge an image, it will generally look best with CV_INTER_CUBIC(slow) or CV_INTER_LINEAR (faster but still looks OK)._

The code for Fast.ai doesn't support the best method for enlarging an image, that is, use ```CV_INTER_LINEAR```.

Do you think it would be a good idea to change that code , so that when we want to shrink the size of image , we continue to use ```cv2.INTER_AREA```, but when we want to enlarge an image we use ```cv2.INTER_LINEAR``` ?

I would be happy to submit a pull request for the change.

sgugger:
This function is almost all the time used to reduce the size of images, which is why the default is cv2.INTER_AREA. A simple PR to define an interpolation depending on the target size wouldn't hurt though.

karanchahal:
👍

PR #619, opened 2018-07-12
Updates image scaler to be more efficient

karanchahal:
Uses a better mechanism to change scale of image if it's expanding. CV_INTER_LINEAR  is a better choice than CV_INTER_AREA for expanding an image. CV_INTER_AREA is better for shrinking an image.

References #618

### DIFF
--- a/fastai/transforms.py
+++ b/fastai/transforms.py
@@ -365,8 +365,7 @@ class Scale(CoordTransform):
 
     def do_transform(self, x, is_y):
         h,w,*_ = x.shape
-        intpr = cv2.INTER_AREA
-        if(self.sz_y < min(h, w)) : intpr = cv2.INTER_LINEAR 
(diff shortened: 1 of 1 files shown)


karanchahal:
@sgugger @jph00  Any updates on the review for this ?

sgugger:
Sorry, I thought I left a comment a few days ago, but I forgot to press the last button. There is a little bug I think.

karanchahal:
Yes, you are right, So sorry for this bug. I have corrected and pushed the code with the correct variable name.

karanchahal:
Cheers ! Is there anything else you would like someone to take a look at? I'm going through the fast.ai codebase as it is. Would be happy to submit a PR for some fixes.

sgugger:
We're in the process of rewriting the version 1 of this library, so in the meantime we only maintain this one with minor fixes. Stay tuned for the new version in October!

PR #623, opened 2018-07-14
add missing lesson1-breeds.ipynb notebook.

stas00:
In the first few DL lessons Jeremy often works with dog breeds notebook (lesson1-breeds.ipynb), but says he can't give it us since the competition is unfinished and he doesn't want us to cheat.

The competition has been finished, so let's make this notebook available to all so it's easier to follow the video lessons.

I found a copy of it in this repo https://github.com/tanakatsu/fastai-course1-v2. I can't tell whether it is identical to Jeremy's copy and how it was acquired. But it's probably better to have it in any form than none.

I tested it that it runs to a successful completion as is.

Here are the changes I have applied to lesson1-breeds.ipynb found under the github repo mentioned above:

- cleaned up the corrections that were made due to the API changes made since the video lessons.
- added a few notes where to find data needed for that notebook to run.
- edited the last section to remove the non-working code and clarify that there are 2 methods.
- removed the hardcoded section numbers (jupyter generates those automatically).

### DIFF
--- a/courses/dl1/lesson1-breeds.ipynb
+++ b/courses/dl1/lesson1-breeds.ipynb
@@ -1,0 +1,5952 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Dogs breeds\n",
+    "\n",
+    "https://youtu.be/JNxcznsrRb8?t=1h31m8s"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "%reload_ext autoreload\n",
+    "%autoreload 2\n",
+    "%matplotlib inline"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from fastai.imports import *\n",
+    "from fastai.torch_imports import *\n",
+    "from fastai.transforms import *\n",
+    "from fastai.conv_learner import *\n",
+    "from fastai.model import *\n",
+    "from fastai.dataset import *\n",
+    "from fastai.sgdr import *\n",
+    "from fastai.plots import *"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 3,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "torch.cuda.set_device(0)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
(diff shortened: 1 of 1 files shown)


jph00:
Thanks Stas - that's a good idea. I kinda hope that people will try to create this themselves from scratch, but it's nice to have a complete example they can look at. The notebook you based this on seems to have been developed independently. I think you should be able to get much better performance by using `learner.bn_freeze()` - feel free to add more PRs to this if you improve it in the future.

stas00:
@jph00, do you have the original notebook that you used while teaching (it was called tmp_something...breed)? It'd be the easiest to use that instead (I can apply whatever API changes were made since then).

I tried following your suggestion, and unless I did something wrong, adding learner.bn_freeze(True) didn't seem to make much of a difference. I tried w/ and w/o unfreezing the whole model and using differential lrs, it gives better accuracy sooner, but converges to the same ceiling.

jph00:
@stas00 I just commited my version of the nb to the repo as lesson1-breeds2.ipynb. Perhaps you can take a look and combine the best bits of the two notebooks, and do a PR that removes the 2nd notebook and updates the first appropriately?

stas00:
I made quite a few beginner attempts to beat 0.926 accuracy of the first notebook, borrowing from your version, but wasn't able to. Your version as is was incomplete (wds undefined) and blowing the gradients into the sky, but after tweaking things to keep it contained it was giving worse results. 

I also had to drastically reduce `bs` as I only have 8GB GPU RAM, and it wanted a lot more for retraining the full network. So it's possible that it's a hardware limitation on my side.

Your notebook used` arch=dn161`, the one I found uses `resnext101_64`. I looked at the forums at it appears that [93% accuracy](http://forums.fast.ai/t/dog-breed-identification-challenge/7464/316) is the max one could get from `resnext101_64` on this dataset. it was suggested `nasnet` could get it to 95%, but the poster didn't leave specific details on how he accomplished that, other than [this](http://forums.fast.ai/t/dog-breed-identification-challenge/7464/322):

> One of differences between nasnet and resnext is that for nasnet you need to define:

```
def nasnet(pre): return nasnetalarge(pretrained = 'imagenet' if pre else None)
model_features[nasnet]=4032*2
stats = ([0.5, 0.5, 0.5], [0.5, 0.5, 0.5])
tfms = tfms_from_stats(stats, sz, aug_tfms=transforms_side_on, max_zoom=1.1)
```

So, I'm not sure how to proceed from here. Any guidance is welcome. I'm also fine leaving the found notebook as it is for now, and probably deleting yours as it's incomplete, unless you'd have a look.

Thanks.

jph00:
Ah well perhaps I'm mis-remembering, or students have gotten
better results than I did. I've removed the 2nd notebook now.
Thanks for checking.

stas00:
Thank you, Jeremy

PR #629, opened 2018-07-15
Observational Weights - Added Functionality

pford221:
Updates to model.py, dataset.py, and column_data.py to incorporate observational weights (i.e. each record has a different influence/weight on the loss function) for structured data. I don't believe observational weights are needed that much outside of structured data so this PR doesn't address. In addition to observation weights being useful in its own right, this bring fastai more in-line with Keras which supports observation weights.

There is also an observational_wgts.ipynb in the tutorials section to show why and how to use the new observation weight functionality. I wasn't sure the best place to put this, so feel free to respond to let me know. I could certainly host this on my github account too.

The one aspect that I didn't feel was worthwhile to tackle is that observation weights impact the size of the gradients in a slightly nuanced way. Essentially, the larger the observation weights the larger will be the weighted sum of the gradients for a particular batch. But PyTorch divides the sum of the gradients by the total number of records in the batch. What we'd really want to to do is divide the weighted sum of the gradients by the sum of the observation weights in the batch and not the number of records. This difference will only matter if the observation weights are much larger or smaller than 1.0 on average. If so, the user should adjust the learning rate commensurately. In this case, using other fastai tools like lr_find become even more important.

### DIFF
--- a/fastai/column_data.py
+++ b/fastai/column_data.py
@@ -19,9 +19,24 @@ class PassthruDataset(Dataset):
         cols = [df[o] for o in cols_x+[col_y]]
         return cls(*cols, is_reg=is_reg, is_multi=is_multi)
 
+class PassthruDataset_wgts(Dataset):
+    def __init__(self,*args, is_reg=True, is_multi=False):
+        *xs,y,w=args
+        self.xs,self.y,self.w = xs,y,w
+        self.is_reg = is_reg
+        self.is_multi = is_multi
+
+    def __len__(self): return len(self.y)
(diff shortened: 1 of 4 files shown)


jph00:
I think this a is pretty great goal, although I think our current code-base doesn't really allow you to do this neatly - the approach you have here requires too much code, and doesn't generalize easily to other types of dataset (and observational weights are absolutely useful for images, nlp, etc too!)

So I think I'd rather try to tackle this in a more general way in our v1 rewrite - we've just started working on it at: https://github.com/fastai/fastai_v1 .

Many thanks for this contribution!

PR #637, opened 2018-07-18
expand asserts to give a meaningful message to the user

stas00:
until now it was throwing: AssertionError: without any message

for more details please see:
http://forums.fast.ai/t/lesson7-cam-mismatch-between-of-lrs-and-of-layer-groups/19425

### DIFF
--- a/fastai/layer_optimizer.py
+++ b/fastai/layer_optimizer.py
@@ -15,8 +15,8 @@ class LayerOptimizer():
         self.opt = opt_fn(self.opt_params())
 
     def opt_params(self):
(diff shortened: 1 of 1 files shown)


sgugger:
Good idea! Could you just modify your code to use python 3.6 string-formatting? Something like f'size mismatch, expected {len(self.layer_groups)} lrs but got {len(lrs)}'
Thanks.

stas00:
Thank you for your guidance, Sylvain. 

I have rewritten the patch to use f'' and also your suggested wording.

sgugger:
No worries, thanks for making the changes!

PR #641, opened 2018-07-19
Script to use classifier to predict against test

nlothian:
As discussed with @sebastianruder at http://forums.fast.ai/t/cant-replicate-ulmfit-validation-predictions/18477/22?u=nickl

### DIFF
--- a/courses/dl2/imdb_scripts/README.md
+++ b/courses/dl2/imdb_scripts/README.md
@@ -30,7 +30,7 @@ create_toks.py --dir-path DIR_PATH [--chunksize CHUNKSIZE] [--n-lbls N_LBLS] [--
 
 - `DIR_PATH`: the directory where your data is located
 - `CHUNKSIZE`: the size of the chunks when reading the files with pandas; use smaller sizes with less RAM
-- `LANG`: the language of your corpus. 
+- `LANG`: the language of your corpus.
 
 `train.csv` and `val.csv` files should be in `DIR_PATH`. The script will then save the
 training and test tokens and labels as arrays to binary files in NumPy format in a `tmp`
@@ -55,7 +55,7 @@ tok2id.py --prefix PREFIX [--max-vocab MAX_VOCAB] [--min-freq MIN_FREQ]
 
 ### 3. Fine-tune the LM
 
-Before fine-tuning the language model, you can run `pretrain_lm.py` to create a 
+Before fine-tuning the language model, you can run `pretrain_lm.py` to create a
 pre-trained language model using WikiText-103 (or whatever base corpus you prefer).
 
 Example command: `python pretrain_lm.py data/wiki/de/ 0 --lr 1e-3 --cl 12`
@@ -105,7 +105,7 @@ finetune_lm.py --dir-path DIR_PATH --pretrain-path PRETRAIN_PATH [--cuda-id CUDA
 - `USE_CLR`: whether to use slanted triangular learning rates (STLR) (`True` by default)
 - `USE_REGULAR_SCHEDULE`: whether to use a regular schedule (instead of STLR)
 - `USE_DISCRIMINATIVE`: whether to use discriminative fine-tuning (`True` by default)
-- `NOTRAIN`: whether to skip fine-tuning 
+- `NOTRAIN`: whether to skip fine-tuning
 - `JOINED`: whether to fine-tune the LM on the concatenation of training and validation data
 - `TRAIN_FILE_ID`: can be used to indicate different training files (e.g. to test training sizes)
 - `EARLY_STOPPING`: whether to use early stopping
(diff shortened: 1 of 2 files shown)


sebastianruder:
Hi Nick, thanks! This is great! I'm just on the way back from a conference. Will test the code once I'm back.

Varal7:
Good idea!
I think, you can just use argmax instead of softmax for inference

nlothian:
@Varal7 this is true. 

There's a whole set of questions on the forum about "what are these numbers which come back as classifier scores" and "why don't they add up to 1" etc. The answer is always softmax, and my thought was to include it to avoid all that confusion.

charlesaugdupont:
I'm getting the following error running this script:

`Traceback (most recent call last):
  File "predict_with_classifier.py", line 130, in <module>
    fire.Fire(predict_input)
  File "/home/charlesdupont1997/anaconda3/envs/fastai/lib/python3.6/site-packages/fire/core.py", line 127, in Fire
    component_trace = _Fire(component, args, context, name)
  File "/home/charlesdupont1997/anaconda3/envs/fastai/lib/python3.6/site-packages/fire/core.py", line 366, in _Fire
    component, remaining_args)
  File "/home/charlesdupont1997/anaconda3/envs/fastai/lib/python3.6/site-packages/fire/core.py", line 542, in _CallCallable
    result = fn(*varargs, **kwargs)
  File "predict_with_classifier.py", line 116, in predict_input
    stoi, model = load_model(itos_filename, trained_classifier_filename)
  File "predict_with_classifier.py", line 20, in load_model
    stoi = collections.defaultdict(lambda:0, {str(v):int(k) for k,v in enumerate(itos)})
TypeError: 'int' object is not iterable`

I am not sure why this is happening. Similar lines seem to work fine in my imdb.ipynb notebook.

The model passed is a clas_1.h5 file, and I am loading up itos.pkl a my int to string mapping

linbojin:
I think this code doesn't pad any "padding index" so the result will be different from which is computed by constructing DataLoader. 

For my own test:
```
encoded text is: [3, 4, 5, 6, 303, 10, 11, 2, 73, 24, 72, 27, 15, 32, 46, 22, 16, 2, 25, 7]
I 
result： [-0.74961,  0.71659]

encoded text for dataloader:
[1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 3, 4, 5, 6, 303, 10, 11, 2, 73, 24, 72, 27, 15, 32, 46, 22, 16, 2, 25, 7]
result : [ 0.31941, -0.15691]
```

@nlothian @sebastianruder  Any idea?

nlothian:
You could be right. 

I remember looking at the padding issues some, but when I was testing it I don't think I found any difference. There's no difference in behavior here between the scripts and the notebook is there?

I don't have the code to retest that, and if you are seeing a difference then it is likely you are correct. I'm sure a patch would be welcomed!

linbojin:
The problem is I can not determine how many 1s should i pad, and different 1s will give different results. It's confusing.

sebastianruder:
Yeah, this is quite weird. I didn't look into this before as I thought we were already ignoring padding using [`pack_padded_sequence`](https://pytorch.org/docs/stable/nn.html?highlight=pack_padded_sequence#torch.nn.utils.rnn.pack_padded_sequence) (see [here](https://towardsdatascience.com/taming-lstms-variable-sized-mini-batches-and-why-pytorch-is-good-for-your-health-61d35642972e) for an explanation).
In the code, the padding is done [here](https://github.com/fastai/fastai/blob/master/old/fastai/dataloader.py#L59) and the number of padding tokens is the difference of the current document length to the length of the largest document in the batch (see [here](https://github.com/fastai/fastai/blob/master/old/fastai/dataloader.py#L55)).
As a short-term measure, for padding individual examples at test time, you could thus use something like the difference to the average document length in the training data.
In the mid-term, we should use `pack_padded_sequence` and `pad_packed_sequence` as described [here](https://towardsdatascience.com/taming-lstms-variable-sized-mini-batches-and-why-pytorch-is-good-for-your-health-61d35642972e).
cc @jph00 @sgugger

ranihorev:
I created a PR to fix this issue. 
https://github.com/fastai/fastai/pull/882#issue-222047595

nicklothian-redmarker:
Just bumping this because I think it's pretty important. @ranihorev's proposed changes broke a lot of the language model, so there should be a better way.

@sebastianruder suggestion of the short term fix of using _difference to the average document length in the training data_ is difficult too, because that isn't actually recorded anywhere at the moment. 

I follow the `pack_padded_sequence` [post](https://towardsdatascience.com/taming-lstms-variable-sized-mini-batches-and-why-pytorch-is-good-for-your-health-61d35642972e) mentioned above up to trick 3, but I need to think about that some more. 

What is the approach in FastAI 1.0 here? Is it fixed somehow?

nlothian:
Just bumping this because I think it's pretty important. @ranihorev's proposed changes broke a lot of the language model, so there should be a better way.

@sebastianruder suggestion of the short term fix of using _difference to the average document length in the training data_ is difficult too, because that isn't actually recorded anywhere at the moment. 

I follow the `pack_padded_sequence` [post](https://towardsdatascience.com/taming-lstms-variable-sized-mini-batches-and-why-pytorch-is-good-for-your-health-61d35642972e) mentioned above up to trick 3, but I need to think about that some more. 

What is the approach in FastAI 1.0 here? Is it fixed somehow?

PR #643, opened 2018-07-21
more user-friendly cycle_save_name verification

stas00:
verify that when `learn.fit()'s cycle_save_name` arg is passed, one of `use_clr, use_clr_beta or cycle_len` args is passed too.

For the initial issue please see: http://forums.fast.ai/t/save-cycle/7779/4, plus by looking at the code I added 2 more arguments that will use cycle_save_name.

I suppose there is no test suite yet here, here is a little something half-baked I crafted:

```
def testme(cycle_save_name=None, use_clr=None, use_clr_beta=None, cycle_len=None):
    if cycle_save_name:
        assert use_clr or use_clr_beta or cycle_len, "cycle_save_name argument requires either of the following arguments use_clr, use_clr_beta, cycle_len"

# ok
testme()
testme(cycle_save_name='cycle-saved', cycle_len=1)
# not ok
testme(cycle_save_name='cycle-saved')
```

### DIFF
--- a/fastai/learner.py
+++ b/fastai/learner.py
@@ -155,7 +155,8 @@ class Learner():
                 the cycles. For an intuitive explanation, please see
                 https://github.com/fastai/fastai/blob/master/courses/dl1/lesson1.ipynb
 
-            cycle_save_name (str): use to save the weights at end of each cycle
+            cycle_save_name (str): use to save the weights at end of each cycle (requires
+                use_clr, use_clr_beta or cycle_len arg)
 
             best_save_name (str): use to save weights of best model during training.
 
@@ -195,6 +196,9 @@ class Learner():
         Returns:
             None
         """
+
+        if cycle_save_name:
+            assert use_clr or use_clr_beta or cycle_len, "cycle_save_name argument requires either of the following arguments use_clr, use_clr_beta, cycle_len"
 
         if callbacks is None: callbacks=[]
         if metrics is None: metrics=self.metrics
--- a/fastai/learner.py
+++ b/fastai/learner.py
@@ -219,12 +219,14 @@ class Learner():
             clr_div,cut_div = use_clr[:2]
             moms = use_clr[2:] if len(use_clr) > 2 else None
             cycle_end = self.get_cycle_end(cycle_save_name)
+            assert cycle_len, "use_clr requires cycle_len arg"
             self.sched = CircularLR(layer_opt, len(data.trn_dl)*cycle_len, on_cycle_end=cycle_end, div=clr_div, cut_div=cut_div,
                                     momentums=moms)
         elif use_clr_beta is not None:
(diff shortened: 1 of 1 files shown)


sgugger:
I think it just requires cyc_len: use_clr and use_clr_beta will also trow back an error if they don't get a cyc_len.

stas00:
I don't see any such asserts, but you're correct that they should assert, as `cycle_len` is being used without validating whether it has been passed:

```
def fit_gen(...cycle_len=None...
... no cycle_len checking... and then using it:
self.sched = CircularLR(layer_opt, len(data.trn_dl)*cycle_len
```

that's not the greatest code. If it's required, shouldn't the API tell the user so?

So perhaps we need 2 unrelated asserts:

1. cycle_save_name requiring cycle_len or use_clr or use_clr_beta
2. use_clr, use_clr_beta both requiring cycle_len

I'm new to this project, so perhaps it's a standard here to rely on python failing as a way to indicate that something is wrong, rather than designing a more user-friendly API?

sgugger:
Jeremy and I are in the process of rewriting the library completely for the v1, so this is going to be the version with a better API. For now we just maintain this one minimally, but a PR aimed at better user interface will always be appreciated.
I think an assert cycle_len in the three tests (cycle_save_name, use_clr, use_clr_beta) would be the best way.

stas00:
I added asserts to cover #2 in my comment above - there are only 2 such tests, not 3.

#1 assert is completely unrelated, since it needs either of 3 to be deployed.

stas00:
> Jeremy and I are in the process of rewriting the library completely for the v1, so this is going to be the version with a better API. For now we just maintain this one minimally, but a PR aimed at better user interface will always be appreciated.

Awesome! Then is there a point of working on this kind improvements in the current codebase? Is this useful and you backport these PRs to the new codebased? or perhaps if you release the beta soon I can just start using that and help with what I can from the user side, making the API more defensive. I'm also aware that this is taking your time, reviewing, validating and commenting, so perhaps this is counterproductive.

I have just spent many hours hunting some strange failure in CUDA, until I finally removed a few parameters in the call in ColumnarModelData.from_data_frame and then it worked. So I am going to dig in and try to understand why fastai didn't handle it gracefully - since CUDA level errors are not helpful at all.

sgugger:
There is always a point in making small adjustments if it helps users in the meantime (and some may not want to go to v1). In any case it's useful for us to take note of all of these bugs/lack of features to make sure the next version is better.
We are in the early stages of development for now, but as soon as there is something to share and test, Jeremy will ask for help (there is a section dedicated to the development on the forum).

stas00:
Sounds good, @sgugger. I will then continue sending what I find can improve user's experience as I gradually find my way through fastai codebase.

Thank you for your support so far.

PR #650, opened 2018-07-24
Enable skipping validation for n-epoch (training speedup)

quocpeyrot:
This simple change enables training speedup when validation computation can take quite some (or same) time comparing to the training time (e.g. for sampling-based fully convoluted semantic segmentation) and you don't really care about the validation metrics up until a given number of epochs, when you know they will start being high enough for you to care, and therefore track.

### DIFF
--- a/fastai/model.py
+++ b/fastai/model.py
@@ -102,6 +102,7 @@ def fit(model, data, n_epochs, opt, crit, metrics=None, callbacks=None, stepper=
     seq_first = kwargs.pop('seq_first', False)
     all_val = kwargs.pop('all_val', False)
     get_ep_vals = kwargs.pop('get_ep_vals', False)
+    validate_skip = kwargs.pop('validate_skip', 0)
     metrics = metrics or []
     callbacks = callbacks or []
     avg_mom=0.98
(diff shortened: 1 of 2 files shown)


quocpeyrot:
It looks like one check failed, but considering my change is unrelated, I'm guessing the upstream master is failing as well.
(and it's not failing in my current sandbox)

sgugger:
That's a good idea: one thing though, I'd rather have the validate function not be called at all and nothing printed when we skip the validation. Then in all the on_epoch_end callbacks, we should add a test to detect if the metrics in the argument is None or not.

quocpeyrot:
I thought about it, but there are several drawbacks with this approach, and that's why I went with the approach in this pull request:

- in my local sandbox, I can also compute global metrics on some other dataset periodically. Not printing anything at all (i.e. not even "nan") would create issues for stats/column alignments during the progress reports

- as you said, that requires to go through all the on_epoch_end callbacks, and it's likely to break ppl's code/notebooks as well. My gut feeling was that "nan" float wouldn't break legacy/other ppl's code as much as not providing metrics at all.
As another example, the "visualize" option does not break with nan, it just doesn't print any graphical arrow (since boolean on 'nan' are False). If you really skip validation values (instead of putting 'nan'), the visualize feature in print_stats() is going to break unless it is changed as well.

- cosmetics: the conditional if in fit() around the validate() call didn't look as nice as an extra argument

I really think using 'nan' would be more elegant and more bug-proof for this feature.

quocpeyrot:
The changes for the class SaveBestModel look fine and look indeed more elegant than mine.

But I'm not quite sure about the changes for the class LossRecorder. If you don't record the nan values for the validation losses, the losses and val_losses lists won't have the same length, which might later be an issue when you want to analyze/plot both recorded losses:
they won't align anymore (side node: plotting a list with some nan values actually works, which is nice because the plot has the proper x axis values, and you can plot both training and validation losses on the same plot even if you skipped some validation epochs).

sgugger:
I don't get what you mean for the plot. If I put some float('nan') in a list, I get an empty plot at the end, which is why I was removing the values. I get it's not ideal because it won't have the same number of elements as the normal losses, but there is some processing to do at the end in any case.

quocpeyrot:
Sorry maybe I was too vague.

let's say you have 10 epochs and you skip validation for 5 epochs. Without your change to the LossRecorder, at the end self.losses will have 10 values, and self.val_losses as well, with the first 5 being 'nan', e.g.  losses=[9,8,7,6,5,4,3,2,1,0]  and val_losses=[nan, nan, nan, nan, nan, 14, 13, 12, 11, 10]

with that, you can directly plot both of these lists and they will properly align with the epoch x axis of the plot, without any post-processing:
plt.plot(list(range(10)), losses, val_losses)

In other words, plot already handles nan values just fine, which is quite convenient. There's no need to (and I'm arguing you shouldn't) filter them out during recording. No processing needed at all.

with your change you will get self.losses [0,1,2,3,4,5,6,7,8,9] and self.val_losses [5, 6, 7, 8, 9], and indeed you will need to post-process those lists if you want to plot them (or do some other analysis).

sgugger:
Ah! This is a skip for the n first epochs, not validate every n epochs. I got confused.
Yes your approach makes more sense then.

quocpeyrot:
👍 :-)

PR #651, opened 2018-07-24
embedding-related fixes in the codebase and the notebook

stas00:
codebase:
- we aren't yet at torch 0.4, so can't use dtype=torch.int64 - switch to long()
- defensive check that emb_szs have the cardinality of 2+

notebook (dl1/lesson3-rossman.ipynb):
- fix the notebook to use all of data and not just the test set for duration variables. the latter leads to broken embedding sizes.
- add m.summary() which now works


sgugger:
Thanks!

Issue #658, opened 2018-07-27
metrics for mixedinput classification


stas00:
Someone raised the issue [here](http://forums.fast.ai/t/structured-learner-for-kaggle-titanic/8378/37) that we don't have metrics functions for mixedinput classification (and we don't have any notebooks demonstrating this class of DL). 

It appears that somewhat similar to `accuracy/accuracy_np` - we need multidimensional torch preds support (plus these come in as log_preds). I implemented a version that works, but need help with better naming, before submitting PR.

```
def recall_torch(preds, targs, thresh=0.5):
    pred_pos = torch.max(preds > thresh, dim=1)[1]
    tpos = torch.mul((targs.byte() == pred_pos.byte()), targs.byte())
    return tpos.sum()/targs.sum()

def precision_torch(preds, targs, thresh=0.5):
    pred_pos = torch.max(preds > thresh, dim=1)[1]
    tpos = torch.mul((targs.byte() == pred_pos.byte()), targs.byte())
    return tpos.sum()/pred_pos.sum()

def log_fbeta_torch(log_preds, targs, beta, thresh=0.5):
    assert beta > 0, 'beta needs to be greater than 0'
    beta2 = beta ** 2
    preds = torch.exp(log_preds)
    rec = recall_torch(preds, targs, thresh)
    prec = precision_torch(preds, targs, thresh)
    return (1 + beta2) * prec * rec / (beta2 * prec + rec)

def log_f1_torch(log_preds, targs, thresh=0.5): return log_fbeta_torch(log_preds, targs, 1, thresh)
```
usage:

```
m.fit(lr, 2, cycle_len=2, cycle_mult=3, metrics=[accuracy,log_f1_torch])
100% 8/8 [00:00<00:00, 12.92it/s]
epoch      trn_loss   val_loss   accuracy   log_f1_torch 
    0      0.472473   0.438977   0.780899   0.722966  
    1      0.429974   0.426494   0.814607   0.763082  
    2      0.417189   0.444517   0.820225   0.778917  
    3      0.372991   0.430631   0.825843   0.7864    
    4      0.323502   0.440252   0.831461   0.791569  
    5      0.284224   0.49835    0.837079   0.806387  
    6      0.251587   0.454803   0.814607   0.760272  
    7      0.23454    0.459216   0.814607   0.760272  

Out[44]:
[array([0.45922]), 0.8146067335364524, 0.760271861001706]
```

Thank you.

stas00:
[...]

PR #661, opened 2018-07-30
metrics additions

stas00:
as discussed here: https://github.com/fastai/fastai/issues/658
- add torch/log_preds versions of the metrics: f1, fbeta, precision, recall
- rename the previous ones with the same name to have _np suffix

I checked the notebooks and didn't see any notebook that used either of `f1, fbeta, precision, recall`, so it's relatively safe to switch. It'll blow for the users who used it.

### DIFF
--- a/fastai/metrics.py
+++ b/fastai/metrics.py
@@ -1,13 +1,18 @@
 from .imports import *
 from .torch_imports import *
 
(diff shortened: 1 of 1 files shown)


stas00:
Thank you, @sgugger

PR #663, opened 2018-07-30
Precision/Recall: prevent divide by zero error

dust0x:
For zero labels, both `targs.sum()` and `pred_pos.sum()` currently throw a divide by zero error. Instead, I propose adding a constant `epsilon=1e-8` for numerical stability.


dust0x:
Sure, it's done.

PR #665, opened 2018-07-30
Precision/Recall: prevent divide by zero error

dust0x:
Redoing #663, #664 following recent merge #661 .

> For zero labels, both `targs.sum()` and `pred_pos.sum()` currently throw a divide by zero error. Instead, I propose adding a constant `epsilon=1e-8` for numerical stability.


dust0x:
Strange. Why should it first pass and then fail when all I did was add a space?

sgugger:
No you're good, the fail is unrelated to this.

dust0x:
Yep, just saw #453. Thanks.

Issue #644, opened 2018-07-21
Notebook pascal broken 


punkt2:
The Notebook [pascal](https://github.com/fastai/fastai/blob/master/courses/dl2/pascal.ipynb) from "Lesson 9: Deep Learning Part 2 2018 - Multi-object detection" is broken in Section "Single object detection".

The problem occurs when setting metrics
```
learn.metrics = [detn_acc, detn_l1]
```

```
learn.fit(lr, 1, cycle_len=3, use_clr=(32,5))
```
After the first epoch following exception is thrown.

```
~/fastai/courses/dl2/fastai/model.py in validate(stepper, dl, metrics, seq_first)
    236             batch_cnts.append(batch_sz(x, seq_first=seq_first))
    237             loss.append(to_np(l))
--> 238             res.append([f(preds.data, y.data) for f in metrics])
    239     return [np.average(loss, 0, weights=batch_cnts)] + list(np.average(np.stack(res), 0, weights=batch_cnts))
    240 

~/fastai/courses/dl2/fastai/model.py in <listcomp>(.0)
    236             batch_cnts.append(batch_sz(x, seq_first=seq_first))
    237             loss.append(to_np(l))
--> 238             res.append([f(preds.data, y.data) for f in metrics])
    239     return [np.average(loss, 0, weights=batch_cnts)] + list(np.average(np.stack(res), 0, weights=batch_cnts))
    240 

AttributeError: 'list' object has no attribute 'data'
```

In this case y is a list and has no data attribute.

This commit cause the error https://github.com/fastai/fastai/commit/d1d0b60558ae6dab9b9ad47b5928b144e95946c5

punkt2:
just reinstalled the fastai environment and the notebook works now.

punkt2:
too fast, the error still exists.

sgugger:
Yeah, I see where it comes from: basically the y.data needs to be mapped on to all the elements of the list if y is a list (like [x.data for x in y]). I'll look at it this week, but feel free to propose a PR in the meantime!

lharries:
The current way the accuracy functions in the pascal notebook is setup requires that the full y list be passed in, which is then handled within the function
i.e.
```
def detn_l1(input, target):
    bb_t,_ = target
    bb_i = input[:, :4]
    bb_i = F.sigmoid(bb_i)*224
    return F.l1_loss(V(bb_i),V(bb_t)).data

def detn_acc(input, target):
    _,c_t = target
[...]

PR #667, opened 2018-07-30
Bug fix - Handle listy predictions in validate()

lharries:
preds.data means that validate() will error when a list of predictions is passed in (a similar problem to #644)

This fix is needed for when model is run on multiple inputs sequentially and hence has multiple predictions, e.g. when I was creating a Siamese network

This is a similar fix to #660

### DIFF
--- a/fastai/core.py
+++ b/fastai/core.py
@@ -14,6 +14,10 @@ def listify(x, y):
     if len(x)==1: x = x * n
     return x
 
+def datafy(x):
+    if is_listy(x): return [o.data for o in x]
+    else:           return x.data
+
 conv_dict = {np.dtype('int8'): torch.LongTensor, np.dtype('int16'): torch.LongTensor,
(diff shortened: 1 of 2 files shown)


lharries:
@sgugger would you mind looking at this, it's a similar fix to #660 which you previously approved.

sgugger:
Mmm, when we start using the same code so many times, it means we need a helper function. I think the best solution is to write a function datafy that takes an object like preds/y and returns:
- preds.data if preds is not a list
- [o.data for o in preds] if preds is a list
Then all of this code will be simplified into 
res.append([f(datafy(preds), datafy(y)) for f in metrics)])
Can you modify your PR into something like that?

lharries:
I've implemented datafy() as suggested, definitely looks cleaner!

sgugger:
Great, thanks for your help!

PR #669, opened 2018-07-30
arxiv notebook improvements/fixes

stas00:
- add instructions on how to acquire arxiv.csv
- add code to generate all_arxiv.pkl (thanks to http://forums.fast.ai/t/where-to-download-data-for-these-lessons-4-5-arxiv-wikipedia-etc/8434/11)
- fix a variety of bugs in the notebook


sgugger:
First thing I see: feedparser isn't in the dependencies of fastai, so you should add a mention that the user should install it.

PR #678, opened 2018-08-03
Improve the flexibility of Dataset.resize

rcoh:
This PR adds 3 features:
1. Resumable resizing. If a previous resize operation failed or if more images were added to the dataset, `resume=True` (the new default) will find and resize them. The previous behavior would not attempt to perform any resizing if the `targ` directory existed. By checking all files at the beginning of the run, it's comparable in speed to the original (checking if files exist individually is significantly slower).
2. Add support for custom resizing functions via `fn` parameter. (default=None which defaults to im.resize)
3. Add error handling to image resizing -- rather than crashing mid run, any errors are output at the end

~I also added mypy types to the functions that I touched.~ I'm happy to remove if they're bothersome, but they make development easier in IDEs like PyCharm as well as making the code a bit easier to use in the future.
Fixes #678

### DIFF
--- a/fastai/dataloader.py
+++ b/fastai/dataloader.py
@@ -22,7 +22,7 @@ def get_tensor(batch, pin, half=False):
 
 
 class DataLoader(object):
-    def __init__(self, dataset, batch_size=1, shuffle=False, sampler=None, batch_sampler=None, pad_idx=0,
+    def __init__(self, dataset: Dataset, batch_size=1, shuffle=False, sampler=None, batch_sampler=None, pad_idx=0,
                  num_workers=None, pin_memory=False, drop_last=False, pre_pad=True, half=False,
                  transpose=False, transpose_y=False):
         self.dataset,self.batch_size,self.num_workers = dataset,batch_size,num_workers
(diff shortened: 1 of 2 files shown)


sgugger:
Thanks a lot for all the work. As you suggested though, we'd rather not have the types inside the code since it's not done anywhere else. That way the code of the library will remain consistent. Could you just remove them?
We're adding types in fastai_v1 by using annotations, I hope that'll make life easier for pycharm user :)

rcoh:
Removed the type annotations and made it a bit more fault tolerant

sgugger:
Thanks. I've added some clean-up (removed one last type and changed the doc strings so that they match the rest of the library).
Also, when resume=False you were putting the filenames list to [] so it wasn't doing anything (instead of starting back from 0). I'll let you check the modifications I made and tell me if it's okay to merge.

rcoh:
Your change wasn't my original intention but I think it's a reasonable behavior -- My intention was to preserve the old behavior in `resume=False` which was to not do anything if `newpath/targs` existed (essentially, "have we ever tried to resize to this size"). In hindsight, it's not super clear that that's a useful thing -- I think your change probably makes sense; resume=False -> redo everything. resume=True -> start from currently existing images

PR #687, opened 2018-08-07
more functional meanshift example

jattenberg:
The existing meanshift example used mutability quite heavily and left out some wanted features like convergence checking. I made the implementation a bit nicer and feature rich.

I havent run this on a host with cuda yet, but if there is interest in merging the branch, then I will.


jph00:
Yes certainly interested! I don't think _cuda makes sense as a suffix, since those funcs are not cuda-specific afaict. Also, you should probably use `to_gpu` as appropriate to respect the USE_GPU flag.

jph00:
@jattenberg are you still interested in following up on this?

jattenberg:
yes, i'll send a revision shortly.

PR #690, opened 2018-08-09
Update Utils.py : Remove unwanted imports

Gokkulnath:
I believe We dont use keras in any notebooks So These dependencies/imports can be removed. Further limit_mem() function will be used only when keras is used.


sgugger:
Eh, this file has no reason to be here, it comes from the old course. Closing this but removing the file entirely, thanks for pointing this out!

PR #711, opened 2018-08-18
fixed RandomBlur never being called

vlad-sensei:
RandomBlur was never applied because `self.apply_transform` was checked, while `self.store.apply_transform` was set


vlad-sensei:
Thank you for clarification! Should be fixed how, hope I understood the purpose of `store` correctly

jph00:
Thanks!

Issue #710, opened 2018-08-17
3 test failures in test_core.py


ReinoutHeeck:
Linux Mint, NVidia GeForce GTX 1080Ti

[testresuts.txt](https://github.com/fastai/fastai/files/2297906/testresuts.txt)

sgugger:
@hermesdt Could you like at this? Not sure, but it looks like it's in the new tests you implemented.

hermesdt:
Sure, I'll take a look asap (today or tomorrow)

hermesdt:
@ReinoutHeeck did you run the tests having a GPU connected?

ReinoutHeeck:
Yes, the other tests ran successfully. 

I also ran the dl course lesson 1 notebook since reporting this issue -- that worked.

sgugger:
Mmm, maybe we should put the tests requiring the GPU under a test torch.cuda.is_available? Even if it's not the first intent, we'd still like the library to be functional in a CPU-only environment.

ReinoutHeeck:
Just to make sure I rebooted my machine and re-ran all the tests. 

As you can see the output is unchanged: 
[testresuts_all.txt](https://github.com/fastai/fastai/files/2300174/testresuts_all.txt)

ReinoutHeeck:
To be clear: in the jupyter notebook both torch.cuda.is_available() and torch.backends.cudnn.enabled evaluate to True

ReinoutHeeck:
One more bit that may or may not help you:

In jupyter the line "from fastai.imports import *" produces the following output which  after some internet research) I believe to be ignorable:

/home/reinz/anaconda3/envs/fastai/lib/python3.6/importlib/_bootstrap.py:219: RuntimeWarning: numpy.dtype size changed, may indicate binary incompatibility. Expected 96, got 88
  return f(*args, **kwds)
/home/reinz/anaconda3/envs/fastai/lib/python3.6/importlib/_bootstrap.py:219: RuntimeWarning: numpy.dtype size changed, may indicate binary incompatibility. Expected 96, got 88
  return f(*args, **kwds)
/home/reinz/anaconda3/envs/fastai/lib/python3.6/site-packages/sklearn/ensemble/weight_boosting.py:29: DeprecationWarning: numpy.core.umath_tests is an internal NumPy module and should not be imported. It will be removed in a future NumPy release.
  from numpy.core.umath_tests import inner1d

hermesdt:
Can you paste your "pip list" here please?

[...]

PR #713, opened 2018-08-20
Mock functions that sends tensors to gpu

hermesdt:
Tests were failing because of some functions were
sending tensors to the gpu when running tests with a gpu enabled


hermesdt:
Closes https://github.com/fastai/fastai/issues/710

PR #736, opened 2018-08-30
Update layer_optimizer

dust0x:
Assertions in [lines 76 and 81](https://github.com/fastai/fastai/blob/master/fastai/layer_optimizer.py#L74-L82) are redundant and are already handled by `zip_strict_`.


sgugger:
Yeah, but the ones you wrote are sufficient.

PR #760, opened 2018-09-08
fix notebook courses/dl2/cifar10-darknet.ipynb on single GPU

stas00:
- Solved error: RuntimeError: cuda runtime error (10) : invalid device ordinal by this [hint](https://discuss.pytorch.org/t/error-in-multi-gpu-training/3996/2).

- Solved error: one of the variables needed for gradient computation has been modified by an inplace operation: by this [hint](https://discuss.pytorch.org/t/encounter-the-runtimeerror-one-of-the-variables-needed-for-gradient-computation-has-been-modified-by-an-inplace-operation/836/5).

I see some attempts at solving the problems [here](http://forums.fast.ai/t/training-a-model-from-scratch-cifar-10/7897), but they lead nowhere it seems.

### DIFF
--- a/courses/dl2/cifar10-darknet.ipynb
+++ b/courses/dl2/cifar10-darknet.ipynb
@@ -80,7 +80,7 @@
     "        self.conv2=conv_layer(ni//2, ni, ks=3)\n",
     "        \n",
     "    def forward(self, x): \n",
-    "        # When used with DataParallel on a single GPU, we can't use the inplace version of x.add_, which leads to error:\n",
+    "        # changed to x.add, as x.add_ leads to error (happened on single GPU):\n",
     "        # one of the variables needed for gradient computation has been modified by an inplace operation\n",
     "        return x.add(self.conv2(self.conv1(x)))"
    ]
--- a/courses/dl2/cifar10-darknet.ipynb
+++ b/courses/dl2/cifar10-darknet.ipynb
@@ -79,7 +79,10 @@
     "        self.conv1=conv_layer(ni, ni//2, ks=1)\n",
     "        self.conv2=conv_layer(ni//2, ni, ks=3)\n",
     "        \n",
-    "    def forward(self, x): return x.add_(self.conv2(self.conv1(x)))"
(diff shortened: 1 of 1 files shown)


courses/dl2/cifar10-darknet.ipynb
```diff
@@ -79,7 +79,10 @@
     "        self.conv1=conv_layer(ni, ni//2, ks=1)\n",
     "        self.conv2=conv_layer(ni//2, ni, ks=3)\n",
     "        \n",
-    "    def forward(self, x): return x.add_(self.conv2(self.conv1(x)))"
+    "    def forward(self, x): \n",
+    "        # When used with DataParallel on a single GPU, we can't use the inplace version of x.add_, which leads to error:\n",
```
sgugger: Not sure this is linked to one GPU. In general in_place operations and variables require gradients don't always mix well and pytorch raises an error when it isn't fully sure it has all the proper graph for gradients computation. I'd change that comment.

stas00: The comment applies to a specific error: "one of the variables needed for gradient computation has been modified by an inplace operation". Perhaps listing the exact error for a reason why `add_` wasn't used would do the right thing? I can't test whether `add_` does work with DataParallel, since i only have one GPU, but it must have worked for Jeremy. That's why I phrased it this way in first place.

sgugger: Most likely the pytorch version is different as I don't see why the number of GPUs would affect this error. I'd leave the comment to 'using add_ leads to error (...) so we use add instead.'

stas00: Same old 0.3. I thought that was the requirement for fastai v0. Unless Jeremy used a higher version when running this notebook.

I will change the comment.

stas00: Done. And I added a simple statement that it happened when run on single GPU, without saying that this was the culprit. Hope this is reasonable.

PR #766, opened 2018-09-10
Added random gradient from https://arxiv.org/pdf/1808.04293v1.pdf

kdubovikov:
Random gradient technique from https://arxiv.org/pdf/1808.04293v1.pdf looks very easy to implement. I've not done any formal tests now, but plan to do this soon.


jph00:
I'd rather not add more params to the old training loop - the new v1 library has a much better callback system that will allow you to do this easily there. Would you mind adding it there instead, if that's of interest?

PR #769, opened 2018-09-11
Update sampled_sm.py

kamalkraj:
Variable name Typo
`n_hid` => `nhid`
`n_layers` => `n_layers`


sgugger:
RNN_Encoder takes n_hid and n_layers as an argument (line 40 of lm_rnn.py) so I don't think there is a typo.

Issue #774, opened 2018-09-14
NameError: name 'get_rnn_classifier' is not defined


ravilkashyap:
I found a bug in the **fastai.lm_rnn** module.

**get_rnn_classifier** is actually spelled as **get_rnn_classifer**
Notice the **'i'** missing in the word **classifier**
Don't know if it's a typo !

sgugger:
This has been corrected a long time ago, you should update your version ;).

PR #775, opened 2018-09-14
Typo in get_rnn_classifier module

ravilkashyap:
When **'get_rnn_classifier'** was called, it would initialize it as **'get_rnn_classifer'**. Notice the **'i'** missing in **classifier**.

Referring to issue #774


sgugger:
This is to have both get_rnn_classifer and get_rnn_classifier available since we changed the name of this function and didn't want to break existing code.

PR #800, opened 2018-09-29
Add CropNoop - to avoid resize/scale. 

burgalon:
This is currently needed since the current functions do no work with sz=(height, width) but rather a rectangular sz=height=width


sgugger:
The corresponding changes have been pushed to the old library in [this commit](https://github.com/fastai/fastai/commit/23398043f9a33466a9daa880e721a2cb03310fea)

PR #822, opened 2018-10-03
Move courses to ./old folder

brtknr:
It doesn't make sense that the old courses are still in the root directory. I have moved them to the ./old folder and changed the symlinks appropriately.


jph00:
No the courses are still current - we don't have new courses.

PR #828, opened 2018-10-03
Cleanup examples/text.ipynb so that all the cells are run

suzil:
I noticed that https://github.com/fastai/fastai/blob/master/examples/text.ipynb has a `ReadError: empty file` without all the cells run. The changes are from me running all the cells without any errors.


jph00:
Many thanks!

PR #850, opened 2018-10-05
Remove version constraint from testpath dependency

wontheone1:
Because this constraint breaks `conda env update` in mac and ubuntu18.04

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -81,7 +81,7 @@ dependencies:
 - sqlite
 - statsmodels
 - testfixtures
-- testpath>=0.4.2
+- testpath
 - tk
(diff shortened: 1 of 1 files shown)


stas00:
Thank you for the heads up and the patch, @wontheone1.

Indeed conda doesn't yet have 0.4.2 (pip does), so we have to downgrade it instead until the fixed testpath is released on conda. I applied the correct fix https://github.com/fastai/fastai/commit/f221e72ed97d86cfd3bc4f1bfbd9800121116ce7

wontheone1:
@stas00 your commit fixed conda env update again ! thanks

PR #851, opened 2018-10-05
added ability to specify text and label columns by column name

ohmeow:
Added the ability for users to specify the text and label column(s) instead of having to rely on the n_labels argument to delineate between text and label columns.

Why is this useful?

First, it makes running ablations and experiments with various configurations much easier.  Instead of having to create multiple .csv files or DataFrames for each experiment you want to run, you can instead use a *single* file and use it to develop datasets/dataloaders using different combinations of text and/or label columns.  For example, I have a .csv file with roughly a dozen candidate text columns and label columns, some of which I want to use for multi-label problems and others for multi-class problems.  Using the "txt_cols" and "label_cols" arguments, I can load my data into a *single* DataFrame and use that one object to create any number of datasets/dataloaders I need.

Second, this approach is more intuitive and less error-prone.  Instead of having to make sure your .csv file/DataFrame is structured with labels first and then text columns second and then that your n_labels split is correct, users can explicitly and clearly define what text/label columns they want to use.

Notes:
1.  Tested against /examples/text.ipynb
2. Presupposes that DataFrames/CSV files will have a header row

### DIFF
--- a/fastai/text/data.py
+++ b/fastai/text/data.py
@@ -21,9 +21,8 @@ class TextDataset():
                  create_mtd:TextMtd=TextMtd.CSV, classes:Classes=None):
         self.tokenizer = ifnone(tokenizer, Tokenizer())
         self.path,self.max_vocab,self.min_freq = Path(path)/'tmp',max_vocab,min_freq
-        self.txt_cols, self.label_cols, self.pd_header = txt_cols, label_cols, None if (txt_cols is None) else 'infer'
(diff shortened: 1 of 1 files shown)


sgugger:
I've refactored a bit because there was no use keeping this n_labels parameter. Creating a label_cols and text_cols fields cover both use cases. It's testes on examples/text.ipynb but could you check it works on your examples of dataframes? I want to be sure I didn't break anything accidentally before merging.

Issue #837, opened 2018-10-04
Adding Instructions on Google CoLab Setup


WizardOfAus:
Quite a lot of people seem to opt to try using Google CoLab as their GPU notebook as it is free. There are a few confusing instructions on set up using this. Would it be worth adding a section on how to set up the lessons with it?

jph00:
Absolutely - feel free to contrib a PR with that! :)

PR #853, opened 2018-10-05
Create GPU Set up and present the options of Colab

WizardOfAus:
Responding to issue: https://github.com/fastai/fastai/issues/837

Added a general description for people who come across the course and get stuck on set up. I definitely had issues, especially on using free resources i.e. Google Colab.  

Please let me know if I missed any style guidelines. If I should include more or add this file in a different section. Happy to fix, edit or add if I missed anything!

### DIFF
--- a/courses/dl1/GPU Setup.md
+++ b/courses/dl1/GPU Setup.md
@@ -1,0 +1,55 @@
+# How To Set Up Your GPU Notebook
+
+There are various ways to go about setting up your Python Notebook to run on the GPU.
+
+Below I've linked those those mentioned in the lectures and relatively quick to set up. They do however require payment based on useage
+basis.
+
+* [AWS](https://github.com/reshamas/fastai_deeplearn_part1/blob/master/tools/aws_ami_gpu_setup.md)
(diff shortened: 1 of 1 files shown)


jph00:
Thanks that looks on the right track. There is a list of setup resources here: https://github.com/reshamas/fastai_deeplearn_part1 .

It would be best to have the main ones (AWS, GCP, Colab, Paperspace, Crestle) all in one place, rather than only discussing Crestle here. Also, best not using imgur - instead, create a course/images/ folder and put your images there. Thanks!

stas00:
All the platform-specific guides are on the forums at the moment. For now, I think it'd be much simpler to post your guide to a new thread at https://github.com/fastai/fastai/pull/892 and then we can link to it from http://forums.fast.ai/t/fastai-v0-7-install-issues-thread/24652/30 - just ping me when you posted it and I will link it to it.

Would that work, @WizardOfAus? Apologies for the confusing messages, as in the other post @jph00 did suggest you make a PR out of it.

Perhaps, for course v3 we will start making documents that will come together with course files and the any improvements over PR will be the way to go.

WizardOfAus:
@stas00 sounds good! #892 has been closed. Should I still post there as in the comments and then ping you?

PR #855, opened 2018-10-06
Callback priority

kevinlu1211:
Implemented class to handle callback priorities and user-defined callback priority levels. The pull request was a result of [this thread](http://forums.fast.ai/t/allow-for-more-than-one-output-for-loss-and-metric/21991/17).

### DIFF
--- a/fastai/basic_train.py
+++ b/fastai/basic_train.py
@@ -171,16 +171,19 @@ class Learner():
 class LearnerCallback(Callback):
     "Base class for creating callbacks for a `Learner`."
     learn: Learner
+    priority: int=CallbackPriority.NOTSET
     def __post_init__(self):
+        super().__init__(self.priority)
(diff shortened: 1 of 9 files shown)


kevinlu1211:
I've tried a few implementations but this one seems to change the whole file structure the least, and to define a callback with some priority still seems easy enough for the user. E.g. `NewCallback(priority=CallbackPriority.HIGH)`

I haven't included the `priority` argument in any of the `Callback` classes in the `fastai` folder, but rather the priority is set through the `super().__init__()`, though I'm not sure if this is correct.

jph00:
Sorry we overlapped on this one - just pushed the same feature! :) My approach doesn't make it part of `__init__`, but instead class authors override `_order` as needed. Not well tested at all (just check at least LRFinder still works), so do let us know if you see problems, or have ideas for improvements (preferably on the forums).

PR #857, opened 2018-10-06
Updated example notebook to demonstrate how to use DataFrames as data source

ohmeow:


jph00:
Thanks for the idea; we're trying to keep the examples notebooks as simple as possible, so could you instead add this example to the docs (fastai_docs repo)?

PR #858, opened 2018-10-07
Fix TabularModel to have a final output layer

tylernisonoff:
Previously, because `len(actns)` is always one less than `len(sizes[:-1])`, etc..
the zip gets cut off early and we never add a layer with an output size
equal to `out_sz`.  By adding a final `None` actn, we avoid adding an
incorrect ReLU to the end of the model without short-circuiting early.

After fixing this, this sufraced a bug in TabularDataset, in which
the property `c` returns a Tensor rather than an int.

`y.max()` returns a rank zero tensor, so we need to call `.item()`
to get the actual int out.


sgugger:
Oh, you're right! Thanks for pointing this out and finding the bug!

PR #863, opened 2018-10-08
Update image.py

chrisdinant:
The function open_mask would return a tensor with the wrong shape [4, height, width] if the pixel values in the original mask image are 0 and 255 instead of 0 and 1. Converting the PIL image to grayscale ('.convert('L')) before running it through pil2tensor fixes this. 
This is also similar to the 'open_image' function which uses .convert('RGB').

I'm not sure how a change in line 362 appeared in this branch (deletion of **kwargs parameter), I didn't make it.
But I'm relatively inexperienced with pull requests, so I might just be unaware of some git functionality.


sgugger:
Your change happens in line 362 because you didn't have my latest version of image.py (I did it this morning) so you overwrote it. Will fix that.

PR #866, opened 2018-10-08
Added ability to specify label datatype and force clear the cache on init

ohmeow:
1) Added ability to specify label datatype since multilabel problems require a float32 whereas multiclassification requires an int64 (this is the default now).

Note:  Only problem I see is that when I get a DataLoader from the "text_data_from_df" method, the labels are cast to ints even though the Dataset has them as floats.  I couldn't determine where/why this was happening, so I'm hoping the fastai folks can :)

2) Also added ability to force clear the cache on TextDataset initialization (default = false).  This is nice because you no longer have to delete the /tmp directory to rebuild the toks, ids, lbls.


sgugger:
Clear_cache is nice. 
For the type, I'd rather have the dataset infer itself if it's a single class (n_labels=1) or multiclassification problem (n_labels >1), then cast the labels to the right type, than add yet another argument to the function. (This is to be consistent with tabular and vision.)

ohmeow:
Makes sense.

Let me change that real quick and add that to the PR.

Also, any idea on why the batches returned from the DataLoader cast the floats back to ints?

sgugger:
That would be `pad_collate` messing around I guess. I can handle it when you're finished.

ohmeow:
Done.

Yah, you are clear to fix the pad_collate unless you need anything else from me.

PR #870, opened 2018-10-09
_all__ variable should be __all__

spiaz:
It's just a typo.
_all__ is defined instead of __all__
As a result, the statement "from fastai.vision import *" behaves inconsistently.


sgugger:
Good find! Your fix was causing a bug since we didn't update the content of __all__ (that wasn't throwing any errors since it was misspelled), which is why I added a commit.

spiaz:
Wow, I'm glad to hear that it was so helpful :)

PR #874, opened 2018-10-09
OLD v0.7: Remove installation obstacles and make bcolz-tolerant

mprostock:
The reasoning behind this pull request is explained here:
http://forums.fast.ai/t/make-fastai-v0-7-actually-usable-for-beginners-removing-installation-obstacles/25006?u=marcmuc

This is my first real pull request, so any feedback very welcome!!

Details: I have opted for simply printing statements to the user in case bcolz is not found. This is preferable to the more pythonic way of using warning.XXX because a) the ImportWarning gets suppressed by default - so noone would see it and b) warnings pop up really ugly in jupyter notebooks and make the user think something really bad happend. 

If have traced all usages of bcolz, the ones that were not modified are all only indirectly called by the modified methods:

create_empty_bcolz    -> only called from get_activations
get_activations            -> only called from save_fc1
predict_to_bcolz          -> only called from save_fc1
save_fc1                      -> only called from __init__ and set_data (which were modified)

The possibility of users directly accessing the unmodified methods was ignored to not clutter the code too much. (I could add that if desired)


jph00:
Thanks for the thoughtful PR. I don't really like the idea of behaviour changing if someone hasn't installed correctly - it means that they're using an unsupported platform or unsupported install process, which isn't suitable for beginners. If they're not a beginner, then they'll know what a missing module means and how to fix it. Also, it's important that the steps shown in the course result in the behaviour shown in the course (we can't edit the videos).

Missing bcolz is a common problem on Windows. I think a good approach would be to leave in the try/catch for importing it, but in the case of failure, raise an exception with something like "You don't have bcolz installed. That means that you haven't installed the fast.ai course environment fully, or are on an unsupported platform. If you are a python expert, you can fix this by installing the bcolz module. Otherwise please refer to the instructions at {link} for step by step installation steps". {link} could go for instance to the excellent tutorials from Reshemas, or a separate README-course.md file that could be created.

I'm open to increasing the version number dependency for pytorch, but I'd want to see someone confirm that they've run all the notebooks without warnings or errors on Linux with 0.4 first. I suspect it's OK, but I'm not sure I've seen confirmation.

I'll close this PR now - if you're interested in contributing either of the above suggestions, please open a new PR with them separately (it's easier to handle a PR that only contains one feature, rather than combining them).

PR #877, opened 2018-10-10
adds test to show error in lr_find when called twice in sequence

fredguth:
When you call lr_find twice, it throws an error.


sgugger:
As said in #878 this has been solved yesterday.
I'm closing this, but in general, sending us a minimal script that reproduces the bug is great, thanks a lot!

PR #879, opened 2018-10-11
Check if start_lr and end_lr is listy

kevinlu1211:
Currently passing in an array of LR's for `start_lr` and `end_lr` for `lr_find` will fail when the LR's are updated. Changing them into `np.array` avoids this.


sgugger:
User is supposed to pass np arrays, but it's a nice check to make their life easier, thanks!

PR #882, opened 2018-10-11
Issue #641 - Use PackedSequence in RNNcore to improve pooling accuracy (and RNN speed)

ranihorev:
In the current implementation, the RNN output pooling calculates the average of the entire tensor including the padding. This pooling creates inconsistent behavior when predicting batches of texts (due to the change in the amount of padding). 
This PR replaces the previous padding with PyTorch PackedSequence to pack and pad the texts in the RNN core. This change also improves the RNN speed.
We also use the lengths to calculate the average pooling while ignoring the padding; 

Note: In the current implementation of MultiBatchRNN, predicting a batch that contains texts with more than 1*bptt difference in length is not recommended.

p.s.
It's my first PR, so feedback will be appreciated :)

### DIFF
--- a/fastai/text/data.py
+++ b/fastai/text/data.py
@@ -268,11 +268,10 @@ class SortishSampler(Sampler):
 
 def pad_collate(samples:BatchSamples, pad_idx:int=1, pad_first:bool=True) -> Tuple[LongTensor, LongTensor]:
     "Function that collect samples and adds padding."
-    ys = torch.tensor([s[1] for s in samples]).squeeze()
-    padded_xs, lengths = pad_packed_sequence(pack_sequence([torch.tensor(s[0]) for s in samples]), batch_first=True,
-                                             padding_value=pad_idx)
+    ys = torch.tensor(np.array([s[1] for s in samples])).squeeze()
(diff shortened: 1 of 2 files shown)


sgugger:
You made changes to RNNCore, so I expect it's going to break the language model. Did you test you could run the text example notebook entirely? In any case I'll need to check it doesn't change the final result of ULMFiT on IMDB.

For the future, please note that we prefer to get significant changes in the form of a new notebook like [the development notebooks](https://github.com/fastai/fastai_docs/tree/master/dev_nb) used to create the library, so we can easily refactor/check it works properly.

jph00:
Thanks so much for this really interesting contribution! :) As @sgugger mentioned, it is breaking existing code, so we're not going to accept it in this state - and we'd like to be able to play around with your code in a notebook environment so we can fully understand it.

Therefore, would you be able to create a notebook which contains your implementation, along with examples showing how it works with language model and classification training and fine-tuning? Please create a gist and post it at forums.fast.ai so we can discuss.

Many thanks!

ranihorev:
Sorry for the late response. 
It does break the language model (I updated the classifier but forgot about the LM...). Would you prefer a different input format that supports both versions? e.g. set the text lengths input as optional?
Anyway, I'll post my notebook sometime this week.

PR #886, opened 2018-10-12
fixed loss_batch to work when multiple values returned; fixed text package to work correctly with no labels and multilabel

ohmeow:
Please feel free to discard the changes to the jupyter notebooks in this PR ... not sure how they got rolled into it (sorry).

Made a few fixes:
1.  LM now works if no labels passed in
2. Fixed basic_train.loss_batch() so will work when multiple values returned in "out" variable.
3. Fixed text classifier to use the number of labels for multilabeled dataset and class for multiclass

PR #892, opened 2018-10-12
loss_batch works when model returns tuple; updated text package to work with multilabel datasets

ohmeow:
1. Updated basic_train.loss_batch to work when models return a tuple instead of a tensor (e.g., in NLP whereby the framework models return a tuple of (predictions, hidden, hidden))

2. Update text package so that labels are optional

3.  Update text package so that for multilabel datasets, the final # of activations = the # of labels instead of the # of classes


sgugger:
Done in #886

PR #909, opened 2018-10-17
Fix bug for backwards=True when creating a LanguageModelLoader

mboyanov:
Currently it fails with the following error:
`ValueError: some of the strides of a given numpy array are negative. This is currently not supported, but will be added in future releases.`
A simple fix is to copy the data so that it is no longer a view.


jph00:
Please let us know when you've signed the CLA. Thanks for your contribution!

mboyanov:
Done.
Apologize for the delay - I was waiting on my employers to review the CLA and come up with a policy. 
Keep up the good work!

sgugger:
Thanks!

PR #915, opened 2018-10-18
fbeta error because of `y_true` dimension

cstorm125:
For a multi-label classification problem,

```
> preds,y = learn.get_preds()
> fbeta(preds,y)
```

gets the following error:

```
<ipython-input-31-97bbcf467e8e> in <module>()
----> 1 fbeta(preds,y)

~/fastai/fastai/metrics.py in fbeta(y_pred, y_true, thresh, beta, eps, sigmoid)
     10     y_pred = (y_pred>thresh).float()
     11     y_true = y_true.float()
---> 12     TP = (y_pred*y_true).sum(dim=1)
     13     prec = TP/(y_pred.sum(dim=1)+eps)
     14     rec = TP/(y_true.sum(dim=1)+eps)

RuntimeError: The size of tensor a (676) must match the size of tensor b (3573) at non-singleton dimension 1
```

I found that it's because `y_pred` has dimension (batch_size, output_size) and `y_true` has dimension (batch_size,). Reshaping `y_true` to be (batch_size,1) seems to solve the problem.

### DIFF
--- a/fastai/metrics.py
+++ b/fastai/metrics.py
@@ -8,7 +8,7 @@ def fbeta(y_pred:Tensor, y_true:Tensor, thresh:float=0.5, beta:float=2, eps:floa
     beta2 = beta**2
     if sigmoid: y_pred = y_pred.sigmoid()
     y_pred = (y_pred>thresh).float()
-    y_true = y_true.float().reshape(-1,1)
+    y_true = y_true.float()[:,None]
     TP = (y_pred*y_true).sum(dim=1)
     prec = TP/(y_pred.sum(dim=1)+eps)
     rec = TP/(y_true.sum(dim=1)+eps)
--- a/fastai/metrics.py
+++ b/fastai/metrics.py
@@ -8,7 +8,7 @@ def fbeta(y_pred:Tensor, y_true:Tensor, thresh:float=0.5, beta:float=2, eps:floa
     beta2 = beta**2
     if sigmoid: y_pred = y_pred.sigmoid()
     y_pred = (y_pred>thresh).float()
-    y_true = y_true.float()
+    y_true = y_true.float().reshape(-1,1)
     TP = (y_pred*y_true).sum(dim=1)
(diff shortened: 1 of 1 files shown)


sgugger:
Thanks for flagging this. Just changed to the way we usually add dims in the library.

CaveSven:
Hi @sgugger!
With the changes from 2c9bc0c8c902ee158c03ccf04a5eb9ace914a09f this breaks:
```python
y_pred = Tensor([[0, 1], [0, 1], [0, 1]])
y_true = Tensor([[0, 1], [0, 1], [0, 1]])
fbeta(y_pred, y_true)
```

with error

```
---------------------------------------------------------------------
RuntimeError                        Traceback (most recent call last)
<ipython-input-101-03d98cc6777e> in <module>
      1 y_pred = Tensor([[0, 1], [0, 1], [0, 1]])
      2 y_true = Tensor([[0, 1], [0, 1], [0, 1]])
----> 3 fbeta(y_pred, y_true)

<ipython-input-60-76cf21da1a1a> in fbeta(y_pred, y_true, thresh, beta, eps, sigmoid)
      6     y_true = y_true.float()[:,None]
      7     TP = (y_pred*y_true).sum(dim=1)
----> 8     prec = TP/(y_pred.sum(dim=1)+eps)
      9     rec = TP/(y_true.sum(dim=1)+eps)
     10     res = (prec*rec)/(prec*beta2+rec+eps)*(1+beta2)

RuntimeError: The size of tensor a (2) must match the size of tensor b (3) at non-singleton dimension 1
```

As a consequence, multilabel training is currently not working. If there is additional complexity here because of batching, I suggest adding a couple of test cases.

sgugger:
fbeta is going to disappear for the class Fbeta that is going to give the true value. The previous approach was to average the fbetas over batches, but this will compute the real score. Fbeta has been tested over training, just pass it like a metric: `metrics=[accuracy_thresh, Fbeta(beta=2)]` for instance.

PR #919, opened 2018-10-18
fix model imports

kevinlu1211:
fix the imports for the models as they weren't being exposed when imported


sgugger:
Will change vision init since the convention will be to do from fastai.vision import models.
Then models.wrn22 or models.Darknet...

PR #922, opened 2018-10-19
Fix references to newly renamed _SAMPLE paths

sd2k:
The sample attribute names were modified in commit 4524202 but some
classmethods still referred to the old names.

This was affecting the 'quick CLI test' in the README (`jupyter nbconvert --execute --ExecutePreprocessor.timeout=600 --to notebook examples/tabular.ipynb`), which now runs as expected.

This commit also adds some unit tests to check that the methods work. The `get_imdb` method currently fails unless the 'en' spacy language has been downloaded, and I'm not sure whether that's an expected prerequisite for the unit tests, so I've left it commented out for now.

### DIFF
--- a/fastai/datasets.py
+++ b/fastai/datasets.py
@@ -27,7 +27,7 @@ class URLs():
 
     @classmethod
     def get_adult(cls):
-        path = untar_data(cls.ADULT)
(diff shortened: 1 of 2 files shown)


jph00:
Extra thanks for the tests! :)

PR #924, opened 2018-10-20
Untar data extracts files to .fastai directory

bearpelican:
Datasets now downloads data to config.yml data_path.


bearpelican:
@jph00 currently having a mental block on how to tests default paths and not screw up all the other tests. But putting this PR up for now if this is something more urgent

bearpelican:
Will fix tests

PR #938, opened 2018-10-23
Allow specifying classes in the TextClasDataBunch.from_folder method

mboyanov:
When creating a `TextClasDataBunch.from_folder` allow the user to specify the classes.
This is useful for example with the imdb dataset, so you can ignore the 'unsup' folder.


sgugger:
This is done more generally in #944 so I'm closing this one.

PR #940, opened 2018-10-23
Fix typo in `get_tabular_learner` and `TabularModel`

allenlawrence94:
Change "ps" argument in `get_tabular_learner` and `TabularModel.__init__` to "drops" so that it agrees with the docs and the other models that use dropout.


sgugger:
Actually it's where there is `drops` that it's inconsistent. `ps` is used in the ConvLearner and RNN.classifier so we want to stick with those.
The docs need to be corrected if they talk about drops.

PR #944, opened 2018-10-23
fix(): change in text/data.py TextDataBunch Class method [from_ids_files, from_tokens,…

StatisticDean:
… from_df, from_csv, from_folder] so that classes argument is passed to the call to TextDataset

### DIFF
--- a/fastai/text/data.py
+++ b/fastai/text/data.py
@@ -14,13 +14,13 @@ def read_classes(fname):
 
 class BaseTextDataset():
     "To directly create a text datasets from `ids` and `labels`."
-    def __init__(self, ids:Collection[Collection[int]], labels:Collection[Union[int,float]], vocab_size:int, 
+    def __init__(self, ids:Collection[Collection[int]], labels:Collection[Union[int,float]], vocab_size:int,
                  classes:Classes=None):
         self.ids,self.labels,self.vocab_size,self.classes = ids,labels,vocab_size,classes
-    
+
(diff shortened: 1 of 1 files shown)


fastai/text/data.py
```diff
@@ -351,7 +351,7 @@ def from_folder(cls, path:PathOrStr, tokenizer:Tokenizer=None, train:str='train'
         "Create a `TextDataBunch` from text files in folders."
         tokenizer = ifnone(tokenizer, Tokenizer())
         path=Path(path)
-        k_names = ['max_vocab', 'chunksize', 'min_freq', 'n_labels', 'clear_cache']
+        k_names = ['max_vocab', 'chunksize', 'min_freq', 'n_labels', 'clear_cache', 'classes']
         txt_kwargs, kwargs = extract_kwargs(k_names, kwargs)
         train_ds = TextDataset.from_folder(path, tokenizer, train, shuffle=shuffle, vocab=vocab, **txt_kwargs)
         datasets = [train_ds, TextDataset.from_folder(path, tokenizer, valid, classes=train_ds.classes,
```
mboyanov: The TextDataset.from_folder will fail if 'classes' is specified, because a value for classes will be passed twice: once from train_ds.classes and once from the **txt_kwargs

sgugger: Ah good catch! Do you want to adapt your previous PR with txt_kwargs.pop()?

mboyanov: Sure, will do :+1:

PR #951, opened 2018-10-23
Added np.darray type to "is_listy()" check

ohmeow:
Verified this work against the imdb dataset (multiclass) and a custom multilabel dataset.  

Now RNNLearner.classifer(...) set the final number of activations correctly for multilabel problems.


navjotts:
@ohmeow: can you list down the exact steps which are causing the issue for you?

I m not sure if its the best idea to bring in a `np.ndarray` for the listy check - was discussing about this function today only with @jph00

May be the approach should just be to check if the input is iterable or not
(though strings are also iterable, but can be handled)

jph00:
Have a look see where the function is used now. Is there anywhere that using an ndarray or more general Iterable could break the code? Does it make sense semantically to allow iterables in general?

navjotts:
Yea I take the iterable suggestion back (even dicts are iterable) - and we are specifically looking for `listy` as the function name suggests

ohmeow:
@jph00 This is my fear as well.  This utility function is probably used throughout the codebase and even though the notebooks I'm testing against work, it may break something else.

To be honest, I'd rather the API be updated to require the developer to be explicit about the number of labels they are using rather than attempting to infer it with methods like "is_listy()".  I find the latter approach to be fragile, problematic, and generally difficult to refactor going forward.  I kinda feel like the code is trying to do too much here.

In this case, if instead of inferring the final linear layer's activations that was made an argument to `RNNLearnder.classifier` in the form of something like `n_out`, the problem would be no more.

@navjotts ... here are the steps:
```
data_clas = TextClasDataBunch.from_df(Path(CLS_PATH), train_df, valid_df, tokenizer=tokenizer, vocab=vocab,  txt_cols=corpus_cols, label_cols=LABELS_SENT[1:], bs=bsz,  clear_cache=True)

print(data_clas.train_ds.labels.shape, data_clas.valid_ds.labels.shape)
# (13692, 8) (1522, 8)

learn = RNNLearner.classifier(data_clas, drop_mult=0.9, bptt=bptt, emb_sz=em_sz, nh=nh, nl=nl,
                             lin_ftrs=[50], ps=[0.1],
                             alpha=2., beta=1.)
learn.clip = 25.
learn.metrics = [opt_th, fscore, multilbl_accuracy]
learn.model_dir = 'models'
learn.callback_fns.append(RocAucEvaluation)
learn.load_encoder(f'{lm_pre}lm_enc')
learn.fit_one_cycle(1, 1e-2)
```

navjotts:
Adding ndarray to `is_listy` definitely breaks this function
```
def tensor(x:Any)->Tensor:
    "Like `torch.as_tensor`, but handle lists too"
    return torch.tensor(x) if is_listy(x) else as_tensor(x)
```
(which is the cause of the failing test)

yea, I don't think `is_listy` should be changed - its behaving as expected (and is used heavily across the code base in its current definition)

@ohmeow : best would be to detail out the steps of the issue/bug

ohmeow:
@navjotts : Lmk if you need something more than what I posted above.

navjotts:
@ohmeow : got it, thanks (my earlier post went at the same time as yours)

StatisticDean:
Maybe the "good" solution to this would be to change the code of RNNLearner.classifier to be able to accept array of label (Which is the kind of label you get from a TextClasDataBunch.from_id_files).

ohmeow:
Unless I'm missing something ...

```
n_class = ( len(ds.classes) if (len(lbl) == 1) else len(lbl) ) 
```

For multiclass, there is a SINGLE label (column in our dataframe) ... whereas for multilabel, there will be  MULTIPLE labels (columns in your dataframe).

jph00:
I'm going to close this PR since it's not desired behaviour. Note also that a PR that changes such a low level function also needs to contain tests of everything that the change might impact (which currently would be a *lot* of work!) So probably better to wait for better test coverage until changing something with such wide impact.

If there is something which has a bug or problematic behaviour, we'd welcome a PR that includes a failing test and a minimal fix for it (including tests for any impacted code).

PR #953, opened 2018-10-23
update to_init() in core.py with type definitions, add and update tests

navjotts:


navjotts:
Closing, recreating with the correct credentials

PR #956, opened 2018-10-24
Update data.py

yang-zhang:
When `ImageMultiDataset.__init__` is passed a `labels` and a non-null `classes`, there is no guarantee that all items in `labels` is already in `classes`, which is typical for a case like `valid_ds = ImageMultiDataset(valid_fns, valid_labels, train_ds.classes)`. This update allows this case to work without raising `KeyError`.

### DIFF
--- a/fastai/vision/data.py
+++ b/fastai/vision/data.py
@@ -114,7 +114,7 @@ class ImageMultiDataset(LabelDataset):
         self.classes = ifnone(classes, uniqueify(np.concatenate(labels)))
         self.class2idx = {v:k for k,v in enumerate(self.classes)}
         self.x = np.array(fns)
-        self.y = [np.array([self.class2idx[o] for o in l], dtype=np.int64)
+        self.y = [np.array([self.class2idx[o] for o in l if o in self.classes], dtype=np.int64)
                   for l in labels]
         self.loss_func = F.binary_cross_entropy_with_logits
 


sgugger:
This works but I'm wondering in what situation you'd have a label in your validation set that wasn't in your training set.

yang-zhang:
This situation can occur for multilabel datasets with a large number of unbalanced labels and/or with a heavy tail. A real example is [deepfashion](http://mmlab.ie.cuhk.edu.hk/projects/DeepFashion/AttributePrediction.html) dataset's attributes data. 

When passing a multilabel df to `ImageDataBunch.from_df`, `ImageMultiDataset.from_folder` will `random_split` the examples, and some infrequent labels may end up only in the validation set. (of course, some may be only in the training set.)

Typically we use a smaller sample of the data to develop the model before running it on all the data. The sampling can make this situation more likely to occur.

jph00:
This is certainly a problem worth solving! But I don't think this is the right solution - it gives an incorrect measure of accuracy. Let's move this discussion to the forum - can you open a thread at https://forums.fast.ai/c/fastai-users/fastai-dev . Generally it's best to discuss changes to functionality there first BTW, so we can all agree on an approach. Also, please include a test in your PRs that fail without your code, and pass with it, as well as a test of a case that already worked without your code (and still works with it).

yang-zhang:
Forum thread created [here](https://forums.fast.ai/t/how-should-imagemultidataset-handle-labels-in-valid-ds-unseen-by-train-ds/27869).

PR #961, opened 2018-10-24
Add n_labels attribute to TextDataSet Class and uses it when calling an RNNLearner.classifier

StatisticDean:
When you create a RNNLearner.classifier from a TextClasDataBunch where you didn't specify the classes, if the labels are in an array, and One Hot encoded, the RNNLearner.classifier gets the wrong number of label. 
This pull request adds n_labels attribute to TextDataSet so that if you call TextClasDataBunch(...,n_labels=n_labels,...) , and you call a RNNLearner.classifier on this TextClasDataBunch, you get the right number of labels in your model.


jph00:
Unfortunately there's not enough here for us to be able to accept this PR. I added this template today that might help explain what we need:

https://github.com/fastai/fastai/blob/master/.github/pull_request_template.md

If you can provide a notebook demonstrating the issue you're facing, using real data, that would be really helpful for ensuring a good discussion on the forum, so we can all agree on a good solution.

PR #968, opened 2018-10-24
Fix tabular example

albertvillanova:
Fix tabular example notebook line from:
`df = URLs.get_adult()`
to:
`df = pd.read_csv(path/'adult.csv')`


jph00:
Well spotted :)

Issue #946, opened 2018-10-23
RNNLearner.classifier uses classes instead of labels for final linear for multilabel problems


ohmeow:
I remember this being fixed awhile back in a PR I made, but it appears its ugly head has resurfaced.

Looks like the final activation are based on classes (0 and 1) instead of the number of labels (which is 8).  The problem appears to be on this line:
`> n_class = (len(ds.classes) if (not is_listy(lbl) or (len(lbl) == 1)) else len(lbl))`

`lbl = array([0., 0., 0., 1., 1., 0., 0., 0.], dtype=float32) `

Error:
"ValueError: Target size (torch.Size([40, 8])) must be the same as input size (torch.Size([40, 2]))"

ohmeow:
I already submitted the PR after posting the issue ... the one that is under review at the moment.  I'm glad to follow-up on this given some direction about how to submit such unit tests.

PR #974, opened 2018-10-25
added test for RNNLearner handling of multilabel datasets

ohmeow:
Per Jeremy's request (#946) ... adding a failing test that demonstrates the RNNLearner does not handle setting the final # of activations correctly for multilabel datasets.

(btw, if there is a better way or a "right" way to structure these tests please let me know ... I'm new to unit testing in python).

### DIFF
--- a/tests/test_text_learner.py
+++ b/tests/test_text_learner.py
@@ -1,0 +1,19 @@
+from fastai import *
+from fastai.text import *
+
+import pytest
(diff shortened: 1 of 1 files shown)


ohmeow:
@navjotts If there is anything else you need besides the test I submitted, lmk.  I'm actively working on some multilabel datasets at work so glad to help.

PR #976, opened 2018-10-25
TextDataset refactor

navjotts:
Consolidate TextMtd.DF and TextMtd.CSV into 1 (as a df can always be formed from a csv).

Would like to add more tests, but opening a PR for @sgugger to review


fastai/text/data.py
```diff
@@ -327,6 +318,7 @@ def from_df(cls, path:PathOrStr, train_df:DataFrameOrChunks, valid_df:DataFrameO
         path=Path(path)
         k_names = ['max_vocab', 'min_freq', 'n_labels', 'txt_cols', 'label_cols', 'clear_cache', 'classes']
         txt_kwargs, kwargs = extract_kwargs(k_names, kwargs)
+        print(txt_kwargs, kwargs)
```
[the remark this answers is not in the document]
navjotts: oops, that was not supposed to be a part of the commit :/
removing

sgugger:
It looks good (test failed because of bad luck on something else, nothing you did). Just tell me when you're done adding tests!

navjotts:
👍 

Yea, something s up with 1 of the deployment machines I think (noticed the same happened on all past few commits) - all tests seem to be passing, but the build is failing on the same machine each time (`Linux_Conda Python37`)

navjotts:
@sgugger : done 
Have added support for `multilabel` tests 
(moving towards the area around which there is an issue and a PR being discussed)
https://github.com/fastai/fastai/pull/974
https://forums.fast.ai/t/fastai-rnnlearner-classifier-wrong-size-of-head-of-classifier/27960

tests/test_text_data.py
```diff
@@ -1,9 +1,24 @@
+import pytest
 from fastai import *
 from fastai.text import *
 
+@pytest.fixture(scope="module")
+def text_df(multilabel=False):
```
navjotts: could make this more generic (multilabel being the number of labels) - but have kept it clean and simpler, as focus is on testing a mutlilabel datatest and not really making it super generic

can update if you feel otherwise

tests/test_text_data.py
```diff
@@ -12,3 +27,52 @@ def test_should_load_backwards_lm():
...
+    filepath = os.path.join(path, filename+'.csv')
+    try:
+        text_csv_file(filepath)
+        data = TextClasDataBunch.from_csv(path, train=filename, valid=filename)
+        assert len(data.classes) == 2
+        assert set(data.classes) == set([True, False])
+    finally:
+        shutil.rmtree(path)
+
+def test_from_csv_multilabel():
```
navjotts: similar, could combine this with `test_from_csv()` into a more generic form (as they do share common code) - but again have kept it separate for clarity instead of an if-else inside 1 longer function

sgugger:
How come the multi-label test passes?

tests/test_text_data.py
sgugger: It's getting a bit long, so if it's not too much to do, I'd prefer one generic test.

navjotts:
because the bug/issue is in the learner part @sgugger (not the DataBunch/Dataset creations)
(i m working on fixing that on a separate branch)

tests/test_text_data.py
navjotts: Ok 👍 
(I assume you meant this for combining singlelabel and multilabel tests into one - the other comment)

updating

PR #977, opened 2018-10-26
test_ds property added to DataBunch

joshfp:
test_ds property on DataBunch class will allow access test dataset in a straightforward way (instead of test_dl.dl.dataset).
BTW, this is my first pull request, so please let me know if any amendment is required.


jph00:
Congrats on your first PR!

PR #979, opened 2018-10-26
test core series2cat

dienhoa:
I create a test for series2cat() in core.py

### DIFF
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -137,3 +137,10 @@ def test_even_mults():
     a = even_mults(start=1, stop=8, n=4)
     b = array([1.,2.,4.,8.])
     np.testing.assert_array_equal(a,b)
+
+def test_series2cat():
+    df = pd.DataFrame({'col1': [1, 2], 'col2': [3, 4], 'col3':[5, 6]})
+    cols = 'col1','col2'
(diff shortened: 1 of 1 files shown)


sgugger:
Thanks! Can you add a last line to check 'col3' is still numeric?

dienhoa:
Thanks @sgugger , I add the last line checking for the others columns types. This is the first time I make a test so if there is something not ok, please tell me. Do you think in this case I need to add more cases (with string for example) or it is enough ?

tests/test_core.py
```diff
@@ -137,3 +137,11 @@ def test_even_mults():
     a = even_mults(start=1, stop=8, n=4)
     b = array([1.,2.,4.,8.])
     np.testing.assert_array_equal(a,b)
+
+def test_series2cat():
+    df = pd.DataFrame({'col1': [1, 2], 'col2': [3, 4], 'col3':[5, 6]})
+    cols = 'col1','col2'
+    series2cat(df,*cols)
+    for col in cols:
```
jph00: Not a big deal, but we prefer to avoid extra vertical space by using 1 line for loops where they're small.
https://docs-dev.fast.ai/style.html

PR #980, opened 2018-10-26
Add maybe copy tests

radekosmulski:
To get my feet wet with pytest, picked something relatively straight forward from `fastai.core`. Fixed a couple of typos along the way and made minor change to `maybe_copy`.

Will start a thread to discuss.

### DIFF
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -1,6 +1,7 @@
 import pytest, torch
 import numpy as np
 from fastai import *
+from tempfile import TemporaryDirectory
 
 def test_cpus(): assert num_cpus() >= 1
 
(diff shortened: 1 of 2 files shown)


tests/test_core.py
```diff
@@ -144,4 +144,36 @@ def test_series2cat():
     series2cat(df,*cols)
     for col in cols:
         assert (df[col].dtypes == 'category')
-    assert (df['col3'].dtypes == 'int64')
\ No newline at end of file
+    assert (df['col3'].dtypes == 'int64')
+
+def _write_file(name):
```
jph00: These can be put on one line.

tests/test_core.py
```diff
@@ -144,4 +144,36 @@ def test_series2cat():
...
+        ([Path('./first')], [Path('./second')]),
+    ])
+    def test_copies_if_does_not_exist(self, old_fnames, new_fnames):
+        for fn in old_fnames: _write_file(fn)
+        maybe_copy(old_fnames, new_fnames)
+        for fn in new_fnames: assert os.path.exists(fn)
+        _clean_up(old_fnames+new_fnames)
+
+    def test_copies_if_older(self):
+        _write_file('first')
```
jph00: Probably best to use the `tempfile` std lib and a `with` block to ensure that the files are created somewhere appropriate, and automatically cleaned up when done.

tests/test_core.py
radekosmulski: this looks nicer :)
```
def _write_file(dn, name): f = open(f'{dn}/{name}', 'w'); f.write(str(name)); f.close()
```

jph00:
Great job!

PR #990, opened 2018-10-27
create x in cuda for model_sizes()

dienhoa:
In my case, as default, tensor x in model_sizes() is created in cpu rather than cuda. It leads to error at `x = m.eval()(x)`.

Link in forum: [error x in model_sizes](https://forums.fast.ai/t/developer-chat/22363/335)


sgugger:
That's not the right answer either since it's now going to bug every time the model is not on the GPU ;). You have to put x to the same device as the model.

dienhoa:
Thanks @sgugger , you're right. I have fixed it in the new commit. pls check if it is ok

sgugger:
You can be shorter with the magic command new:
```
x = next(m.parameters()).new(1,ch_in,*size)
``` 
will create a tensor on the same device as the m.parameters()

dienhoa:
Ah, I didn't know that new() command. Much better now. Thank you

PR #991, opened 2018-10-28
Added test for half precision

marii-moe:
First attempt at adding a test to fastai. This specific test is a unit test to cover model2half function that is part of fastai. 

class ConvBNNet - this might need to go somewhere else?

### DIFF
--- a/docs/troubleshoot.md
+++ b/docs/troubleshoot.md
@@ -124,6 +124,28 @@ Replace `1.0.7` with the version you're trying to install. (Without version it'l
 
 
 
+### Can't install the latest fastai conda package
+
+Say there is `fastai-1.0.12` out there, but when you run:
+
+```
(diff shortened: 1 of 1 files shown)


sgugger:
Great idea! Just two little things:
1. can you create this in a new file test_callback_fp16 ? Even if the method you're testing is in torch_core, it would be better there.
2. can you use the function `conv_bn_relu` to create your model? It's in the layers module.

jph00:
Better still would be to use the `simple_cnn` function.

Also note that we can't rely on cuda in tests, since they run on a CI server without a GPU. Does this test run OK if we use cpu() instead? I suspect it might not, since CPU doesn't support half precision.

If that's the case, you'll probably need to check if cuda is available and skip otherwise:

https://docs.pytest.org/en/latest/skipping.html#id1

PR #993, opened 2018-10-28
Add save_image helper function for save Image class to a file

noklam:
Adding a helper function save_image similar to fastai 0.7, which take a fastai Image class as input and save as a file.


noklam:
@stas00 signed, thanks.

sgugger:
Thanks for your contribution! I think it would be best if it was a method from the Image class: `Image.save(fname)`. Can you amend your PR to do this?

noklam:
I agree it's should belong to the Image Class instead, I have updated my PR. I have no experience amending PR before, please let me know if I did something wrong. Thx!

sgugger:
Thanks!

PR #994, opened 2018-10-28
Add unit tests for tabular transformations

floscha:
I noticed that there are no unit tests that specifically cover the `tabular/transform.py` module. Thus, I have implemented a new `test_tabular_transform.py` test suite to cover all relevant functionality of this module, including both the `Categorify` as well as the `FillMissing` classes.

### DIFF
--- a/tests/test_tabular_transform.py
+++ b/tests/test_tabular_transform.py
@@ -8,15 +8,24 @@ def test_categorify():
     df = pd.read_csv(path / 'adult.csv')
     cat_names = ['workclass', 'education', 'marital-status', 'occupation',
                  'relationship', 'race', 'sex', 'native-country']
-    df = df[cat_names]
+    cont_names = [column_name for column_name in df.columns
+                     if column_name not in cat_names]
     train_df, valid_df = df[:1024].copy(),df[1024:1260].copy()
-    categorify_transform = Categorify(cat_names, [])
-
+    original_cont_train_df = train_df[cont_names].copy()
+    original_cont_valid_df = valid_df[cont_names].copy()
+    
+    categorify_transform = Categorify(cat_names, cont_names)
     categorify_transform.apply_train(train_df)
     categorify_transform.apply_test(valid_df)
 
-    assert all([isinstance(dt, CategoricalDtype) for dt in train_df.dtypes])
-    assert all([isinstance(dt, CategoricalDtype) for dt in valid_df.dtypes])
+    # Make sure all categorical columns have the right type
+    assert all([isinstance(dt, CategoricalDtype) for dt
+                in train_df[cat_names].dtypes])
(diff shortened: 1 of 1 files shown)


sgugger:
Thanks a lot for your contribution. Could you adjust it to fit the usual fastai style? That means removing the empty lines. Then I have a few comments/suggestions of thins to add if you can that I'll put directly in the code.

tests/test_tabular_transform.py
```diff
@@ -0,0 +1,58 @@
...
+                 'relationship', 'race', 'sex', 'native-country']
+    df = df[cat_names]
+    train_df, valid_df = df[:1024].copy(),df[1024:1260].copy()
+    categorify_transform = Categorify(cat_names, [])
+
+    categorify_transform.apply_train(train_df)
+    categorify_transform.apply_test(valid_df)
+
+    assert all([isinstance(dt, CategoricalDtype) for dt in train_df.dtypes])
+    assert all([isinstance(dt, CategoricalDtype) for dt in valid_df.dtypes])
```
sgugger: Could we keep the whole df and check the columns not in cat_names aren't modified?

tests/test_tabular_transform.py
```diff
@@ -0,0 +1,58 @@
...
+    cont_names = ['age', 'fnlwgt', 'education-num', 'capital-gain',
+                  'capital-loss', 'hours-per-week']
+    df = df[cont_names]
+    train_df, valid_df = df[:1024].copy(),df[1024:1260].copy()
+    fill_missing_transform = FillMissing([], cont_names)
+
+    fill_missing_transform.apply_train(train_df)
+    fill_missing_transform.apply_test(valid_df)
+
+    train_df.isna().values.sum() == 0
```
sgugger: I think there is an assert missing there.

tests/test_tabular_transform.py
```diff
@@ -0,0 +1,58 @@
...
+    expected_valid_median = valid_df.median()
+    fill_missing_transform = FillMissing([], cont_names, add_col=False)
+
+    fill_missing_transform.apply_train(train_df)
+    fill_missing_transform.apply_test(valid_df)
+    train_median = train_df.median()
+    valid_median = valid_df.median()
+
+    assert np.array_equal(train_median.values, expected_train_median.values)
+    assert np.array_equal(valid_median.values, expected_valid_median.values)
```
sgugger: The median here could have changed since we're filling with the median of the training set, not the validation set. Maybe check that missing values are replaced by that median instead?

floscha:
Hi @sgugger, thanks for your feedback. I agree with all your points and consequentially adapted the code. Let me know if further changes are needed. Cheers

sgugger:
Thanks!

PR #1000, opened 2018-10-28
Add unit tests for vision data

anishpdalal:
I renamed test_vision.py to test_vision_data.py to make the file naming pattern identical for text. I also added unit tests for denormalization, train_validation_split, and for factory method from_name_re.


navjotts:
@anishpdalal : I think these tests (apart from `test_from_name_re`) do belong in the `test_vision.py` file -- we didn't have a `test_vision_data.py` as we didn't have any particular *data_loading* tests for vision
(infact I was working on creating some currently on a `test_vision_data.py` file)

I would suggest moving `test_from_name_re` to a new `test_vision_data.py` file - and have rest in the original `test_vision.py` file

navjotts:
Adding for reference: https://github.com/fastai/fastai/pull/1003

anishpdalal:
@navjotts I implemented the changes you suggested.

tests/test_vision_data.py
```diff
@@ -0,0 +1,19 @@
+import pytest
+from fastai import *
+from fastai.vision import *
+
+@pytest.fixture(scope="module")
+def path(request):
+    path = untar_data(URLs.MNIST_TINY)
+    d = defaults.device
+    defaults.device = torch.device('cpu')
```
sgugger: Why impose a device? The tests are run on cpu only and GPU machines so you don't need to make sure we're on cpu (and it defeats the purpose of testing widely).

tests/test_vision_data.py
```diff
@@ -0,0 +1,19 @@
...
+    def _final(): defaults.device = d
+    request.addfinalizer(_final)
+    return path
+
+def test_from_name_re(path):
+    fnames = get_files(path/'train', recurse=True)
+    pat = r'\/([^/]+)\/\d+.png$'
+    data = ImageDataBunch.from_name_re(path, fnames, pat, ds_tfms=(rand_pad(2, 28), []))
+    assert data.c == 2
+    assert all(["3" in data.classes, "7" in data.classes]) is True
```
sgugger: Could you add something to test the elements in the training and validation set are correctly labeled?

tests/test_vision.py
```diff
@@ -40,3 +40,16 @@ def test_normalize(path):
...
+    original_x, y = data.valid_dl.one_batch()
+    data.normalize()
+    normalized_x, y = data.valid_dl.one_batch()
+    denormalized = denormalize(normalized_x, original_x.mean(), original_x.std())
+    assert round(original_x.mean().item(), 4) == round(denormalized.mean().item(), 4)
+    assert round(original_x.std().item(), 4) == round(denormalized.std().item(), 4)
+
+def test_training_validation_split(path):
+    data = ImageDataBunch.from_folder(path, ds_tfms=(rand_pad(2, 28), []), valid_pct=0.20)
+    assert len(data.train_ds) > len(data.valid_ds)
```
sgugger: This is too loose, data.train_ds should be 4 times bigger, so let's test for that (maybe a 3.95 or something smaller with the rounding errors).

sgugger:
Thanks for your contribution. I've added a few comments and suggestions of things to add if you can.

tests/test_vision.py
anishpdalal: That was my initial thought as well. However, when running this multiple times I found the train/validation ratio to vary widely around 4 (probably because of a smaller sample size) and couldn't account for the difference with rounding errors. Thoughts?

tests/test_vision_data.py
```diff
@@ -0,0 +1,17 @@
+import pytest
+from fastai import *
+from fastai.vision import *
+
+@pytest.fixture(scope="module")
+def path(request):
```
navjotts: @anishpdalal : we won't need the `request` variable anymore (not being used)

Also note that there might be a merge conflict with this PR - https://github.com/fastai/fastai/pull/1003
(as we both added the same file and the same function)

Please merge upstream and resolve the conflict if that gets merged 1st

tests/test_vision.py
sgugger: That's weird. What's the lower ration you find if you run this, say... 1000 times?

anishpdalal: Min ~ 3.05
Mean ~ 4.02
Max ~ 5.39

anishpdalal:
I've opened up a new PR #1017 that includes tests for denormalize and from_name_re.

PR #1003, opened 2018-10-29
Add from_folder test for vision

navjotts:
**Discussion:** https://forums.fast.ai/t/bug-when-using-imagedatabunch-from-folder-and-valid-pct-with-test/28292
**Related fix:** https://github.com/fastai/fastai/pull/997

### DIFF
--- a/tests/test_vision_data.py
+++ b/tests/test_vision_data.py
@@ -1,0 +1,24 @@
+import pytest
+from fastai import *
+from fastai.vision import *
+
+def image_file(path):
+    file = open(path, 'w')
+    file.close()
+    return file
+
+def test_from_folder():
+    n_classes = 2
+    for valid_pct in [None, 0.5]:
+        path = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'data', 'tmp')
+        folder_paths = [os.path.join(path, folder) for folder in ['train', 'valid', 'test']]
+        class_paths = [os.path.join(folder_paths[0], str(i)) for i in range(n_classes)]
+        for each in [path]+folder_paths+class_paths: os.makedirs(each)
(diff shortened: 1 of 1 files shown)


sgugger:
I've added a test folder in mnist with 20 new images. Could you amend your PR to use that real data?

navjotts:
> I've added a test folder in mnist_tiny with 20 new images. Could you amend your PR to use that real data?

👍 updating

tests/test_vision.py
```diff
@@ -3,12 +3,8 @@
 from fastai.vision import *
 
 @pytest.fixture(scope="module")
-def path(request):
+def path():
     path = untar_data(URLs.MNIST_TINY)
-    d = defaults.device
```
navjotts: referring: https://github.com/fastai/fastai/pull/1000#discussion_r228928591
@sgugger : I had removed the device dependency existing in `test_vision.py` tests (it was present from before)

but right now on checking more, it seems like @jph00 had pushed that change few weeks back
commit: https://github.com/fastai/fastai/commit/d343a36caadc97214f7276f69d2413888a448b5e

Marking him, may be there was some specific thought behind it - we can revert the current commit if that was the case

navjotts: Okay just discussed with Jeremy in person - he was checking some specific CPU compatibility issue at that time - we don't need it anymore
Cc @sgugger 
(keeping the commit, PR is ready for review)

navjotts: Ignore my above comment - we do that need those **particular tests** that way - Jeremy just clarified 

I m going to rename this file to *_cpu* to avoid confusion

navjotts:
Updated, ready for review.

PR #1006, opened 2018-10-29
Add tests for get total length

radekosmulski:
just a small test for fastai.core.get_total_length


radekosmulski:
oops sorry let me fix this

radekosmulski:
accidentally pushed a commit that should not be here, it is fixed now

PR #1015, opened 2018-10-30
Made TextDataBunch call DataBunch.create properly.

matiu2:
DataBunch.create takes three initial args 'train_ds, valid_ds, test_ds',
we were passing it a list of 'datasets', which it was putting all into
train_ds.

Now we break those out before calling DataSet.

Now some of the text examples and documentation are starting to work for
me.

If you are adding new functionality, please first create a thread on [fastai-dev](https://forums.fast.ai/c/fastai-users/fastai-dev) describing the functionality. Generally it's best to discuss changes to functionality there first, so we can all agree on an approach. Please include a test in your PR that fails without your code, and passes with it, as well as a test of a case that already worked without your code (and still works with it). Currently fastai has poor test coverage, so don't take the current tests as a role model - we're all working to fix it together! When creating your PR, please remove all the text in this template, and add details about your PR.

### DIFF
--- a/fastai/text/data.py
+++ b/fastai/text/data.py
@@ -352,9 +352,7 @@ class TextDataBunch(DataBunch):
     @classmethod
     def create(cls, datasets:Collection[TextDataset], path:PathOrStr, **kwargs) -> DataBunch:
         "Call's `DataBunch.create` but changes the arguments so it'll work OK"
-        train_ml, valid_ml = datasets[:2]
-        test_ml = None if len(datasets) < 3 else datasets[2]
-        return DataBunch.create(train_ml, valid_ml, test_ml, path=path)
+        return DataBunch.create(*datasets, path=path, **kwargs)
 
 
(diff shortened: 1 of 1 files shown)


navjotts:
I don't think we needed a separate function for this (or I am missing something)

just changing this `cls.create(datasets, path, **kwargs)` to `cls.create(*datasets, path, **kwargs)` should be it, right?
(and looks like the former exists at many places text/data.py)

> Now some of the text examples and documentation are starting to work for
me.
@matiu2 : I ll put some tests for this - can you post an example what didn't work for you, so that we cover whatever you saw

sgugger:
using `cls.create(*datasets, path, **kwargs)` breaks the subclassing in `TextLMDataBunch` or `TextClasDataBunch`.
Not too sure why we would want to create a `TextDataBunch` and not one of the above, but we never know.

navjotts:
gotcha

adding tests regardless

PR #1016, opened 2018-10-30
Add run length decode encode, and open_mask_rle

noklam:
Add open_mask_rle for opening mask when input is a run-length encoding instead of a image file. It's common to see run-length encoding in Kaggle Competition as well. For example, the Carvana train_mask.csv

Two supplementary function `rle_encode` and `rle_decode` are supplied to support this function.

A notebook showing how does it work: 
https://github.com/noklam/log/blob/master/nbs/Open_mask_rle.ipynb

Example of rle_mask
https://www.kaggle.com/c/carvana-image-masking-challenge/data


sgugger:
Hey, thanks a lot for your PR. Can you share a notebook with an example of use of this new functionality? Also some unit tests would be most appreciated. Lastly, don't forget to type-annotate the return of your functions.

jph00:
Can you please update your notebook to show how to actually use these to create a DataBunch? I'm guessing you'll need to create or modify a dataset class to do so.

noklam:
I am not familiar with the PR process, would love to know more how to keep my fork sync while not adding a lot of merge commit especially fastai library update very frequently.

A notebook showing how does it work:
https://github.com/noklam/log/blob/master/nbs/Open_mask_rle.ipynb

I try my best to build the example, advice are welcome. For the data only exist on my local computer, as there is no existing data in this format in the current fastai/data. I try to follow the lesson 3 notebook, no clue how can I build a cleaner method with something like ImageDataBunch.from_xxx for this custom dataset.

Sorry I notice that I should start a forum thread before creating this PR. I open a thread now, as I feel it still need to be polished a bit.
https://forums.fast.ai/t/segmentation-dataset-for-run-length-encoding/28967

PR #1021, opened 2018-10-30
Added tests requirs grad and initialization

marii-moe:
Added tests for requires_grad. Also added test for apply_init because it tested a path through requires grad.


tests/test_torch_core.py
```diff
@@ -19,3 +21,31 @@ def test_tensor_with_tensor():
...
+
+def test_requires_grad():
+    m = simple_cnn(b)
+    assert requires_grad(m) == True
+
+def test_requires_grad_set():
+    m = simple_cnn(b)
+    requires_grad(m,False)
+    disj = False
+    for p in m.parameters():
```
sgugger: Those little for loops fit best in one single line. Can you edit your PR accordingly?

jph00:
Actually you should be able to do it without a for loop using something like `torch.all()`.

Please either name your tests with a full name of exactly what's being tested, or add a 2nd string argument to have your assertions explain exactly what they're testing.

marii-moe:
> Actually you should be able to do it without a for loop using something like `torch.all()`.
I spent a bit of time checking for this myself, and was not able to find anything of this nature. I will implement your other suggestions though.

marii-moe:
Latest commit should have fixed above mentioned issues.

PR #1024, opened 2018-10-31
added test that shows issue with batch size 1

benediktbrandt:
added a test that fails due to an accidental batch of size of 1 at the end of training (pytorch checks batch sizes in BatchNorm2D).

For more details see: https://forums.fast.ai/t/fix-batch-of-size-1-raises-exception-during-batchnorm-in-training/28242/4


benediktbrandt:
This is the output that I see
[output.txt](https://github.com/fastai/fastai/files/2532320/output.txt)

benediktbrandt:
Also 
```
(fastaiv3dev) b1@b1-desktop ~/fastaiv3/fastai/tests $ 
python
Python 3.6.6 |Anaconda, Inc.| (default, Oct  9 2018, 12:34:16) 
[GCC 7.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import fastai
>>> fastai.__version__
'1.0.19.dev0'
```

jph00:
We can't have tests that download all of resnet pretrained weights every time - so please use pretrained=False.

Also, a test needs to run in no more than a second or so on CPU, except for one integration fixture per application.

So this test probably needs to be updated to use a dummy dataset that's created of the right size to trigger the error, and a network that does the same thing.

Let me know if you have any questions.

jph00:
Oh also, when contributing a failing test, please mark them with the pytest skip flag so they don't trigger the CI.

benediktbrandt:
@jph00 @sgugger I updated the PR, incorporating your feedback. The test runs now in about a second on my CPU. The pr contains one failing and and one succeeding test, illustrating the issue. 

@sgugger It doesn't matter what resnet is used, in fact the only important thing is the occurrence of a BatchNorm2d layer.

sgugger:
Thanks a lot, I'll work on this later today

PR #1030, opened 2018-10-31
To match the type of smoothener.add_value

Pendar2:
Loss would always be a single float tensor

### DIFF
--- a/fastai/callback.py
+++ b/fastai/callback.py
@@ -218,7 +218,7 @@ class CallbackHandler():
 
     def on_backward_begin(self, loss:Tensor)->None:
         "Handle gradient calculation on `loss`."
-        self.smoothener.add_value(loss.detach())
+        self.smoothener.add_value(loss.detach().item())
         self.state_dict['last_loss'], self.state_dict['smooth_loss'] = loss, self.smoothener.smooth
         for cb in self.callbacks:
             a = cb.on_backward_begin(**self.state_dict)


jph00:
Please explain what bug you're fixing, and include the tests requested in the PR template.

Pendar2:
It's not broken as it is right now. Currently the callback.py file is almost completely decoupled from pytorch except in 2 places:
https://github.com/fastai/fastai/blob/master/fastai/callback.py#L221
https://github.com/fastai/fastai/blob/master/fastai/callback.py#L270

This change would allow me to easily use everything in callback.py in the https://github.com/fastai/tf-fit project by adding 2 functions to tensorflow tensors
tf.Tensor.detach = lambda x: x
tf.Tensor.item = lambda x: x.numpy()

jph00:
That's really interesting. Just checking on this - will get back to you asap.

Pendar2:
How about removing .item() on https://github.com/fastai/fastai/blob/master/fastai/callback.py#L270 would that be possible?

sgugger:
That's possible without breaking anything, it that helps you.

sgugger:
I added it in [this commit](https://github.com/fastai/fastai/commit/d5eef11e7f38066ed45439519ca38d9ffeb63179) accidentally. Let me know if I can close this PR.

Pendar2:
That will work, thanks.

PR #1039, opened 2018-10-31
Add RemoveMaxCorrelation TabularTransform

floscha:
This PR adds a new TabularTransform class to automatically remove features that correlate with the target variable more than a given threshold (0.95 by default). The PR is heavily inspired by TransmogrifAI's [SanityChecker](https://github.com/salesforce/TransmogrifAI/blob/master/core/src/main/scala/com/salesforce/op/stages/impl/preparators/SanityChecker.scala) and includes the required unit tests.

PR #1040, opened 2018-10-31
Add RemoveMinVariance TabularTransform

floscha:
This PR adds a new TabularTransform class to automatically remove features with a variance below a given threshold (0.00001 by default) as we expect these features to not add any value to the model's predictive power. Like #1039, this PR is also heavily inspired by TransmogrifAI's [SanityChecker](https://github.com/salesforce/TransmogrifAI/blob/master/core/src/main/scala/com/salesforce/op/stages/impl/preparators/SanityChecker.scala) and includes the required unit tests.

### DIFF
--- a/fastai/tabular/transform.py
+++ b/fastai/tabular/transform.py
@@ -1,7 +1,8 @@
 "Cleaning and feature engineering functions for structured data"
 from ..torch_core import *
 
-__all__ = ['Categorify', 'FillMissing', 'FillStrategy', 'TabularTransform']
+__all__ = ['Categorify', 'FillMissing', 'FillStrategy', 'TabularTransform',
+           'RemoveMinVariance']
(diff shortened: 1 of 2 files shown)


jph00:
Thanks for the contribution! I'd been thinking that we'd use this algorithm for that purpose:

http://topepo.github.io/caret/pre-processing.html#zero--and-near-zero-variance-predictors

I've found it's pretty reliable. Have you compared these approaches?

(Note that it's generally a good idea to raise an idea in the forum first, so we can agree on an approach, before you jump in to coding. That way, you know you're working on something that's likely to fit with our plans!)

sgugger:
Please amend this PR to work with the latest big refactor and adapt to the comment above. I'm closing this for now, thanks!

PR #1041, opened 2018-10-31
Replace csv data by simpler fabricated test data

floscha:
This PR replaces the rather complex data, read from a test CSV file, by individual much smaller test DataFrames. This both makes the tests run slightly faster and, even more importantly, makes it much easier to comprehend the functionality assessed by the unit tests and why they are expected to produce a certain result.


sgugger:
Thanks!

PR #1044, opened 2018-10-31
move model output to CPU

kevinlu1211:
fixes gpu memory leakage when no loss function is defined in `loss_batch`


sgugger:
As discussed on the forum, I'm not sure this is needed.

PR #1047, opened 2018-11-01
Leave URLs already ending in .tar.gz or .tgz alone

drscotthawley:
If you are adding new functionality, please first create a thread on [fastai-dev](https://forums.fast.ai/c/fastai-users/fastai-dev) describing the functionality. Generally it's best to discuss changes to functionality there first, so we can all agree on an approach. Please include a test in your PR that fails without your code, and passes with it, as well as a test of a case that already worked without your code (and still works with it). Currently fastai has poor test coverage, so don't take the current tests as a role model - we're all working to fix it together! When creating your PR, please remove all the text in this template, and add details about your PR.

### DIFF
--- a/tests/test_vision_train.py
+++ b/tests/test_vision_train.py
@@ -16,6 +16,9 @@ def learn():
 def test_accuracy(learn):
     assert accuracy(*learn.get_preds()) > 0.9
 
+def test_error_rate(learn):
+    assert error_rate(*learn.get_preds()) < 0.1
+
 def test_1cycle_lrs(learn):
     lrs = learn.recorder.lrs
     assert lrs[0]<0.001


drscotthawley:
I have now signed the CLA.

fastai/datasets.py
```diff
@@ -64,7 +64,11 @@ def url2name(url): return url.split('/')[-1]
 def _url2path(url, data=True):
     name = url2name(url)
     return datapath4file(name) if data else modelpath4file(name)
-def _url2tgz(url): return datapath4file(f'{url2name(url)}.tgz')
+def _url2tgz(url):
```
jph00: Generally we prefer to avoid using more vertical space than necessary. E.g. here this can simply something like:
```
ext = "" if url.endswith((".tar.gz", ".tgz")) else ".tgz"') 
return datapath4file(f'{url2name(url)}{ext}')
```

jph00: BTW the style guide is here if you're interested: https://docs-dev.fast.ai/style.html

drscotthawley: Thanks for your patient instruction.  I'm happy to comply to the style guide and will gladly do as you ask -- and/or feel free to simply add that code yourself and delete my PR.  

I would like to learn how to write the test as well. I don't see anything comparable for _url2tgz() currently in [fastai/tests/test_datasets.py](https://github.com/fastai/fastai/blob/master/tests/test_datasets.py), and having now read the ["Writing tests" section of the testing guide](https://docs-dev.fast.ai/test.html#writing-tests)...I don't see how it applies.  Is there a different guide I should be reading for this?   Thanks. 

...and now that I modified the code and pushed it, I see your [git guide](https://docs-dev.fast.ai/git.html#step-2-create-a-branch), and have already pushed to the master branch which I was not supposed to do.  I will follow the instructions to switch to a new branch...

PR #1050, opened 2018-11-01
Add sorting in uniqueify

radekosmulski:
When creating an `ImageDataBunch`, ordering of classes in the `classes` list will differ depending on the files in the source of the data (even if the classes are the same, but say we removed some files to improve data quality of we get a hold of a bigger data set).

I will start a thread on this on the forums. Also, very happy to write the tests for this but might not have a chance to do so before Monday.

### DIFF
--- a/fastai/core.py
+++ b/fastai/core.py
@@ -59,7 +59,7 @@ def ifnone(a:Any,b:Any)->Any:
 
 def uniqueify(x:Series)->List:
     "Return unique values of `x`"
-    return list(OrderedDict.fromkeys(x).keys())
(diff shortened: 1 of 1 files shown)


radekosmulski:
I thought about this PR a bit more (and also got very helpful feedback from @sgugger :slightly_smiling_face:) - I don't like my proposed way of fixing this. I provide further commentary in the [forum thread](https://forums.fast.ai/t/pr-1050-add-sorting-in-uniqueify/28908/5).

sgugger:
I'm closing this for now then, if that's ok.

PR #1056, opened 2018-11-02
test_ds not loaded

test9723497834565:
In a nutsell, the problem is that the test dataset is empty when using ImageDataBunch.from_folder(...).
The cause is that the method cannot find the test images like it does for the training images.

I've created a notebook to demonstrate the problem and its solution:
[debug is_test.zip](https://github.com/fastai/fastai/files/2540605/debug.is_test.zip)

Here are some screenshots of the problem from the notebook:

image
image
image


sgugger:
This is not a bug: the test folder should have the images directly inside it, not inside other folders, since it's supposed to be unlabeled data.

test9723497834565:
Thank you for your comment but I don't understand it fully.
 
You said that the test data should be unlabeled. That doesn't make much sense to me as you need to know the label of the test data to be able to produce an accuracy result.

On top of that, the dataset I used for my test is a fastai one (URLs.MNIST_TINY) and the test data for MNIST arranged with subfolders. 

I think that you should consider discussing the issue of how the data is supposed to be laid out on disk with Jeremy.

sgugger:
I think you are confusing validation and test set. In fastai, the test set is intended for unlabeled data, like in a Kaggle competition. This is also the case with the URLs.MNIST_TINY dataset (at least the latest version of it) where there is a valid dataset with labelled data in folders and unlabelled data in a test folder.

Thanks for your concerns but I work closely with Jeremy to know how he wants the library to be and I ask for his advice when in doubt. This specific topic has been discussed twice already on the [forum](https://forums.fast.ai/) and Jeremy is the one who wrote that specific bit of code, so I'm sure it's how he intended it.

PR #1057, opened 2018-11-02
Speed up image resizing and verifying

gazay:
I've noticed this new functionality when thought how to implement it myself and tried it today on openimages dataset from Google (1.7mln images). I noticed that progress bar works strange and it was looks like empty string printed for once in a while. I've debugged code and found that assertion error will be empty if there is no message provided. And assertion error was raised because some of the images were opened as one-channel images by mistake (they are all RGB images).

@sgugger it's really awesome that you've added this functionality! Thank you very much for your excellent work! Though I think this PR can speed up the whole process by moving few operations out of cycle (directory creating, file checking). What I've done here:
1. Moved directory creation out of cycle, so we'll not be calling os to check if a directory exists
2. Make `dest` path for `verify_image` required if there is max_size set and make it `Path` type. This way we'll skip creation of `dest` folder path variable and win a bit of time. I'm not sure about this one, but I don't see when we'll need to call this function without a cycle and for different paths, so it looks like an easy win.
3. Added `resume` param which will skip already existed images in `dest` path and we'll win a lot of time if something unexpected happened in the middle of resizing. Also, I moved path variable creation and checking more up – so we'll skip computations if we don't need them
4. I've added `RGB` conversion for images if 3 channels requested (it was done similar in prev version of library https://github.com/fastai/fastai/blob/master/old/fastai/dataset.py#L42)

### DIFF
--- a/fastai/vision/data.py
+++ b/fastai/vision/data.py
@@ -390,8 +390,8 @@ def download_images(urls:Collection[str], dest:PathOrStr, max_pics:int=1000, max
         for i,url in enumerate(progress_bar(urls)):
             download_image(url, dest/f"{i:08d}.jpg")
 
-def verify_image(file:Path, delete:bool, max_size:Union[int,Tuple[int,int]]=None, dest:PathOrStr='.', n_channels:int=3,
-                 interp=PIL.Image.BILINEAR, ext:str=None, img_format:str=None, **kwargs):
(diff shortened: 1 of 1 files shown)


sgugger:
Thanks! I have just one little comment: if path is set to the default '.', resume should default to False as the idea is to overwrite current images.

gazay:
@sgugger agreed, made changes for it in latest commit

PR #1069, opened 2018-11-03
Fixing test typo

devforfu:
There is a small typo in one of the tests suits where `False` is written in lower case. Not sure if that was already fixed but wanted to let you know.


sgugger:
Thanks!

PR #1074, opened 2018-11-03
Add a basic implementation of a Variational Autoencoder to examples/

tylernisonoff:
modeled after pytorch's vae example (https://github.com/pytorch/examples/tree/master/vae)


jph00:
Cool notebook! However we just want one example for each of fastai's main applications.

PR #1078, opened 2018-11-04
Fix #1077

### LINKED ISSUE #1077: AssertionError: Torch not compiled with CUDA enabled
`learn = Learner.create_unet(data, models.resnet34, metrics=acc_camvid)` 
throws an assertion error on CPU.

AssertionError: Torch not compiled with CUDA enabled

ram-ai:
`learn = Learner.create_unet(data, models.resnet18, metrics=acc_camvid)` 
throws an assertion error Torch not compiled with CUDA enabled.

After fixing up above test pass.


sgugger:
Thanks!

PR #1085, opened 2018-11-05
new version of pil2tensor with test, images, and docs

kasparlund:
Hi silvain her is a clean PR with pil2tensor - pardon


sgugger:
Thanks! Last step is to clean up the docs a bit. You have deleted cells that are weird compared to the current doc. Also, you should use different cells and markdown to comment instead of putting everything in one big cell, it would be more readable on the web version.

kasparlund:
This is my first PR - could you please make those last corrections. I tried to do it but have again run into this infernal synching with master and updating the PR. I am back to  The flow of syncing with the master and updating the PR has gone wrong again.  Would be really nice with a recipe  for dummies (like myself) on how to update a PR when master has moved one. I will leave the PR here for somebody more skilled with git

PR #1087, opened 2018-11-05
solved fp16  bug for multilabel clasification in mixup callback.

tcapelle:
Just added .float() here and there in mixup callback.


sgugger:
This is a good idea but your PR
- deletes one example
- doesn't convert `y1` to float as well

tcapelle:
my bad= (

sgugger:
You still have tabular.ipynb deleted.
Maybe close this PR and paste your code in a new one?

PR #1091, opened 2018-11-06
Image cleaner

lesscomfortable:
Modified download_url so that it now uses `iter_content` insead of `.read` and `.raw` since the latter cannot handle chunked transfer encoding urls.


sgugger:
This breaks the tests.

PR #1093, opened 2018-11-06
Updated the import statement to resolve NameError: name 'DatasetType' is not defined

hiromis:
When running lesson2-download.ipynb notebook, there was an issue in file_picker.py:

screen shot 2018-11-06 at 6 01 33 pm

Updated the import statements to include basic_data.py to resolve the issue.


sgugger:
Thanks!

PR #1095, opened 2018-11-06
set drop_last=True for batch norm training safe

crcrpar:
# Objective
Make training of models with batch normalization safer.

# Details
Current initialization of `DataLoader` causes errors when `len(data_set) % batch_size = 1` and BatchNorm is used.


sgugger:
This is a known issue but I don't think you're addressing it the right way. For one, we don't need to do this on the validation and the test set because the problem only comes during training.
Then we only need to remove the last batch if it's of size 1, not all the time. I'm also working on a fix for this so closing this.

PR #1096, opened 2018-11-06
Add kwargs for Segmentation Dataset to accept shape argument

noklam:
The current SegmentationDataset has a np.array type for y, while open_mask_rle accept str only, so I add one step to convert np.array str type to a pure str. I am not sure is this the best way to do, but it works. 

I also have to add a kwargs because there is no way to pass an argument `shape` if I want to use open_mask_rle as the mask_opener.

A expected way to call the function will be:
databunch = (ImageFileList.from_folder(TRAIN)
        .label_from_df(image_df)
 .random_split_by_pct(0.2)
 .datasets(SegmentationRLEDataset, classes=codes, shape=(768,768))
   .transform(get_transforms(), size=size, tfm_y=True)
        .databunch(bs=bs)
        .normalize(imagenet_stats)
)

https://forums.fast.ai/t/segmentation-dataset-for-run-length-encoding/28967/3

### DIFF
--- a/fastai/vision/data.py
+++ b/fastai/vision/data.py
@@ -147,14 +147,15 @@ class ImageMultiDataset(ImageClassificationBase):
 
 class SegmentationDataset(ImageClassificationBase):
     "A dataset for segmentation task."
-    def __init__(self, x:FilePathList, y:FilePathList, classes:Collection[Any]):
+    def __init__(self, x:FilePathList, y:FilePathList, classes:Collection[Any], **kwargs:Any):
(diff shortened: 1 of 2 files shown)


sgugger:
That's not the right way to do it: you should put your `mask_opener` function to `partial(open_mask_rle, args)` for now. I'm looking at a way to pass the size of x to `_get_y` which would support the case where all images aren't the same size.
Closing this, feel free to open a new one for just the str conversion.

PR #1101, opened 2018-11-07
Is camvid_tiny.tgz here by mistake?

brtknr:
I don't think this should be here...


sgugger:
You're absolutely right!

PR #1106, opened 2018-11-08
Download images suffixes

lesscomfortable:
Conserves image suffixes when downloading with `download_images`


sgugger:
Can you add the test that show it works with the three images I added yesterday on files.fast.ai? Thanks!

lesscomfortable:
Yes, done!

sgugger:
Thanks!

stas00:
@lesscomfortable, the new test for download_images rendered the test suite dependent on the Internet connection. Would it be less of a test if it were to use file:// url instead of http://files.fast.ai, by fetching the images from the local fs, rather than online?

stas00:
and the solution based on the advertised file extension, is good, but could be improved, as many file online carry invalid extensions. A better solution would be to use https://github.com/ahupp/python-magic (a wrapper around libmagic, and file(1).

lesscomfortable:
Good suggestions, thanks Stas! So considering that python-magic is not on conda do you think its ok if we leave it as is? Or is there some better way that does not conflict with the current installation process?

stas00:
Looking at other possible solutions I have stumbled upon this: https://docs.python.org/3/library/imghdr.html which seems to be the perfect solution and it's a core python library.

PR #1115, opened 2018-11-09
save gpu memory on relu

erikgaas:
Add inplace=True to the relu at the base of U-Net.

### DIFF
--- a/fastai/vision/models/unet.py
+++ b/fastai/vision/models/unet.py
@@ -49,7 +49,7 @@ class DynamicUnet(nn.Sequential):
         ni = sfs_szs[-1][1]
         middle_conv = nn.Sequential(conv2d_relu(ni, ni*2, bn=True), conv2d_relu(ni*2, ni, bn=True))
         x = middle_conv(x)
-        layers = [encoder, nn.ReLU(), middle_conv]
+        layers = [encoder, nn.ReLU(inplace=True), middle_conv]
 
         for idx in sfs_idxs:
             up_in_c, x_in_c = int(x.shape[1]), int(sfs_szs[idx][1])


marii-moe:
For referencee: https://pytorch.org/docs/master/notes/autograd.html#in-place-operations-on-variables

Pytorch generally recommends against using in place operations.

erikgaas:
This is really interesting. Has this changed in v1? Has inplace ever been
recommended?

erikgaas:
On that note, pytorch is being a bit contradictory. Their resnet code all
uses in place relus. Some for other architectures. Is this now depricated?

marii-moe:
It works for resnet because you are guaranteed to have a batchnorm before the inplace relu. 
`
class DynamicUnet(nn.Sequential):
    "Create a U-Net from a given architecture."
    def __init__(self, encoder:nn.Module, n_classes:int):
    # --stuff here---
    layers = [encoder, nn.ReLU(inplace=True), middle_conv]
`
DynamicUnet allows you to pass in the encoder, which does not necessarily have a batchnorm. In-place ReLU is architecture specific and therefore cannot go here. If you try running this using a network without batch norm you will get this error: 

RuntimeError: one of the variables needed for gradient computation has been modified by an inplace operation

erikgaas:
Interesting. Thanks. I appreciate the response. I think I get it but I have
a couple other questions. This one I hope actually misses the point. I'm
looking at alexnet. It does not have batchnorm, but it does have inplace
relus. Why? Does the incompatibility have to do with hooks for dynamic net?

Also I am somewhat familiar with in place abn.
https://github.com/mapillary/inplace_abn

Based on what you are saying, is batchnorm with in place relu parameter is
equivalent to this?

marii-moe:
It really just depends on if the in-place operation is overwriting a variable that is used to compute a gradient for back propagation. 

So...
```
input = torch.rand(3,4,4)-0.5
F.relu(input) #does not overwrite input
F.relu(input, inplace=True) #overwrites input
```

If output was used in the computation of a gradient then you would get the "RuntimeError: one of the variables needed for gradient computation has been modified by an inplace operation" error from the above `F.relu(input, inplace=True)` operation. If a COPY of the variable used for the gradient is passed to `F.relu` you would not get an error. 

Does that help to explain the issue?

PR #1121, opened 2018-11-09
fix arguments for isinstance

erikgaas:
Segmentation notebook for camvid fails otherwise.


sgugger:
Saw one, missed the other. Thanks for flagging!

PR #1125, opened 2018-11-10
We need to import is1d in order to use it later on line 121

matiu2:
In commit 2b6baf12003fbc020e07c3088f370b7d50e3a94d we started to use is1d.
This commit will make it possible to use it by importing it


sgugger:
`from ..torch_core import *` imports everything in core, so it's not needed.

matiu2:
Thanks @sgugger . I thought that it just imported pytorch core stuff; I didn't think it imported fastai.core

PR #1128, opened 2018-11-11
Utility for language model inference and serialization

pzelasko:
I've come up with an API for performing LM inference to compute likelihood/perplexity for a given sentence/batch of sentences. It makes my life a bit easier because it serializes all the necessary stuff (e.g. Vocab) alongside the model, and doesn't require you to know the model's architecture when you deserialize (contrary to e.g. loading pre-trained weights).

I thought it makes sense to share that. If you like it I can contribute a bit more e.g. for text classifier as well. Open to suggestions as well (maybe there's already a better way to do that, but I missed it).


sgugger:
Please amend your PR to be able to merge with the latest big refactor. If it's easiest, close this one and reopen another. Also, don't forget to sign the CLA or we won't be able to merge, thanks!

pzelasko:
Thanks. I'll try to do it by the end of the week.

sgugger:
Also you might want to use our processor to handle the tokenization and numericalization so that we avoid dupe code.

pzelasko:
It seems that I've fixed it (the test failed in seemingly not relevant part of the codebase - correct me if I'm mistaken). I'm still not sure if there's a better way to decouple batching + masking from `NumericalizeProcessor`. What do you think?

sgugger:
Hi there, sorry it took us so long to reply. I am closing this, not because this isn't useful, but because we think it would be best in a repo separate from fastai. We can then link to it in our docs so that interested users can install it and use it.

PR #1131, opened 2018-11-11
Single line update to basic_train.py to calculate weighted average of

pford221:
Allows for weighted average of validation loss.

Single line update to `basic_train.py` to account for observation weights in validation loss. Added a notebook observation-weights to docs_src/ to demonstrate the need for this.

See this post:
https://forums.fast.ai/t/developer-chat/22363/426?u=patrick


pford221:
Not sure why this now failing. It did pass initially, I think. I configured `nbstripout` and it appears to have stripped some output of some jupyter notebooks in `old/tutorials/...' based on the cell execution counts.

jph00:
Thanks for your contribution. docs_src is just for docs, so please don't add a nb there. Instead, please write some tests to show the need for this - i.e. the tests would fail without your change, and pass with them.

Also, please remove the extra 2 notebook changes that you seem to have accidentally included in your PR.

sgugger:
Just reviewed this, with the last changes in fastai, you shouldn't need this change anymore if you write your metric/validation loss as callbacks (see the [doc page of metrics](https://docs.fast.ai/metrics.html#Creating-your-own-metric) for an example).

PR #1132, opened 2018-11-12
Test for  image cleaner

noklam:
Add test for widgets, related to the lesson2 download notebook.

test should fail when wrong dataset is passed in, where lenght should not match. If a databunch is passed, it will fail as well as it does not have a length properties.
https://github.com/fastai/course-v3/pull/93


noklam:
Fixed earlier typo,  use fixture to initialize databunch for all test,  not sure if this is the best way assert fail case.

sgugger:
Sorry it took me a bit of time to get to this. First let's not change __all__ in image_cleaner, you can manually import ImageCleaner for your test with
```
from fastai.widgets.image_cleaner import ImageCleaner
```
Then, I think only the succeding test with the right length is necessary. Could you amend your PR for this? Thanks!

noklam:
No problem,  totally understandable you guys are busy on the lesson as we move to text and tabular.  Thanks for the comment @sgugger.

I can explain a little bit.  I think the test for too long and short may combine to 1 test,  but it is still need.

For ImageCleaner,  I add an assert len statement,  which prevent user pass in unmatched dataset/indexes pair.  For example,  if I pass in indexes for validation set and training dataset,  the class will have no problem to accept that,  as long as _get_item does not throws an error.  The result is user deleting the wrong files and probably unaware of it,  as we use the ipython widget to click only.

There is also a test fail case for a more complicated problem. As the FileDeleter change to ImageDeleter,  the api has slightly changed. In course-v3
 there is a time where ImageDeleter swap with FileDeleter but the argument was not update,  as a result you can pass in a databunch but not dataset.  Due to the fact that I have an assert statement in ImageCleaner,  it will throws Databunch does not have attribute len error,  so I think I need a fail case test just to not mixing with the wrong dataset problem.

In short,  test unmatched length for the assertion error correctly thrown.  And a expect fail case for wrong input type.

noklam:
@sgugger Sorry I have done a force-pushed, as I accidentally bring in a lot of commits when I merge from master.
I saw the ImageDeleter and ImageRelabeler is combine into ImageCleaner, so seems that I do not have to change anything and unit tests are fine.
 I combine 2  length assertion tests into 1,  found there is a missing import for display as well.

sgugger:
Thanks!

PR #1135, opened 2018-11-12
Fix the QRNN performance issue when no split_fn is provided.

PiotrCzapla:
The PR adds two tests but they can be run only if you have a cupy installed to pinpoint the issue.
It also fixes the issue by calling reset() during the init of WeightsDrops:
https://github.com/fastai/fastai/pull/1135/commits/7706e228d938222640faa35b448ca52e282434a1

But It isn't the best idea since it may cause case-cading resets during initialization, so I've changed it to simply execute relevant part of the reset function without propagating the reset call it down the tree.

https://github.com/fastai/fastai/pull/1135/commits/d60adca369f6e548a494109a849ea5ebb1061a61


PiotrCzapla:
replaced with #1138

Issue #1136, opened 2018-11-12
Mixup learner predict throws `reduction` argument error


bharadwaj6:
Predictions on learner with mixup applied throws reduction argument error instead of giving predictions.

In any notebook:

```
learn = create_cnn(data, models.resnet18, metrics=error_rate)
learn.fit_one_cycle(1)
learn_mx = mixup(learn=learn) # either this or using learn.mixup() - both result in same error
learn_mx.fit_one_cycle(1, max_lr=slice(1e-2))

interp = ClassificationInterpretation.from_learner(learn)
```

results in:

```
TypeError                                 Traceback (most recent call last)
<ipython-input-22-aa7f7b70a42b> in <module>
----> 1 interp = ClassificationInterpretation.from_learner(learn)

~/projects/fastai/fastai/vision/learner.py in from_learner(cls, learn, ds_type, sigmoid, tta)
    113     def from_learner(cls, learn:Learner, ds_type:DatasetType=DatasetType.Valid, sigmoid:bool=None, tta=False):
    114         "Create an instance of `ClassificationInterpretation`. `tta` indicates if we want to use Test Time Augmentation."
--> 115         preds = learn.TTA(with_loss=True) if tta else learn.get_preds(ds_type=ds_type, with_loss=True)
    116         return cls(learn.data, *preds, sigmoid=sigmoid)
    117 

~/projects/fastai/fastai/basic_train.py in get_preds(self, ds_type, with_loss, n_batch, pbar)
    209         lf = self.loss_func if with_loss else None
    210         return get_preds(self.model, self.dl(ds_type), cb_handler=CallbackHandler(self.callbacks),
--> 211                          activ=_loss_func2activ(self.loss_func), loss_func=lf, n_batch=n_batch, pbar=pbar)
    212 
    213     def pred_batch(self, ds_type:DatasetType=DatasetType.Valid, pbar:Optional[PBar]=None) -> List[Tensor]:

~/projects/fastai/fastai/basic_train.py in get_preds(model, dl, pbar, cb_handler, activ, loss_func, n_batch)
     37     res = [torch.cat(o).cpu() for o in
     38            zip(*validate(model, dl, cb_handler=cb_handler, pbar=pbar, average=False, n_batch=n_batch))]
---> 39     if loss_func is not None: res.append(calc_loss(res[0], res[1], loss_func))
     40     if activ is not None: res[0] = activ(res[0])
     41     return res

[...]

PR #1137, opened 2018-11-12
bugfix: set reduction attribute as none on loss fn if nothing is specified

bharadwaj6:
Possible fix for #1136 

Previously it was passed as a kwarg to the loss function, but mixup loss function doesn't have setters for the same.
Pytorch core loss functions seem to import from `nn._Loss` which has the setters for `reduction` parameter, so inheriting `MixupLoss` from that instead of `nn.Module` seems like another way to fix this bug. 

Setting the attribute explicitly when it doesn't exist seemed like a better approach, which is why the changes.


sgugger:
Not sure this is the right way to do it. I think the reduction attribute should be passed to the loss function inside MixupLoss instead. Closing this, the issue is there and I'll try to solve it today or tomorrow.

bharadwaj6:
@sgugger Have created another PR with suggested changes at #1163. Hope I got the details right this time.

PR #1138, opened 2018-11-12
Fixes the QRNN performance issue when no split_fn is provided during learner construction.

PiotrCzapla:
The PR adds two tests but they can be run only if you have a cupy installed to pinpoint the issue.
It also fixes the issue by calling reset() during the init of WeightsDrops:
7706e22

But It isn't the best idea since it may cause case-cading resets during initialization, so I've changed it to simply execute relevant part of the reset function without propagating the reset call it down the tree.
d60adca

it replaces this PR https://github.com/fastai/fastai/pull/1135


sgugger:

sgugger:
I tried to move the reset call in post_init of Sequential as it would be cleaner to do it there. I'll try and revert if that breaks anything. Thanks for your PR!

PiotrCzapla:
> I tried to move the reset call in post_init of Sequential as it would be cleaner to do it there. I'll try and revert if that breaks anything. Thanks for your PR!

I've thought that `__post_init__` works only in `dataclasses` and `nn.Module` is not one of them, or maybe I'm missing something. But either way, since the reset call is fixed I can run it if needed my self.

sgugger:
Yeah, it doesn't properly work. I'll edit and change soon-ish, I'm currently fixing the tests.

PR #1139, opened 2018-11-12
Combining build-docs and update-nbs scripts

bearpelican:
@sgugger @stas00 

Putting this here to check if this is what you guys had in mind.

previous build-docs command becomes -
tools/build-docs --no-update-line-num --no-update-nb-links

previous update-nbs command becomes -
tools/build-docs

1061b3d: fixes moving div when different versions of nbconvert are run. We will need to run html update script one last time to fix this. This change occured in [5.4.0](https://github.com/jupyter/nbconvert/commit/82d6979fb9c1a593b661d8bff362b099bcd80d8d)


stas00:
looking good @bearpelican, thank you!

Here are some nuances to discuss:

I'm not sure about:

```
previous build-docs command becomes -
tools/build-docs --no-update-line-num --no-update-nb-links
```
That's too much to type. Can we have one letter flag that combines both options? But let's ask the main user first of his needs:

@sgugger - which of the current scripts `build-docs`, `update-nbs` do you use most of the time? I think the default (no flags) should be for the setup used most of the time, the second most used setup should have a single short flag, and then we have the long options for the rest of the rare cases.

This is also a good moment to rename the combined script if you feel that `build-docs` is not it.

stas00:
and thank you for tracking that nbconvert version issue, @bearpelican! I have now updated mine. Perhaps we need to pin that version instead (if it's on pypi and conda).

bearpelican:
Awesome. Thanks for the feedback!

Totally right about too many flags. Old build-docs script has been changed to:
tools/build-docs --html-only

I combined --update-line-num and --update-nb-links into one flag because it seemed like they would always be run together. I can change it back though if that's not the case.

bearpelican:
@stas00 by pinning the version, do you mean setting it ==5.4 instead of >=5.4?

stas00:
> Awesome. Thanks for the feedback!
> 
> Totally right about too many flags. Old build-docs script has been changed to:
> tools/build-docs --html-only
> 
> I combined --update-line-num and --update-nb-links into one flag because it seemed like they would always be run together. I can change it back though if that's not the case.

And I'd add `-h` to make it even shorter, as a short alias, e.g. in `tools/trust-origin-git-config` we have:

```
parser = argparse.ArgumentParser()
parser.add_argument('-e', '--enable',  action="store_true", help="Trust repo-wide .gitconfig (default action)")
parser.add_argument('-d', '--disable', action="store_true", help="Distrust repo-wide .gitconfig")
parser.add_argument('-t', '--test',    action="store_true", help="Validate repo-wide .gitconfig config")
```

stas00:
> @stas00 by pinning the version, do you mean setting it ==5.4 instead of >=5.4?

I meant `nbconvert>=5.4` in requirements in setup.py and conda/meta.yml

stas00:
Let's wait a little bit with merging this, since I hear @sgugger is busy rewriting all of the docs, so we don't want to drop a potential surprise if suddenly it works differently, or even unfamiliar, since he probably is using this functionality a lot as we speak.

sgugger:
tools/update-nb was what I was using all the time so having it changed to build-docs seems clearer.
Merging this so I can test today, but if one of you can update the docs to reflect the change, that would be great.

stas00:
will do.

PR #1141, opened 2018-11-12
bugfix: TextClasDataBunch.from_df and TextClasDataBunch.from_folder  ignore `vocab` parameter

tchambon:
Fix of issue fastai#1140: TextClasDataBunch.from_df and TextClasDataBunch.from_folder  ignore `vocab` param


sgugger:
Thanks for flagging this!

PR #1143, opened 2018-11-12
Added a few unit tests and a minor test fix

movefast:
Added a path_if_exists check to test_download_url and several other new unit tests.


sgugger:
Thanks!

PR #1144, opened 2018-11-12
Image2image

313v:
Add support for image to image training:

- normalization of y variable can be turned on w/ tfm_y in normalize call
- standardize do_y to tfm_y
- support optional y_kwargs (to override y sz for example)
- support optional separate y_tfms, if none use same tfms as x assuming y_tfm is true
- for show_results, denormalize y if it is an image


sgugger:
Note that there a no datasets anymore.

PR #1150, opened 2018-11-13
Added various tests for convenience functions in torch_core

marii-moe:
Added various tests for torch_core.py. Should bring coverage up to 100% if to_fp6 test is also approved in other pull request. 

Tests cover various functions from determining number of children of a layer(think sequential). Splitting a model at a particular layer. Setting bn to eval only mode when requires_grad = false. Changing ap array to a tensor. Calculating loss given two tensors and a loss function. Test for monkey patching array-like-ness into Tensors(use was for matplotlib).


sgugger:
Thanks!

PR #1152, opened 2018-11-13
SqeezeNet meta data for create_cnn function

devforfu:
Based [on forum's discussion](https://forums.fast.ai/t/cannot-use-squeezenet-with-create-cnn/28661/15), the PR makes an attempt to bring the support of SqeezeNet architecture into the library. Now you're able to use the architecture like ResNets:
```python
learn = create_cnn(data, models.squeezenet1_1)
```
However, I've chosen the layers groups split quite arbitrary (splitting model on maxpool layers) so would be glad to see your comments if there is a better configuration.


sgugger:

devforfu:
Ok, sure, will do.

devforfu:
@sgugger Removed all training from the tests except one place with `simple_cnn` and merged with the most recent `master`. Now should be better.

devforfu:
Seems there is a bug in tests, going to fix soon.

devforfu:
@sgugger Seems like there is a timeout issue, however, I don't train any model in the tests I've added. Do you think it is worth to change something in the tests to pass through the limit?

fastai/vision/learner.py
```diff
@@ -13,14 +13,22 @@
 def _default_split(m:nn.Module): return (m[1],)
 # Split a resnet style model
 def _resnet_split(m:nn.Module): return (m[0][6],m[1])
+# Returns body of sqeezenet style model
+def _squeezenet_body(m:nn.Module): return m.features
```
jph00: I don't think this should be necessary any more.

tests/test_vision_train.py
```diff
@@ -87,3 +95,8 @@ def test_preds(learn):
 
 def test_lrfind(learn):
     learn.lr_find(start_lr=1e-5,end_lr=1e-3, num_it=15)
+
+@pytest.mark.parametrize('arch', [models.resnet18, models.squeezenet1_1])
+def test_models_meta(mnist_tiny, arch, image):
```
jph00: By default this uses a pretrained model, which will try to download the whole weight file from the internet. We want our tests to run quickly and without internet, so you shouldn't use pretrained models in your tests unless you mark them as 'slow'.

devforfu: Ok, understood. The only reason to use the `torchvision` models was to verify that `create_cnn` still works. Would you propose to mark them 'slow' or replace all together with something computationally efficient?

devforfu:
I've tried to address the issues. I think now the `create_cnn` should look better. How do you think, should we change anything else? Are these changes are still relevant to the library? Also, I've included the following test:
```python
@pytest.mark.slow
@pytest.mark.parametrize('arch', [models.resnet18, models.squeezenet1_1])
def test_models_meta(mnist_tiny, arch, zero_image):
    learn = create_cnn(mnist_tiny, arch, metrics=[accuracy, error_rate])
    pred = learn.predict(zero_image)
    assert pred is not None
```
Note the last line. I am not really sure what to assert here so I am just making sure that the output is not `None`. The main goal of this test is to make sure that architectures are instantiated correctly. Actually, the test has passed if we just don't get any exception so it could be `assert True` as well.

fastai/vision/learner.py
devforfu: Agree, removed!

tests/test_vision_train.py
devforfu: Done!

sgugger:
Seems good, thanks!

PR #1159, opened 2018-11-14
Fix inconsistent argument name data block

noklam:
As shown in the lesson4, it should be `cols` instead of `col`, I found there is more cols/col inconsistent.


sgugger:
As this argument is called by name in all the tests (and probably a few users), this can't be changed easily now. Closing this as I feel it's more the notebook that should be fixed than the code.

PR #1161, opened 2018-11-14
Addded tests for verifying images

marii-moe:
Added tests for verifying images. Had to use `from fastai.vision.data import verify_image` because `verify_image` was excluded from `__all__`

I was not able to test `verify_image` using `verify_images`, because the coverage tool looses track of test coverage when `with ProcessPoolExecutor` is used in `verify_images`


sgugger:
Thanks!

PR #1163, opened 2018-11-14
added reduction attribute in mixuploss function

bharadwaj6:
For #1136 

Added reduction attributes to the MixupLoss function.


sgugger:
Way better, thanks!

PR #1164, opened 2018-11-14
y transform support in datablocks

313v:
exposes do_y to normalize on databunch

adds transform_labels to LabelList which allows y specific overrides of kwargs like sz or even a completely separate pipeline of tfms

i'm using this to do super resolution

### DIFF
--- a/fastai/data_block.py
+++ b/fastai/data_block.py
@@ -346,8 +346,8 @@ class LabelList(Dataset):
             else:                 x,y = self.item   ,0
             if self.tfms:
                 x = x.apply_tfms(self.tfms, **self.tfmargs)
-                if self.tfm_y and self.item is None:
-                    y = y.apply_tfms(self.tfms, **{**self.tfmargs, 'do_resolve':False})
+            if self.tfm_y and self.item is None:
+                y = y.apply_tfms(self.tfms_y, **{**self.tfmargs_y, 'do_resolve':False})
             return x,y
         else: return self.new(self.x[idxs], self.y[idxs])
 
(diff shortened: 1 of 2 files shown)


313v:
example of usage (note call to transform_labels and do_y in normalize)

def open_grayscale(fn):
    x = PIL.Image.open(fn)
    return Image(pil2tensor(x,np.float32).div_(255)[0:1])   

class SuperResLabelList(ImageItemList):
    def __init__(self, items:Iterator, **kwargs):
        super().__init__(items, **kwargs)
        self.loss_func,self.create_func = combo_loss,open_grayscale
        

    def new(self, items, classes=None, **kwargs):
        return self.__class__(items, **kwargs)


class SuperResItemList(ImageItemList):
    def __post_init__(self):
        super().__post_init__()
        self._label_cls = SuperResLabelList
        self.loss_func = combo_loss
        self.create_func = open_grayscale

valid_pct = 0.10
src = (SuperResItemList(lr_names_full).
       random_split_by_pct(valid_pct).
       label_from_func(match_hr_fn))
      

def get_data(src, bs, sz_lr, scale=4, tfms=None, **kwargs):
    sz_hr = sz_lr*scale
    salk_stats = ( [0.10], [0.20])
    if tfms is None: tfms = get_transforms() 
    data = (src.transform(tfms, size=sz_lr)
            .transform_labels(size=sz_hr)
            .databunch(bs=bs, **kwargs) #, num_workers=0)
            .normalize(salk_stats, do_y=True)
           )
    return data

Issue #1173, opened 2018-11-15
Getting UnicodeDecodeError while trying to read text files in TextFilesList


ameyyadav09:
**Describe the bug**
UnicodeDecodeError: 'charmap' codec can't decode byte 0x9d in position 1757: character maps to <undefined>

**To Reproduce**
```
path = untar_data(URLs.IMDB)
data_lm = (TextFilesList.from_folder(path)                           
           #Inputs: all the text files in path
            .filter_by_folder(include=['train', 'test']) 
           #We may have other temp folders that contain text files so we only keep what's in train and test
            .random_split_by_pct(0.1)
           #We randomly split and keep 10% (10,000 reviews) for validation
            .label_for_lm()           
           #We want to do a language model so we label accordingly
            .databunch())
```

**Additional context**
Below is the stack trace
```
---------------------------------------------------------------------------
UnicodeDecodeError                        Traceback (most recent call last)
<ipython-input-5-49723f1480eb> in <module>
      3             .filter_by_folder(include=['train', 'test'])
      4            #We may have other temp folders that contain text files so we only keep what's in train and test
----> 5             .random_split_by_pct(0.1)
      6            #We randomly split and keep 10% (10,000 reviews) for validation
      7             .label_for_lm()

~\AppData\Local\Continuum\anaconda3\lib\site-packages\fastai\data_block.py in _inner(*args, **kwargs)
    260             self.valid = fv(*args, template=self.train.y, **kwargs)
    261             self.__class__ = LabelLists
--> 262             self.process()
    263             return self
    264         return _inner

~\AppData\Local\Continuum\anaconda3\lib\site-packages\fastai\data_block.py in process(self)
    301     def process(self):
    302         xp,yp = self.get_processors()
--> 303         for ds in self.lists: ds.process(xp, yp)
    304         return self
    305 

~\AppData\Local\Continuum\anaconda3\lib\site-packages\fastai\data_block.py in process(self, xp, yp)
    363     def process(self, xp=None, yp=None):
    364         "Launch the preprocessing on `xp` and `yp`."
--> 365         self.x.process(xp)
    366         self.y.process(yp)
    367         return self

~\AppData\Local\Continuum\anaconda3\lib\site-packages\fastai\data_block.py in process(self, processor)
     59         if processor is not None: self.processor = processor
     60         if not is_listy(self.processor): self.processor = [self.processor]
---> 61         for p in self.processor: p.process(self)
     62         return self
     63 

[...]

PR #1174, opened 2018-11-15
added fix for UnicodeDecodeError in data.py

ameyyadav09:
[#1173](https://github.com/fastai/fastai/issues/1173), I followed the stack trace and found it was a simple encoding issue and was able to resolve it by setting encoding parameter to 'utf-8' in open method.


sgugger:
Thanks!

PR #1176, opened 2018-11-15
Recorder added mets didn't work

holli:
Recorder has `add_metrics(self, metrics)`  and `add_metric_names(self, names)` but they didn't have affect because `'_added_met_names'` was not used anywhere after commit in: 
 https://github.com/fastai/fastai/commit/6b17961574d738d4fbc8b8b8f6268eebca5f61f9#diff-0730afdfa67f9712e46ad7866b0123f8L202

(noticed when examples in https://forums.fast.ai/t/allow-for-more-than-one-output-for-loss-and-metric/21991/34 didn't work)


sgugger:
Thanks! Note that now that metrics are callbacks, the `add_metrics` will probably be deprecated in a future version. Can you try to implement your example from the forum as a metric now and see if we need to keep this alternative?

holli:
Yea I've noticed the callback change. It seems an elegant approach!

That original approach doesn't work as the callbacks are expected to return a single value. Of course its possible to create function for all the extra metrics but its a bit complicated approach..

Then again if callbackhandler (these two lines) would accept a tuple in the return. Then all callback metrics could return multiple values if they want to do so. 

- https://github.com/fastai/fastai/blob/master/fastai/callback.py#L251
- https://github.com/fastai/fastai/blob/master/fastai/callback.py#L193

That could be easy approach but is there some downside and do others need that kind of features?

sgugger:
That would mess up the recorder then. I'll think about it.

holli:
Would it?  I didn't mean that it would add list inside the list but add all values.

```
m = met.metric
if is_listy(m): self.state_dict['last_metrics'] += m
else: self.state_dict['last_metrics'].append(m)
```

PR #1178, opened 2018-11-15
Add torchvision models

Gokkulnath:
Added All the Torchvision models. Tried running one epoch of fit_one_cycle on one of each type of architecture.


Gokkulnath:
inception_v3 architecture alone fails with current code base

Gokkulnath:
@sgugger  thanks for your reply. Sounds better to have it our self whenever needed.
Will close this.

PR #1189, opened 2018-11-16
fix callback logic

radekosmulski:
There were issues with the earlier version of Tracker callbacks. `ReduceLROnPlateauCallback` was never getting the `self.wait` reset and the logic of `patience` was wrong and inconsistent between the callbacks. For instance, with patience=0 `ReduceLROnPlateauCallback` was never triggered and patience 0 and 1 were equivalent on the `EarlyStoppingCallback`.


sgugger:
Thanks for flagging!

PR #1190, opened 2018-11-16
Return files from get_files() in sorted order

cwerner:
A simple fix to get sorted class names...

If one imports classes from folders with names in the form of:
```
data/subclassA_itemsA 
     subclassA_itemsB 
     subclassB_itemsA 
     subclassB_itemsB 
     subclassB_itemsC 
     ...
```

I would prefer the class numerical representation (data.c) to follow the sorted order of the names (data.classes).

Currently I might get 
```
data.classes = [subclassB_itemsA, subclassB_itemsC, subclassA_itemsA, ...]
```


sgugger:
Thanks, but I'm not sure this is the right fix for this. `get_files` is already very slow compared to old fastai and I'm afraid a sort will only add to it. Closing this, but you can open a new PR if you think of a better way. 
Note that in the meantime, you can always pass your list of classes to override the default behavior.

cwerner:
No worries. How would I pass the list of classes with the new API?

sgugger:
When you call your label function, you can pass `classes=...` as kwarg.

PR #1196, opened 2018-11-16
Update tabular.ipynb

navneetkrc:
From Release 1.0.25 we have 
**tabular_learner** in place of **get_tabular_learner**.

### DIFF
--- a/examples/tabular.ipynb
+++ b/examples/tabular.ipynb
@@ -239,7 +239,7 @@
     }
    ],
    "source": [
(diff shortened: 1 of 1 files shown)


navneetkrc:
Can someone guide me why the pipelines failed for this minor modification?
It says an assertion failed for the size of Tensor Object (9==8). 
I am just updating the function name as per the latest release.

navneetkrc:
Initially one more Import needs to be done.
Renames for Harmonization is missing here.
@sgugger 
You already have updated this file as https://github.com/fastai/fastai/blob/master/docs_src/tabular.ipynb

PR #1198, opened 2018-11-17
Add model summary funcitons

noklam:
Open this PR for adding a model summary function that could take a learner as input and print nicely formatted summary leveraging the existing hooks. 

Timeline: Intend to finish by this weekend

Intended  functionality:
model_summary(learner)

This list is created for tracking the progress
Progress:
- [x] Print summary which takes a list of tuples as argument and print with nice format
- [x] Return a correct list of tuple object from learn.model sequential
- [x] Breakdown if element of model is Sequential 
- [x] Figure out how to get correct size for adaptive layers.
- [x] Combine both to print out a summary for a learn.model with hierarchy of Sequentials.


noklam:
have a little blocker,  `model_sizes` does not work on a nn.Module but Sequential only, need to find a way to record the all shape of a Block instead of input and output only.

def model_sizes(m:nn.Module, size:tuple=(64,64), full:bool=True) -> Tuple[Sizes,Tensor,Hooks]:

noklam:
the model_summary sit inside callbacks module, feel free to move it to a more appropriate location.

Example of usage:
https://gist.github.com/noklam/bdb4e6e3a73f89c7cbbbef7d42140a0b

image

Trial on vision.learner.ipynb
image

sgugger:
Thanks!

sgugger:
Just ran a few tests, it works great on vision model. On text or tabular model that expect a different kind of input... not so much. If you have time to look at this issue, I think maybe the dummy should be inferred from the dataloader, so this function should take a learner and not a model.

noklam:
@sgugger I will start looking at the lesson for text & tabular, see if I can add a few unit tests and make it work for text & tabular along the way.

pouannes:
Also maybe the ability to see what's the core of the architecture and where's the head ?

noklam:
@oostopitre @pouannes
Discussion of adding feature are better on forum as other may contribute to this too. 
Thank for your input. 
https://forums.fast.ai/t/keras-like-summary-of-a-model/28845/14?u=nok

PR #1204, opened 2018-11-18
y transform support in datablocks

313v:
exposes do_y to normalize on databunch

adds transform_labels to LabelList which allows y specific overrides of kwargs like sz or even a completely separate pipeline of tfms

i'm using this to do super resolution

### DIFF
--- a/fastai/data_block.py
+++ b/fastai/data_block.py
@@ -332,6 +332,14 @@ class ItemLists():
         if self.test: self.test.transform(tfms[1], **kwargs)
         return self
 
+    def transform_labels(self, tfms:Optional[Tuple[TfmList,TfmList]]=(None,None), **kwargs):
+        if not tfms: tfms=(None,None)
+        self.train.transform_labels(tfms[0], **kwargs)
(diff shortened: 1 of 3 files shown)


fastai/data_block.py
```diff
@@ -419,6 +427,18 @@ def from_lists(cls, path:PathOrStr, inputs, labels)->'LabelList':
     def transform(self, tfms:TfmList, tfm_y:bool=None, **kwargs):
         "Set the `tfms` and `` tfm_y` value to be applied to the inputs and targets."
         self.tfms,self.tfmargs = tfms,kwargs
-        if tfm_y is not None: self.tfm_y=tfm_y
+        if tfm_y is not None:
+            self.tfm_y=tfm_y
+            self.tfms_y=tfms
```
sgugger: Don't we need to pass the kwargs to self.tfmargs_y here? I think this is going to make camvid and object detect fail.

313v: yes you're right, will fix that

PR #1206, opened 2018-11-18
New Feature - Facets Dive Exploration Widget

pgollakota:
New feature to explore an `ImageDataBunch` using [PAIR's Facets Dive exploration](https://pair-code.github.io/facets/). You may have seen the excellent visualization of the [Quick, Draw! dataset in Facets Dive](https://pair-code.github.io/facets/quickdraw.html). 

Facets Dive enables rapid exploration of images. This is particularly helpful in quickly evaluating the images that have high loss, low probability score and more.

facets_dive_01

**Caveat**: Facets Dive (the source library, not this widget in the PR) has still not been ported to work with Jupyter Lab. It only works with _Jupyter Notebook_. There is an [outstanding issue in the Facets repo](https://github.com/PAIR-code/facets/issues/113) to fix the Jupyter Lab incompatibility.

Docs, example and tests have been added. The tests ensure that the right metadata payload gets generated and that a sprite image of right dimensions is generated in the right folder.

See Also: [FastAI forum thread](https://forums.fast.ai/t/new-feature-proposal-facets-dive-interactive-exploration/30534)

Note: Don't be scared by the file count. Most of them are PNGs for tests. The actual changes are only 3 files (1 py, 1 notebook, 1 tests).


sgugger:
Hi there, sorry it took us so long to reply. I am closing this, not because this isn't useful, but because we think it would be best in a repo separate from fastai. We can then link to it in our docs so that interested users can install it and use it.

PR #1207, opened 2018-11-18
Add Fbeta_binary to metrics.py

wy-q:
Add `Fbeta_binary` metric for binary classification to metrics.py, as existing `fbeta`  is meant for multiclassification.

Made @sgugger's suggested changes in the [forum thread](https://forums.fast.ai/t/f1-score-as-metric/30370/16):
1)removed y_pred = y_pred.softmax(dim = 1)
2)named class with a capital and used a @dataclass

[Simple test for text classification](https://github.com/wy-q/Fbeta_binary/blob/master/Fbeta_binary_text_classification_test.ipynb)

### DIFF
--- a/fastai/metrics.py
+++ b/fastai/metrics.py
@@ -2,7 +2,7 @@
 from .torch_core import *
 from .callback import *
 
-__all__ = ['error_rate', 'accuracy', 'accuracy_thresh', 'dice', 'exp_rmspe', 'fbeta']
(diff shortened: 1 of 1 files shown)


wy-q:
👌🏻

sgugger:
Thanks! I just added a clas argument that defaults to 1, to make it usable if we want the fbeta of a given class in a problem with more than two classes.

wy-q:
Hi @sgugger 
Not sure if I should add a `self.clas` as mentioned [here](https://github.com/fastai/fastai/issues/1271)

PR #1208, opened 2018-11-18
Fix an issue in ImageItemList.from_csv method

mukesh-k-jangid:
There is a bug in ImageItemList.from_csv method. When we pass a string instead of a Path object to the path parameter, this causes an error.

Fix for this bug is to convert string into Path object.


mukesh-k-jangid:
It's a pleasure to be able to contribute to the fastai library.

PR #1210, opened 2018-11-19
Added some metrics

SvenBecker:
I added some additional metrics which I think might be quite useful. Those implementations are closely related to the Scikit-Learn and TensorFlow/Keras implementations.

### Classification (binary and multiclass):

- Top K accuracy
- Balanced accuracy (might be useful for classes imbalance)
- Confusion matrix (mostly the same as the ImageClassifierInterpreter method confusion_matrix)
- (Cohen-) Kappa score
- Precision
- Recall
- Matthews correlation coefficient
- Fbeta_score

### Regression:

- Mean squared error (MSE)
- Mean abolute error(MAE)
- Root mean squared error (RMSE)
- Mean squared logarithmic error (MSLE)
- Explained variance
- R2 score

Input-dimensions are inline with the currently supported metrics.
Functionality check and comparison with sklearn implementations: https://gist.github.com/SvenBecker/041c988c1fbbb4d846e5e5f9ff3b066a


navneetkrc:
It would be really nice to have these metrics.
Do you have any example of using Top-K accuracy for the newer version?
TIA

navneetkrc:
Hey Sven,
These metrics would be really nice addition to the library, please consider adding/updating the docs as well. 
I used the link provided for my scenario as well. Happy that I could use readymade solution by you.
Thanks a lot for this 👍

SvenBecker:
Hey,
glad you like it. I'm currently on a business trip. I will check it/add them when I'm back.

sgugger:
Thanks for this work. There's one main issue IMO for metrics that are not an average over batches (recall, precision for instance). If they are used as a metric by the a fastai user, they will return a wrong result if you code them like this, which can be very misleading.

That's why metrics are Callback behind the scenes, to allow to an exact computation. There is an example  with an Fbeta class in the last version of metrics.py, and another one in the [doc page of metrics](https://docs.fast.ai/metrics.html).

SvenBecker:
Yes I will. But there was a family situation, thus I was and still am a little bit busy. I will try to look into this over the next week. Did some metrics implementations using Callback but didn't test them yet.

SvenBecker:
Ok I assume it should work now. Some of the implementations are very similar to the ones provided before but now using the Callback class.

I added different averaging options to the metrics which rely on the computation of a confusion matrix. Those are inline with the ones provided by sklearn ("binary", "micro", "macro", "weighted" or None for no averaging at all).

sgugger:
This looks great, thanks a lot! I'm just thinking we could refactor the common bits between Recall, Precision and FBeta. Maybe Recall and Precision could be subclasses of FBeta, where all the work would be done? (a bit like you made them subclass ConfusionMatrix)

SvenBecker:
Hey @sgugger. Thank you for your feedback. Well I did something similar. There were some issues related to the averaging functions (setting average to micro or weighted was not working properly for FBeta) so I came up with a little workaround using another subclass 'CMScores'.

sgugger:
Thanks a lot, this looks great!

PR #1215, opened 2018-11-19
added a unit test for split_bn_bias

movefast:
Added a unit test for split_bn_bias in torch_core.py 

@sgugger I saw this todo sitting on top of the function i was writing test for but didn't quite get the idea here, could you help explain it to me so I can fix it along with the test
`baba12c2 (Ubuntu 2018-10-13 13:40:05 +0000 146) fastai/torch_core.py`
`#TODO: add the test to put bias with bn layers`
`def split_bn_bias(layer_groups:ModuleList)->ModuleList:`


sgugger:
It's not a unit test to add, it's to add a test in the function to separate BatchNorm layers and Bias parameters from the rest (so that we don't apply weight decay to them).

PR #1216, opened 2018-11-20
Update tabular.ipynb

navneetkrc:
From Release 1.0.25 
"get_tabular_learner" is updated as "tabular_learner"


sgugger:
Good catch!

navneetkrc:
Feels good to contribute. My first Pull request ever and that too for FASTAI.
I am going through the lectures by myself so I should be able to contribute more.

PR #1219, opened 2018-11-21
Fix some typos in 'from_token

matiu2:
The arg names didn't match the variable names. Now they do. Also I've used `from_tokens` successfully.


sgugger:
Thanks for flagging this!

PR #1220, opened 2018-11-21
Fix `max_vocab` for `TextLMDataBunch`

jfilter:
It's currently not possible to pass a `max_vocab` parameter when creating a `TextLMDataBunch`. I added a test and also fixed it.


sgugger:
Thanks, I'm merging mostly for the test as I don't like to pass needless kwargs around. Will change that little bit ;)

PR #1225, opened 2018-11-22
fix bug. progress bar does not accept cuda tensors

simonm3:

### DIFF
--- a/fastai/train.py
+++ b/fastai/train.py
@@ -54,7 +54,7 @@ class ShowGraph(LearnerCallback):
             iters = range_of(rec.losses)
             val_iter = np.array(rec.nb_batches).cumsum()
             x_bounds = (0, (n_epochs - len(rec.nb_batches)) * rec.nb_batches[-1] + len(rec.losses))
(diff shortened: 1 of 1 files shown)


sgugger:
I'm not sure why this is useful as losses are put on the cpu before being recorded on [this line](https://github.com/fastai/fastai/blob/3079d077d4af83824839923210c332451250362f/fastai/callback.py#L221). To double-check, after a training, type `learn.recorder.losses` gives me only CPU tensors.

simonm3:
The line I changed creates a new Tensor. If the default tensor type is set
to cuda.FloatTensor then it creates a new cuda tensor and the progress bar
fails. It is useful to set the default tensor type to cuda as it avoids
having to pass a device parameter around and litter your code with
.to(device). In virtually all applications I either want to run on cuda or
cpu rather than mixing them

simonm3:
The line you quote is adding losses to the smoothener. The losses are
separately added to the LossRecorder without forcing to cpu. My
learn.recorder has losses as a list of cuda tensors and val losses as just
a list of floats. I guess the training losses should also be a list of
floats. Will give it a try and resubmit.

PR #1228, opened 2018-11-22
fixes bug where progress bar fails if default tensor type is cuda

simonm3:

### DIFF
--- a/fastai/basic_train.py
+++ b/fastai/basic_train.py
@@ -301,7 +301,7 @@ class Recorder(LearnerCallback):
 
     def on_backward_begin(self, smooth_loss:Tensor, **kwargs:Any)->None:
         "Record the loss before any other callback has a chance to modify it."
-        self.losses.append(smooth_loss)
+        self.losses.append(smooth_loss.item())
         if self.pbar is not None and hasattr(self.pbar,'child'):
             self.pbar.child.comment = f'{smooth_loss:.4f}'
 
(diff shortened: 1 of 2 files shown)


sgugger:
Again, I'm not sure how you manage to get CUDA tensors recorded there since they are put on the CPU by the CallbackHandler.
The `.item()` method is very slow because it triggers a CUDA synchronization, which is why we don't use it for the losses.

simonm3:
It is because I have default tensor type of cuda. And they are not put on
the CPU by the CallbackHandler - I agree they should be!

torch.set_default_tensor_type(torch.cuda.FloatTensor) => Now all new
tensors are created cuda by default.
basic_train/25 => loss = cb_handler.on_backward_begin(loss) => this is a
cuda tensor
basic_train/302 is Recorder.on_backward_begin. This does NOT put the tensor
on the CPU.

If not item() then perhaps just use .cpu()? There does not seem to be any
logic in having training losses stored as tensors nor in passing tensors to
the progress bar. Especially as validation losses are already just stored
as list of floats not as tensors.

sgugger:
At line 25, the loss that is passed is on the GPU, as it can still be modified by other callbacks. At line 309, the loss that is given back by the callback handler is on the CPU, as it was sent to the Smoothener like this by the callback handler [here][https://github.com/fastai/fastai/blob/17ab600b4d530bca07e12100470fde9f505fdf38/fastai/callback.py#L227). 

Again, if I do any training (on the GPU), then ask for learn.recorder.losses, everything is on the CPU (as tensors, but sadly .item() seems to make a big loss in efficiency). I just retested it on master.

simonm3:
There is a comment in pytorch issues re speed of "item()". Says that it
shows up on profilers as slow because GPU calls are async. So it may be
that "item()" is not actually slowing things down or triggering a sync.
Instead it is waiting for a sync but other code carries on in parallel.
Hence it just looks slow.

sgugger:
Ah interesting point. I'll make timing tests of training with/without .item() to see if this is correct.

PR #1234, opened 2018-11-24
Updated the metrics.py file so that dice metric returns a value of 1 instead of nan

ksasi:
Currently the "dice" function in metrics.py is returning "nan" when intersect and union are both zeros.
Updated the  "dice" function in metrics.py so that 1 is returned when ever intersect and union are both zeros.


sgugger:
Thanks!

Issue #1230, opened 2018-11-23
DataLoader with batch_size=1 fails silently


amrav:
https://github.com/fastai/fastai/blob/ac45d71d4fab9430f0c195aeb7f2ad7c823d43a1/fastai/basic_data.py#L80

Why is `skip_size1` set to `True`? When batch size is set to 1 on a `DataLoader` passed into a `DataBunch`, where labels are scalars, the effect of this is that the generator is exhausted because it skips over every element, so the training is never run, and then it fails with (what appear to be) unrelated errors.

If batch size should not be set to 1 for some reason, would be great to document and fail with a helpful error message.

sgugger:
Pytorch will return an error with a batch size of 1 if you're using batchnorm in your model, so the option is there to skip the last batch if it's of size 1.
You're welcome to make a PR to suggest an appropriate warning or error message.

PR #1239, opened 2018-11-24
Few small fixes for edge cases that appears when one writes functional tests.

PiotrCzapla:
The following fixes would save me quite some effort when I was developing tests for ulmfit-multilingual. It is often useful to run models on very small data sets and currently this is not supported 

The content of this PR is as per commit messages: 
- Add warning regarding batch_size=1 and skip_size1
- Make SortishSampler work on datasets if size 1
- Make it possible to turn off multiprocessing to allow PyCharm to debug tests.

### DIFF
--- a/fastai/basic_data.py
+++ b/fastai/basic_data.py
@@ -44,6 +44,7 @@ class DeviceDataLoader():
 
     def __iter__(self):
         "Process and returns items from `DataLoader`."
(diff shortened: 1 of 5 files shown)


PiotrCzapla:
it fixes https://github.com/fastai/fastai/issues/1230

sgugger:
Will refactor the ToInt preprocessor differently, but merging this as a start.

PR #1244, opened 2018-11-26
feat: manage deleted and relabeled state from CSV (written as cleaned…

zcaceres:
….csv) to project directory.


zcaceres:
We had some git issues during collaboration, so this includes a copy of @fpingham's duplicate detector, but does not include his branch.

PR #1254, opened 2018-11-28
Fix bug with random permutation in TextDataBunch.from_csv()

cudawarped:
The current implementation of TextDataBunch.from_csv() misses a step when randomly sampling data for the training and validation sets.   Although the indices are permuted
`idx = np.random.permutation(len(df))`
these permuted indices are not used.

I am not sure what the most efficient way to do this in pandas is without testing, that said the pr contains the below simple change
`df = df.iloc[np.random.permutation(len(df))]`
which seems easier to grasp when quickly scanning through the source than
`train_df, valid_df = df.iloc[idx[cut:]], df.iloc[idx[:cut]]`


sgugger:
Thanks!

PR #1255, opened 2018-11-28
Fix a bug in ItemList.from_csv method.

mukesh-k-jangid:
There is a bug in ItemList.from_csv method. When we pass a string instead of a Path object to the path parameter, this triggers the bug.

Fix for this bug is to convert string into Path object.


sgugger:
Thanks for flagging!

PR #1257, opened 2018-11-29
Use ThreadPoolExecutor on Windows 

H4dr1en:
Calling parallel_trees function on windows crashes because it uses internally ProcessPoolExecutor that throws a BrokenProcessPool error.

Error thrown: 

> BrokenProcessPool: A process in the process pool was terminated abruptly while the future was running or pending.

This issue is fixed by rather using ThreadPoolExecutor on Windows systems


jph00:
Since internally you have the GIL I don't see that this is actually useful on Windows.

PR #1260, opened 2018-11-30
Fix multilabel classification for text

navjotts:
I think this broke in the refactor for Datablock (https://github.com/fastai/fastai/pull/1156)

Marking @sgugger : may be it was intended to set the `classes` consistently at 1 place somewhere else, but the **missing** thing which was making the multilabel classification work earlier was 
```
        if classes is None:
            if len(label_cols) == 0:   classes = [0]
            elif len(label_cols) > 1:  classes = label_cols
```
inside `TextDataset.from_df()` (which was refactored out)

### DIFF
--- a/tests/test_text_train.py
+++ b/tests/test_text_train.py
@@ -81,7 +81,7 @@ def text_df(n_labels):
     return df
 
 def test_classifier():
-    for n_labels in [1]: # , 8 - does not work for multilclass yet
+    for n_labels in [1, 8]:
         path = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'data', 'tmp')
         os.makedirs(path)
         try:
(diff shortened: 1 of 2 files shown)


tests/test_text_train.py
```diff
@@ -93,7 +93,6 @@ def test_classifier():
             assert len(data.train_dl) == math.ceil(len(data.train_ds)/data.train_dl.batch_size)
             assert next(iter(data.train_dl))[0].shape == (9, 2)
             assert next(iter(data.valid_dl))[0].shape == (9, 2)
-            classifier.fit(1)
```
navjotts: I have removed this temporarily as a `multi-classifier.fit()` would need different metrics

jph00:
Thanks @navjotts . There's currently a test failing in test_text_train . Could you take a look?

navjotts:
Fixed, PR updated

PR #1265, opened 2018-11-30
Fix windows compatibility issue

cudawarped:
In windows the default data type of a numpy array is Int32 not Int64 as on linux.  Because of this the following change from
`def process(self, ds:Collection):        ds.items = [self.process_one(item) for item in ds.items]`
to
`def process(self, ds:Collection):        ds.items = array([self.process_one(item) for item in ds.items])`
breaks fastai compatibility with windows.  

I know that the official docs have been changed to say that fastai is not compatible with windows, however given that
`type(ds.items[0]) == numpy.int64`
on linux, the following change to allow this to work again on windows should not hurt.
`def process(self, ds:Collection):        ds.items = array([self.process_one(item) for item in ds.items],'int64')`


jph00:
As you can see from the test, this change breaks many things.

PR #1268, opened 2018-12-01
Add a use_partial_data to the data block api

kevinbird15:
This would allow somebody to easily take a smaller sample of the data.  By default it would use 100% of the data, but that could be lowered by setting sample_pct to a lower number.  For example, by using sample_pct=0.01, it would select a random 1% of the samples.

Here is the thread for discussion and I am open to suggestions if there is a better way to structure this.  

https://forums.fast.ai/t/add-a-use-partial-data-to-the-data-block-api/31718

I would also appreciate help building a test to verify this works.


sgugger:
Thanks!

PR #1269, opened 2018-12-01
Added unit tests for metrics.py

movefast:
Added some unit tests for metrics.py. Also added a tensor size check and its related unit test for exp_rmspe to prevent # of element mismatch since as long as the batch size matches it would still return a loss without explicitly failing the function call even if pred differs in size along other dims.


sgugger:
Thanks!

PR #1272, opened 2018-12-02
Remove weight-file argument from ImageCleaner

fpingham:


sgugger:
There are weird files in this PR on top of your changes to image_cleaner.

Issue #901, opened 2018-10-16
'async' is a reserved word in Python >= 3.7


cclauss:
PyTorch >= 0.4.0 has replaced __async__ with __non-blocking__.

[flake8](http://flake8.pycqa.org) testing of https://github.com/fastai/fastai on Python 3.7.0

$ __flake8 . --count --select=E901,E999,F821,F822,F823 --show-source --statistics__
```
./setup.py:98:15: F821 undefined name '__version__'
    version = __version__,
              ^
./old/fastai/models/inceptionresnetv2.py:316:20: F821 undefined name 'pretrained_settings'
        settings = pretrained_settings['inceptionresnetv2'][pretrained]
                   ^
./old/fastai/models/inceptionresnetv2.py:321:17: F821 undefined name 'InceptionResNetV2'
        model = InceptionResNetV2(num_classes=1001)
                ^
./old/fastai/models/inceptionresnetv2.py:337:17: F821 undefined name 'InceptionResNetV2'
        model = InceptionResNetV2(num_classes=num_classes)
                ^
./old/fastai/models/cifar10/utils.py:114:32: F821 undefined name 'random'
  return string + '-{}'.format(random.randint(1, 10000))
                               ^
./old/fastai/models/cifar10/main_dxy.py:179:32: E999 SyntaxError: invalid syntax
      target = target.cuda(async=True)
                               ^
./old/fastai/models/cifar10/utils_kuangliu.py:17:18: F821 undefined name 'torch'
    dataloader = torch.utils.data.DataLoader(dataset, batch_size=1, shuffle=True, num_workers=2)
                 ^
./old/fastai/models/cifar10/utils_kuangliu.py:18:12: F821 undefined name 'torch'
    mean = torch.zeros(3)
           ^
./old/fastai/models/cifar10/utils_kuangliu.py:19:11: F821 undefined name 'torch'
    std = torch.zeros(3)
          ^
./old/tutorials/kmeans.py:58:30: F821 undefined name 'v_data'
        n_samples = tf.shape(v_data)[0]
                             ^
./courses/dl2/cgan/models/pix2pix_model.py:59:57: E999 SyntaxError: invalid syntax
            input_A = input_A.cuda(self.gpu_ids[0], async=True)
                                                        ^
./courses/dl2/cgan/models/cycle_gan_model.py:75:57: E999 SyntaxError: invalid syntax
            input_A = input_A.cuda(self.gpu_ids[0], async=True)
                                                        ^
[...]

PR #1279, opened 2018-12-03
'async' is a reserved word in Python >= 3.7

cclauss:
#901 was closed without fixing the syntax errors that were reported.  This PR proposes fixes for those syntax errors.


sgugger:
The current version of the course is built on top of fastai v0.7 which isn't compatible with python 3.7 (and we have no plan to make it so because...)
fastai v1 and the new version of the course (to come in January) will fully be. That's why the other PR was closed and I'm closing this one too.

PR #1280, opened 2018-12-04
fix saving of test set labels

joel-odlund:
test set labels are  saved properly.  Currently, no labels are saved causing silent errors when loading.


sgugger:
I'm unsure which silent error you are referring to. The test set is supposed to be a set of unlabeled data, and the load function doesn't look for test labels, or the `from_ids` function that is ultimately called doesn't take a test_lbl argument.

sgugger:
It looks like you want to validate on a different dataset. You should do `data_test=...` command that creates a DataBunch where your validation set is this labelled test. When you're done training, you can type `learn.data = data_test` then `learn.validate()`.

PR #1284, opened 2018-12-04
Make load_pretreined support models with optimizer saved

PiotrCzapla:
This is a workaround for ppl not to get into troubles if they saved their LM  along with the optimizer. Ideally, we should warn user with UserWarning but I haven't seen it in the code base so I haven't added it.

fixes: https://github.com/n-waves/ulmfit-multilingual/issues/20


sgugger:
Sounds good.

PR #1295, opened 2018-12-05
Declare conditional dataclasses dependency in a compatible way

Lothiraldan:
Latest release (1.0.33) moved dataclasses from a mandatory dependency to a
conditional dependency based on python version.

This condition was made with a Python condition in the setup.py. This works
fine when installing the source distribution but don't works with the wheel
distribution made with Python 3.7 as the condition isn't met at the time the
wheel is generated.

setuptool allows to define a conditional dependency based on python version
using the `python_version` environment marker. This way, the dependency is
explicitly marked as optional and always present on the wheel distribution.

See https://hynek.me/articles/conditional-python-dependencies/ for a good
summary of the possibilities.

I didn't updated the setup.py with the older way of declaring conditional
dependencies (with extra_requires) as the minimum version of setuptools that
support the `python_version` environment marker is more than one year old and
people having wheel support will likely have an up-to-date version of
setuptools.

I didn't removed the Python condition from the `setup.py` file either because
I don't know if fastai supports old environments with an old setuptools
version.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -51,6 +51,7 @@ requirements = to_list("""
     torchvision-nightly
     typing
     pyyaml
+    dataclasses ; python_version<'3.7'
 """)
 
 if sys.version_info < (3,7): requirements.append('dataclasses')


stas00:
Thank you, @Lothiraldan. I will complete the cleanup shortly and make a new release.

PR #1301, opened 2018-12-06
Fixes #1278

### LINKED ISSUE #1278: Multiple errors with SaveModelCallback when applied to tabular_learner
Code ...
```
learn = tabular_learner(data, layers=[1000,500], ps=[0.001,0.01], emb_drop=0.04, 
                        y_range=y_range, metrics=exp_rmspe)
best_model_cb = partial(callbacks.SaveModelCallback, monitor='exp_rmspe', mode='min')
learn.callback_fns.append(best_model_cb)

learn.lr_find() # or learn.fit_one_cycle(...)
```
If "bestmodel.pth" does not exist, it returns an error indicating it can't open (which will be the case the first time you run the LR finder or fit the model)

If "bestmodel.pth" does exist, it returns this error:
```
~/anaconda3/envs/fastai-course-v3/lib/python3.7/site-packages/fastai/callbacks/tracker.py in get_monitor_value(self)
     45         "Pick the monitored value."
     46         values = {'trn_loss':self.learn.recorder.losses[-1:][0].cpu().numpy(),
---> 47                   'val_loss':self.learn.recorder.val_losses[-1:][0]}
     48         for i, name in enumerate(self.learn.recorder.names[3:]):
     49             values[name]=self.learn.recorder.metrics[-1:][0][i]

IndexError: list index out of range
```

ohmeow:
Multiple errors with SaveModelCallback when applied to tabular_learner #1278

Fix does two things:
1.  Ensures there are validation losses to look at (parent class TrackerCallback requires)
2. Makes sure a <best_model>.pth exists before trying to load it


sgugger:
Thanks!

jph00:
@ohmeow @sgugger please always include tests with your bug-fix PRs - a test that would fail without your fix, and a test that should pass with the previous approach (if we don't already have a test for it). In this case, this PR refers to `monitor` instead of `self.monitor` so it causes the callback to fail in all cases. I've fixed it in master now - this is just a reminder that fixing broken code often takes longer than writing the code in the first place, since we have to get up to speed with the code and then fix it. So it's worth being careful that your PR really works the way you think. :)

Issue #1304, opened 2018-12-07
IndexError: "too many indices for array" for Multi-Label-List


gshashank84:
When I was executing learn.fit_one_cycle(1,1e-1, metrics = [acc_02,f_score]) it shows IndexError, by the way the one row of text contains more than one labels.
Please refer Images below for better Understanding.
![screenshot from 2018-12-07 22-18-38](https://user-images.githubusercontent.com/16812816/49661571-681f5d80-fa6f-11e8-8f21-dd423f4e53f0.png)
![screenshot from 2018-12-07 22-18-45](https://user-images.githubusercontent.com/16812816/49661584-71102f00-fa6f-11e8-9178-96fa6b1cd206.png)

Here is a look of DataClasBunch that I have created
![screenshot from 2018-12-07 22-18-57](https://user-images.githubusercontent.com/16812816/49661627-8dac6700-fa6f-11e8-8c34-2100fb337b21.png)

gshashank84:
After looking out carefully, I observed that the MultiCategoryList class have not passed argument of init function, as in this case **items must be 'train.csv' where as sep = '|'**, rather it is set as default to None. Please tell me how to pass the argument sepearation in this class through **learn** object?
```
class MultiCategoryList(CategoryListBase):
    "Basic `ItemList` for multi-classification labels."
    _processor=MultiCategoryProcessor
    def __init__(self, items:Iterator, classes:Collection=None, sep:str=None, **kwargs):
        if sep is not None: items = array(csv.reader(items.astype(str), delimiter=sep))
        super().__init__(items, classes=classes, **kwargs)
        self.loss_func = BCEWithLogitsFlat()

    def get(self, i):
        o = self.items[i]
        if o is None: return None
        return MultiCategory(one_hot(o, self.c), [self.classes[p] for p in o], o)
```

gshashank84:
Or I do think that at the one_hot method creates error due to multiple integers in the 'o'.
Here res[x] =1 raises IndexError because the x contains multiple integers. So I think there must be change in the code for **get** method in  **MultiCategoryList** Class for Multi-Labels.
```
def one_hot(x:Collection[int], c:int):
    "One-hot encode `x` with `c` classes."
    res = np.zeros((c,), np.float32)
    res[x] = 1.
    return res
```

gshashank84:
[...]

PR #1306, opened 2018-12-07
Update one_hot Method for MultiCategoryList

gshashank84:
If x has more than one integers then it will throw IndexError which is generally happening in this case where the Text of one row of a DataFrame contain more than one label as Target..
For more reference: [visit this Issue](https://github.com/fastai/fastai/issues/1304)


gshashank84:
But the one hot encoding part is being done in Numpy, not Pytorch I guess

sgugger:
Which also supports fancy indexing, the problem you are facing probably comes from your data as I said in the issue.

PR #1309, opened 2018-12-08
Update model.py

yang-zhang:
Old code seems to address "torch.nn.utils.clip_grad_norm is now deprecated in favor of torch.nn.utils.clip_grad_norm_." (https://pytorch.org/docs/0.4.1/nn.html#torch.nn.utils.clip_grad_norm_) but runs the same code regardless of the value of `IS_TORCH_04`.


sgugger:
Good catch!

PR #1310, opened 2018-12-08
Fixes path for saving best model

ohmeow:
Didn't catch this in initial PR because I actually had a variable named "path" set in my notebook.  This should fix up the SaveModel callback.


ohmeow:
@jph00 :  if possible, can you merge this into v 1.0.35 release as a hotfix?  if not, i'm sure we're fine waiting for 1.0.36.

sgugger:
Right. I'm afraid it's going to be in the next release (PS: this is why we need a test when adding a PR that fixes a bug ;) )

ohmeow:
Will do. Being lazy plus overconfident has been my downfall before ... eventually I'll learn.

Thanks.

Issue #1311, opened 2018-12-08
Unable to install with `python setup.py install`, possible `regex` version conflict


tallamjr:
Unable to install via `python setup.py install`

**Describe the bug**

error: regex 2018.11.22 is installed but regex==2018.01.10 is required
by {'fastai', 'spacy'}

**To Reproduce**

`pip install torch torchvision`
`pip install fastai`
`python setup.py install`

**Expected behavior**

This should run to completion without error

**Screenshots**

**Additional context**

stas00:
Applied your fix, https://github.com/fastai/fastai/pull/1312, thank you.

PR #1312, opened 2018-12-08
Fixes issue #1311 of not being able to to run installation with pip

tallamjr:
Removal of duplicate 'regex' package in requirements list, and ensuring
version is set to 2018.01.10, otherwise this is then incompatible with
'spacy' package.

error: regex 2018.11.22 is installed but regex==2018.01.10 is required
by {'fastai', 'spacy'}

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -38,12 +38,11 @@ requirements = to_list("""
     pandas
     Pillow
     pyyaml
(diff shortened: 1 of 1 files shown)


stas00:
Thank you, @tallamjr

PR #1314, opened 2018-12-09
Tidy up dependencies

honnibal:
Some of spaCy's dependencies were specified explicitly, when I think the fastai doesn't depend on them directly. This might cause problems for upgrading, especially where you had specific versions pinned.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -29,7 +29,6 @@ def to_list(buffer): return list(filter(None, map(str.strip, buffer.splitlines()
 
 requirements = to_list("""
     bottleneck
-    cymem==2.0.2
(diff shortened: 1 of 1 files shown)


stas00:
That's correct, @honnibal.

This is on purpose, since when there was the trouble with spacy's dependencies some weeks back, we found this pinned set to work everywhere and since we recommend to have a fastai-dedicated conda environment, in theory these should cause no conflicts. 

We are waiting for spacy 2.0.18 to become available on conda's main channel (currently it's only on conda-forge) and once this happens we will switch to depending on just `spacy==2.0.18` and remove its dependencies.

PR #1316, opened 2018-12-09
Fix. Should be mnist_stats instead of imagenet_stats, because it is a mnist dataset. Accuracy improoves to 0.997547 and 0.996535 accordingly.

mgrankin:
Small bug fix. Should by mnist_stats instead of imagenet_stats, because it is a mnist dataset. Accuracy improoves to 0.997547 and 0.996535 accordingly.

### DIFF
--- a/examples/vision.ipynb
+++ b/examples/vision.ipynb
@@ -82,7 +82,7 @@
    ],
    "source": [
     "data = ImageDataBunch.from_folder(path, ds_tfms=(rand_pad(2, 28), []), bs=64)\n",
-    "data.normalize(imagenet_stats)\n",
+    "data.normalize(mnist_stats)\n",
     "img,label = data.train_ds[0]\n",
     "img"
    ]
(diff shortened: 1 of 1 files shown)


sgugger:
No, it should be imagenet stats because we use a model pretrained on imagenet. Note that the small improvement in accuracy may very well come from the randomness of training.

mgrankin:
You think we should recreate Imagenet mean and std and feed that data to model. I believe this is not true. An Imagenet model was trained and expects data with mean = 0 and std = 1. In order to get those you should calc mean and std from your own dataset and normalise it. imagenet_stats usefull for Imagenet subsample, so you don't need to recalc mean and std for it. 

Look to the code

```
def _normalize_batch(b:Tuple[Tensor,Tensor], mean:FloatTensor, std:FloatTensor, do_x:bool=True, do_y:bool=False)->Tuple[Tensor,Tensor]:
 ….
 x = normalize(x,mean,std)

def normalize(x:TensorImage, mean:FloatTensor,std:FloatTensor)->TensorImage:
"Normalize `x` with `mean` and `std`."
return (x-mean[...,None,None]) / std[...,None,None]
```

This means we don’t recreate the original mean and std with out code. This is the formula to make it 0 and 1.

mgrankin:
I think I finally got it. Still feels weird.

PR #1319, opened 2018-12-11
Recursion error inside data_block.py

cudawarped:
I am not sure if this is a window's only issue, but with the latest changes the call to `hasattr` inside `__getattr__` causes a recursion error.

This occurs in the dogs_cats example as soon as you call 
`learn.fit_one_cycle(1)`
and gives me the following output in the anaconda prompt

> Traceback (most recent call last):
>   File "D:\Anaconda3\envs\test_fastai\lib\multiprocessing\spawn.py", line 115, in _main
>     self = reduction.pickle.load(from_parent)
>   File "d:\dev\repos\fastai_fork\fastai\data_block.py", line 498, in __getattr__
>     if hasattr(self,'x'):
>   File "d:\dev\repos\fastai_fork\fastai\data_block.py", line 498, in __getattr__
>     if hasattr(self,'x'):
>   File "d:\dev\repos\fastai_fork\fastai\data_block.py", line 498, in __getattr__
>     if hasattr(self,'x'):
>   [Previous line repeated 995 more times]
> RecursionError: maximum recursion depth exceeded while calling a Python object

With the following changes from [ stackoverflow](https://stackoverflow.com/questions/25471268/maximum-recursion-depth-error-in-the-python-function), `learn.fit_one_cycle(1)` no longer causes a recursion error.


cudawarped:
This error only occurs when `num_workers!=0` which does not currently work on windows anyway.  I am not sure if this change will still need to be made when `num_workers!=0` works on windows so I have left it open.

jph00:
Many thanks - fixed in master using an alternative approach.

PR #1322, opened 2018-12-12
Fix process_one() for CategoryProcessor

ronlut:
Exception that should be thrown when we try to get the index of class we didn't see before was silenced causing `None` to be returned (`dict.get()` returns None if not found)


sgugger:
Thanks!

PR #1324, opened 2018-12-12
Add ability to unpickle `learner` object without recursion stack overflow

falschparker82:
This Pull request fixes unpickling previously pickled FastAI `learner` and `data_bunch` objects on quite a lot of different Python versions.

Currently, unpickling a `learner` object will lead to recursion errors:

```
Exception ignored in: <object repr() failed>
  File "/home/ubuntu/jw-pymatcher/jw-pymatcher/fastai/basic_train.py", line 196, in __del__
    def __del__(self): del(self.model, self.data)
AttributeError: model
  File "match_predict.py", line 99, in <module>
    predictor = pickle.load(f)
  File "/home/ubuntu/jw-pymatcher/jw-pymatcher/fastai/callback.py", line 54, in __getattr__
    def __getattr__(self,k:str)->Any: return getattr(self.opt, k, None)
  File "/home/ubuntu/jw-pymatcher/jw-pymatcher/fastai/callback.py", line 54, in __getattr__
    def __getattr__(self,k:str)->Any: return getattr(self.opt, k, None)
  File "/home/ubuntu/jw-pymatcher/jw-pymatcher/fastai/callback.py", line 54, in __getattr__
    def __getattr__(self,k:str)->Any: return getattr(self.opt, k, None)
  [Previous line repeated 996 more times]
RecursionError: maximum recursion depth exceeded
```

For the bug and the fix see:
https://bugs.python.org/issue5370

This will eventually be fixed in Python itself, but the fix seems to have been released only today in RC and will take at least a month to propagate to stable: https://bugs.python.org/issue35444

This patch should not harm functionality in any way.


falschparker82:
Note: This may or may not be related to https://forums.fast.ai/t/recursion-error-fastai-v1-0-27-windows-10/30673 — I found this during my own bug hunt, but can't test due to not having Windows.

PR #1350, opened 2018-12-17
Set tools/run-after-git-clone to use python3 by default

odysseus0:
I just ran through the steps listed in the documentation contribution guide and encountered an error when running `tools/run-after-git-clone`. I am using OS X.

Here is the error message:

```python
  File "tools/run-after-git-clone", line 17
    cmd = f"{sys.executable} {script}"
                                     ^
SyntaxError: invalid syntax
```

By correcting the shebang of the file, the script will now run correctly regardless of the relative order of the paths of Python2 and Python3 in the user's $PATH variable.

### DIFF
--- a/tools/run-after-git-clone
+++ b/tools/run-after-git-clone
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 # if you're a fastai developer please make sure you do this:
 #


sgugger:
@stas00 I think that was the script changed because of Windows, correct me if I'm wrong. Is there some way to make it both Linux and Windows compatible?

stas00:
> @stas00 I think that was the script changed because of Windows, correct me if I'm wrong. Is there some way to make it both Linux and Windows compatible?

No, I have already changed that to python3 and then rolled back to python, because you said windows doesn't have python3.exe and you couldn't figure out how to symlink (or if it's even possible - no idea).

Absolutely no to 2 separate scripts with identical content.

And yes, changing the instruction to `python3 tools/run-after-git-clone` is one way.

Another way is require python3 at the top of the script, since fastai requires python 3.6+

```
import sys
if sys.version_info < (3,6): sys.exit("Python %s.%s or later is required.\n" % MIN_PYTHON)
```

odysseus0:
I believe that [PEP 394 -- The "python" Command on Unix-Like Systems](https://www.python.org/dev/peps/pep-0394/) will shed some light on this issue. Here are some relevant information I found.

> - for the time being, all distributions should ensure that python, if installed, refers to the same target as python2, unless the user deliberately overrides this or a virtual environment is active.
>
> - however, end users should be aware that python refers to python3 on at least Arch Linux (that change is what prompted the creation of this PEP), so python should be used in the shebang line only for scripts that are source compatible with both Python 2 and 3.
>
> - In order to tolerate differences across platforms, all new code that needs to invoke the Python interpreter should not specify python, but rather should specify either python2 or python3 (or the more specific python2.x and python3.x versions; see the Migration Notes). This distinction should be made in shebangs, when invoking from a shell script, when invoking via the system() call, or when invoking in any other context.

stas00:
Thank you for the link, @odysseus0, and re-initiating this change (which we have tried before).

Since your proposed change has been applied to all tools/* already I'm closing this PR. but if someone has issue with the change please don't hesitate to comment, including posting the error messages and information about your setup if you do.

PR #1354, opened 2018-12-17
simple import correction

Benudek:
had failures when running pytest with imports, corrected as below.

Hope that's helpful


stas00:
Good catch, thank you, @Benudek!

PR #1358, opened 2018-12-18
Fixed the callbacks iteration to go over the correct collection

aslisabanci:
Updated the callbacks iteration in `on_batch_begin` to go over the correct collection of `cbs` and not `self.callbacks`


sgugger:
Oh good catch!

PR #1359, opened 2018-12-18
test method show batch

Benudek:

### DIFF
--- a/tests/test_text_data.py
+++ b/tests/test_text_data.py
@@ -125,3 +125,17 @@ def test_from_ids_works_for_variable_length_sentences():
                                       train_ids=ids, train_lbls=lbl,
                                       valid_ids=ids, valid_lbls=lbl, classes={0:0}, bs=8)
     text_classifier_learner(data).fit(1)
+
+def test_show_batch(capsys):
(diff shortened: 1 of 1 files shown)


Benudek:
**Description: Simple test with assert for show batch method.** 

Check manually e.g. with: pytest -k 'show_batch' tests/test_text_data.py

Potential Follow Ups:
- Test method could be extended to try to reproduce error, e.g. issue here: https://forums.fast.ai/t/getting-cuda-oom-on-lesson3-imdb-notebook-with-a-bs-8/30080/14

- Object _captured.out_ delivers on IPython notebook _IPython.core.display.HTML_ object. If we had a way to get to the HTML body (see some failed attempts below), we could assert over string 'label' e.g.
-- sys.stdout.write(captured.out)
-- print(captured.out)
-- print(HTML(captured.out).data)
-- display(HTML(captured.out))
-- print(str(captured.out))
-- print(str(HTML(captured.out)))
-- print(str(HTML(captured.out).data))

sgugger:
With the new fake data object, show_batch prints something that can be intercepted by stdout if we want a basic of if show_batch works.

Benudek:
ok, thx. will close this PR here then and see if I can come up with a better test.

PR #1363, opened 2018-12-19
Fix link to Pillow-SIMD section

jacquerie:
A small typo that I noticed while reading https://docs.fast.ai/performance.html#libjpeg-turbo.


stas00:
Thank you, @jacquerie

PR #1364, opened 2018-12-19
Minor docs fix, added comment - torch_core.defaults.device

tpietruszka:
There was a typo in torch_core documentation - `default_device` instead of `defaults.device`, also an incorrect line was linked under the "source" link. 

Added an explicit comment about how to make fastai run on CPU, as I have lost some time figuring it out myself.


sgugger:
Yes, you made the PR before I fixed that test. Thanks!

PR #1366, opened 2018-12-20
vision.image docs transforms not showing for mask, points and bbox

KeremTurgutlu:
I was going through documents to try out different transforms on a particular bounding box data. In vision.image docs all the example images seemed untransformed (including point and mask). I thought maybe those are little changes that human eye can't distinguish but when I checked in my notebook with `p_affine=1` it was still the same with same code chunk. Then I figured that transformed `img` and correspond `y` are not assigned to variables that been used to show in docs. Little change, though it may help people to not get confused like me :D

### DIFF
--- a/docs_src/vision.image.ipynb
+++ b/docs_src/vision.image.ipynb
@@ -1544,8 +1544,8 @@
     "_, axs = plt.subplots(2,4,figsize=(12,6))\n",
     "for ax in axs.flatten():\n",
     "    img,mask = get_seg_ex()\n",
-    "    img.apply_tfms(tfms[0], size=224)\n",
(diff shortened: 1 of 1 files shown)


sgugger:
Looks like someone forgot to adjust those when they changed the behavior of apply_tfms (used to be inplace)!
Thanks for spotting and fixing this, just be sure to run the command tools/run-after-git-clone so that your notebooks are automatically stripped (for this one, have a look [here](https://docs.fast.ai/dev/develop.html#unstripped-notebook-repair) to force the stripping). That's why tests are failing.

KeremTurgutlu:
Got it, thanks a lot !

PR #1367, opened 2018-12-20
Add some description to the callbacks overview page

pouannes:
[As discussed here.
](https://forums.fast.ai/t/documentation-improvements/32550/34?u=pierreo)
The image are represented as some a humongous block of characters, I hope that's normal.

Also all the first commits are of another PR that's already merged, I suppose I should have done something so they are not shown here but I don't really know what.


sgugger:
Thanks!

PR #1371, opened 2018-12-21
Fixing notebook lesson2-rf_interpretation from ML course

juanpedrovel:
The latest version of PDPbox has an issue with pdp.pdp_interact_plot as it only returns 1 plot instead of 3.

The changes made are for the environments for GPU and CPU to include the proper versions of PDPBox and matplotlib to work.

Also, function pdp.pdp_isolate() calls were modified on notebook to take 3 arguments and be compatible with previous versions


jph00:
Many thanks.

Issue #1335, opened 2018-12-13
Cannot train LM in fp16 mode


rpowalski:
**Describe the bug**
Cannot train language model in half precision mode. 
That's what I got during training:
```
/opt/conda/lib/python3.7/site-packages/fastai/text/models.py in forward(self, words, scale)
     69         if scale: masked_embed.mul_(scale)
     70         return F.embedding(words, masked_embed, self.pad_idx, self.emb.max_norm,
---> 71                            self.emb.norm_type, self.emb.scale_grad_by_freq, self.emb.sparse)
     72 
     73 #def _repackage_var(h:Tensors)->Tensors:

/opt/conda/lib/python3.7/site-packages/torch/nn/functional.py in embedding(input, weight, padding_idx, max_norm, norm_type, scale_grad_by_freq, sparse)
   1452         # remove once script supports set_grad_enabled
   1453         _no_grad_embedding_renorm_(weight, input, max_norm, norm_type)
-> 1454     return torch.embedding(weight, input, padding_idx, scale_grad_by_freq, sparse)
   1455 
   1456 

RuntimeError: Expected tensor for argument #1 'indices' to have scalar type Long; but got torch.cuda.HalfTensor instead (while checking arguments for embedding)
```

**To Reproduce**
I tried to run examples/text.ipynb with mixed precision
To do that I modified the following line:
`learn = language_model_learner(data_lm, pretrained_model=URLs.WT103_1).to_fp16()`

**Additional context**
`cifar.ipynb` example works fine with fp16.
```text
=== Software === 
python        : 3.7.1
fastai        : 1.0.37
fastprogress  : 0.1.18
torch         : 1.0.0
nvidia driver : 384.111
torch cuda    : 9.0.176 / is available
torch cudnn   : 7401 / is enabled

=== Hardware === 
nvidia gpus   : 8
torch devices : 8
  - gpu0      : 16152MB | Tesla V100-SXM2-16GB
  - gpu1      : 16152MB | Tesla V100-SXM2-16GB

=== Environment === 
platform      : Linux-4.4.0-112-generic-x86_64-with-debian-stretch-sid
distro        : #135-Ubuntu SMP Fri Jan 19 11:48:36 UTC 2018
conda env     : Unknown
python        : /opt/conda/bin/python
sys.path      : /home/usr

/opt/pytorch/pytorch/torch/lib/python3.6/site-packages
/opt/conda/lib/python3.6/site-packages
/usr/local/onnx
[...]

PR #1374, opened 2018-12-21
Fix backpropagation of LSTM when trained on fp16

PiotrCzapla:
Otherwise, you get an error that the Backpropagation mixes fp16 with fp32. loss is fp32 while mean of weights is fp16. The proposed solution simply converts AR & TAR to fp32 always. This should be a noop in normal training and it fixes the issue during fp16 training.

I don't have a card that supports fp16 temporarily available so I can't test the changes.

fixes: #1335


sgugger:
Probably yes, thanks!

PR #1375, opened 2018-12-21
Break cyclic references to Learner from callbacks

PiotrCzapla:
This way when Learner goes out of scope it is imidatelly freed along with the model by reference counting algorithm.

As per [discussion on the forum](https://forums.fast.ai/t/ipyexperiments-getting-the-most-out-of-your-gpu-ram-in-jupyter-notebook/30145/25) I've added weak references to callbacks so that Learner can be automatically freed when it goes out of scope. Without the week references, Callbacks are making a cyclic reference to Learner, then it can be freed only when garbage collection is executed, which runs each time `n` objects were created which isn't often enough.

In addition, this makes repr(learn) shorter.

### DIFF
--- a/fastai/basic_train.py
+++ b/fastai/basic_train.py
@@ -1,4 +1,7 @@
 "Provides basic training and validation with `Learner`"
+import weakref
+from dataclasses import InitVar
+
 from .torch_core import *
 from .basic_data import *
 from .callback import *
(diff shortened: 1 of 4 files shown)


stas00:
the test is failing with `assert 101 < 100`:

should this test line have been:
```
assert gc_collected < 102
```
but then you comment says there should be 100 of those - we get 101.

PR #1376, opened 2018-12-21
vision.image docs minor change - tfms apply no longer inplace

KeremTurgutlu:
Due to change to tfms apply not being inplace anymore docs don't show the transformed images for points, bbox and segmentations.


sgugger:
Perfect like this, thanks!

PR #1382, opened 2018-12-23
Image downloader

xnutsive:
Proposing the `ImageDownloader` widget. Similarly to `ImageCleaner`, it's an ipywidget that works in Notebook or Lab, allows users to quickly research and download images from google search. 

[Here's the thread on fastai dev forums](https://forums.fast.ai/t/imagedownloader-widget/32776). 

**Why:** 

Building image classifiers in the first weeks of the MOOC is twice as fun with your own dataset, right? I noticed fastai v1 ships with a JS snippet to download images and that's mentioned in the new version of the MOOC's notebooks. 

`ImageDownloader` would allow people experiment with new datasets easily, we'll have more people implementing their own classifiers and apps based on them → more fun, and more inspiration for them and other students to learn and implement their ideas. 

**What works:**

`ImageDownloader()` works in Jupyter Notebook and Lab: 

- [x] User types in the search term / query, number of images, and image resolution they want. 
- [x] Downloads images from google, stores them in a directory. The directory is created automatically if it doesn't exist. 
- [x] Previews a few images from the download. 
- [x] Allows downloading >100 images per label, but requires `selenium` and `chromedriver` to do that. Shows an error message explaining how to install those if the user requests more than a hundred images. 
- [x] Uses fastai's `download_url()` and `parallel()` to speed things up ;-) 
- [x] Integrates with `fastprogress` and shows progress bar when downloading. 
- [x] Calls `verify_images()` after the download to clean up the dataset. 
- [x] Directory structure compatible with datablock API, i.e. `from_folder()` “just works”.
- [x] Also provides a function to download images in iPython / python scripts: `download_google_images()`, although it's in `fastai.widgets.image_downloader`. It maybe useful to move it to `fastai.visual.data`, but I wanted to start small and show @sgugger the diff first. 

**Also taken care of:** 

- [x] Docs: I created a section in `docs_src/widgets.ipynb`, documented both the widget, and the function, with examples, using `show_doc()`, and regenerated the html docs with `tools/build-docs`
- [x] Added a test to make sure the widget init runs without errors. 

**Ideas to improve this:**

- [ ] I'm working on a full example: fetch a dataset, train, save the model, and serve it as a REST API and a website, just to give other students some inspiration. Probably not for `fastai/examples`, I can post that outside of fastai if that'd be more appropriate.
- [ ] It'd be good to add some tests to test the widget internal logic, it's decoupled into several helper functions, but they're not unit tested.
- [ ] Since it's questionable whereas it's legal to scrape google search like that, maybe this can't be merged at all, or maybe it's a good idea to add a legal disclaimer to the docs? 

@fpingham, @sgugger, I'm not really a Python developer, and I'd be happy to improve the code, fix any smells, code styles, anything that comes up. I'd be happy to help in other projects as well, just wanted to start with something simple.

Thank you both very much for your work on fastai, for your comments on the forum thread, and for reviewing the PR. 

What do you think, @sgugger?

### DIFF
--- a/fastai/widgets/image_downloader.py
+++ b/fastai/widgets/image_downloader.py
@@ -1,0 +1,106 @@
+from ..core import *
+from ipywidgets import widgets, Layout, Output, HBox, VBox, Text, BoundedIntText, Button, Dropdown, Box
+from IPython.display import clear_output, display
+from google_images_download import google_images_download
+
+__all__ = ['ImageDownloader']
+
+class ImageDownloader():
+    """
+    Displays a simple widget that allows searching and downloading images from google images search
+    in a Jupyter notebook.
+    """
+
+    def __init__(self, path:Union[Path,str]='data'):
+        "Setup path to save images to, init the UI, and render the widgets."
+        self._path = Path(path)
+        self._ui = self._init_ui()
+        self.render()
+
+    def _init_ui(self) -> VBox:
+        """
+        Initialize the widget UI and return the UI, 
+        only used to initially create all the components and setup event handlers."
+        """
+        self._search_input = Text(placeholder="What images to search for?")
+        self._count_input = BoundedIntText(placeholder="How many pics?", 
+                                        value=10, min=1, max=100, step=1, 
+                                        layout=Layout(width='60px'))
+        self._size_input = Dropdown(options= ['>400*300', '>640*480', '>800*600', '>1024*768', '>2MP', '>4MP', '>6MP', '>8MP', '>10MP'],
+                                    value='>400*300',
+                                    layout=Layout(width='120px'))
+        self._download_button = Button(description="Search & Download", icon="download", layout=Layout(width='200px'))
+        self._download_button.on_click(self.on_download_button_click) 
+
+        self._output = Output()
+
+        # Top horizontal controls bar
+        self._controls_pane  = HBox([self._search_input, 
+                                    self._count_input,
+                                    self._size_input,
+                                    self._download_button],
+                                    layout=Layout(width='auto', height='40px'))
+
+        self._heading = ""
+        self._download_complete_heading = "<h3>Download complete. Here are a few images</h3>"
+        self._preview_header = widgets.HTML(self._heading, layout=Layout(height='60px'))
+        self._img_pane = Box(layout=Layout(display='inline'))
+        return VBox([self._controls_pane, self._preview_header, self._img_pane])
+
+
+    def render(self) -> None:
+        "Render the image search widget."
+        clear_output()
+        display(self._ui)
+
+    def clear_imgs(self) -> None:
+        "Clear the widget's images preview pane."
+        self._preview_header.value = self._heading
+        self._img_pane.children = tuple()
+
+    def validate_search_input(self) -> bool:
+        "Check if input value is empty."
+        input = self._search_input
+        if input.value == str():
+            input.layout = Layout(border="solid 2px red", height='auto')
+        else:
+            self._search_input.layout = Layout()
+        return input.value != str()
+
+    def on_download_button_click(self, btn) -> None:
+        "Download button click handler: validate search term and download images."
+        term = self._search_input.value           
+        limit = int(self._count_input.value)
+        size = self._size_input.value
+
+        if not self.validate_search_input(): return
+        
+        self.clear_imgs() 
+        self.download_images(term, limit=limit, size=size)
+        self.display_images_widgets(self.get_preview_images_fnames(term)[:min(limit, 12)])
+        self._preview_header.value = self._download_complete_heading
+        self.render()
+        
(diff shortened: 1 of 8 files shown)


xnutsive:
🧐 looks like I broke the build somewhere, but I'm not sure how to get to the build log, and at least the unit tests pass. Probably broke something in the docs (?).

I'll take a look tomorrow.

stas00:
You can see the failing tests in the CI logs:
https://dev.azure.com/fastdotai/fastai/_build/results?buildId=2535
You didn't strip the notebooks. See:
https://docs.fast.ai/dev/develop.html#things-to-run-after-git-clone
https://docs.fast.ai/dev/develop.html#unstripped-notebook-repair

xnutsive:
@stas00, thank you for pointing me to the build pipelines and nbstripout problem. 

@sgugger, I think it's ready for review. I'll do a couple examples myself and if anything breaks I'll add more tests here. 

@fpingham, I merger `master` into this and updated `widgets.image_cleaner.ipynb`.

docs_src/sidebar/sidebar_data.py
```diff
@@ -47,7 +47,7 @@
                 'vision.gan': '/vision.gan',
                 'vision.model overview': '/vision.models',
                 'vision.models.unet': '/vision.models.unet',
-                'vision.widgets': '/widgets.image_cleaner'
+                'widgets.image_cleaner': '/widgets.image_cleaner'
```
xnutsive: @fpingham, I changed the namespace for widgets in the sidebar to reflect where they actually are, seems like a mismatch in master.

fastai/widgets/image_downloader.py
```diff
@@ -0,0 +1,206 @@
...
+        self.render()
+
+    def display_images_widgets(self, fnames:list) -> None:
+        "Display a few preview images in the notebook"
+        imgs = [widgets.Image(value=open(f, 'rb').read(), width='200px') for f in fnames]
+        self._img_pane.children = tuple(imgs)
+
+
+def download_google_images(path:PathOrStr, search_term:str, size:str='>400*300', n_images:int=10, format:str='jpg',
+                            max_workers:int=8, timeout:int=4) -> FilePathList:
```
sgugger: I'd make max_workers defaut to defaults.cpus (so a None in the default arg then that value passed in a test inside the function that uses it, which seems to be _download_images).
Also we prefer to have just one line for the docstring and the full documentation on the corresponding doc notebook.

fastai/widgets/image_downloader.py
```diff
@@ -0,0 +1,206 @@
...
+    def _init_ui(self) -> VBox:
+        """
+        Initialize the widget UI and return the UI,
+        only used to initially create all the components and setup event handlers."
+        """
+        self._search_input = Text(placeholder="What images to search for?")
+        self._count_input = BoundedIntText(placeholder="How many pics?",
+                                        value=10, min=1, max=5000, step=1,
+                                        layout=Layout(width='60px'))
+        self._size_input = Dropdown(options= ['>400*300', '>640*480', '>800*600', '>1024*768', '>2MP', '>4MP', '>6MP', '>8MP', '>10MP'],
```
sgugger: Maybe options need to go on several lines here as it's really wide.
Or maybe have the dict defined later be a private dict defined in this module and use {that_dict}.keys() here to get things a little bit shorter?

sgugger:
Seems good, I've just a added a few comments on style + defaults. Thanks a lot for this!

fastai/widgets/image_downloader.py
xnutsive: Stripped extra docstrings in public functions (that are documented in notebooks). Left some extra docs in private functions, since those might not be documented otherwise. 

I agree, using `defaults.cpus` is a good idea and exactly the purpose of having `defaults.cpus` in the first place, thanks for pointing that out!

fastai/widgets/image_downloader.py
xnutsive: Moved `_img_sizes` to a private variable in the module, using that in both `_search_url` and the UI options. Tidies things up indeed. 👍

xnutsive:
Thanks for the tips! Updated the code styles and used the defaults you pointed out. LGTM or should I do a full example notebook in `examples` to show the all the way from an idea to serving?

docs_src/sidebar/sidebar_data.py
fpingham: Yes, this is correct.

sgugger:
Just made a little bit of restyle to save vertical space which shouldn't change anything.

sgugger:
One last problem @xnutsive : the doc notebook widget.image_cleaner can't be executed automatically now, which breaks our expanded test suite (all the doc notebooks are run when we launch the tests before a release). I've fixed it by moving some cells in markdown code, but you may want to have a second look.

xnutsive:
Sure, will do.

PR #1384, opened 2018-12-23
show_doc: make the arguments/defaults stand out from types

stas00:
bold the argument name and bold+italicize the default value, to make them standout from very complex at times annotations, to improve function signature readability and quickly tell arguments from types (which otherwise is not easy when types are complex structures with commas, etc.)

before and after:

snap5


sgugger:
It looks nicer that way. Let's try on the docs and may revert if it's ugly in HTML, but I don't think so.

PR #1386, opened 2018-12-24
Throw exception when column we want to normalize isn't numerical.

erwald:
So I was creating a learner with a large (~2gb) data set, like so:

```python
procs = [Categorify, Normalize]
data = (TabularList.from_df(df,
                            path=data_folder,
                            cat_names=category_names,
                            cont_names=continuous_names,
                            procs=procs)
        .split_by_idx(valid_idx)
        .label_from_df(cols=dep_var, label_cls=FloatList)
        .databunch())
```

But execution would mysteriously stop someway in the middle, Python would hang and I would have no idea what went wrong. By peppering the fastai library with logs I figured out that I was sending a categorical column in as a continuous one, leading to a failure when normalizing it.

This commit, which just adds a check to make sure a column is numerical before normalizing it, will perhaps help future users of fastai to avoid having to spend half a day debugging it as I had to!

// Erich

### DIFF
--- a/fastai/tabular/transform.py
+++ b/fastai/tabular/transform.py
@@ -1,5 +1,6 @@
 "Cleaning and feature engineering functions for structured data"
 from ..torch_core import *
+from pandas.api.types import is_numeric_dtype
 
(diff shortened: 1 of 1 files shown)


sgugger:
Good idea, thanks!

PR #1392, opened 2018-12-26
Split df automatically into continuous and categorical column names.

RealLankinen:
When we are handling tabular data Fastai assumes us to give column names of continuous and categorical values. This new function make it easier to get the column names. Int and float type columns will be returned as continuous if the cardinality is more than max else all other columns are categorical.


RealLankinen:
Done!

sgugger:
Thanks!

PR #1394, opened 2018-12-26
Fix typo in docs: basic_train.ipynb

matanhershberg:
A Learner creates a Recorder object automatically - you do not need to explicitly pass **_it_** to callback_fns - because other callbacks rely on it being available. It stores the smoothed loss, hyperparameter values, and metrics **_for_** each batch, and provides plotting methods for each. Note that Learner automatically sets an attribute with the snake-cased name of each callback, so you can access this through Learner.recorder, as shown below.


sgugger:
Thanks!

PR #1403, opened 2018-12-28
Updated apply_cats

suleepkumar:
The previous method fails in below case where the column is of bool type but contains null values. 
df = pd.DataFrame({'col1' : [1, 2, 3], 'col2' : [np.NaN, False, True]})
df2 = pd.DataFrame({'col1' : [1, 2, 3], 'col2' : [np.NaN, False, True]})
So is_string_dtype will still be True for the column. And you will get error "buffer dtype mismatch, expected 'Python object' but got 'unsigned long' pd categorical"


sgugger:
Thanks!

PR #1410, opened 2018-12-29
Support for (ch, h, w) size in vision transforms

xnutsive:
I noticed `ImageDataBunch` exploded when providing `size=(3,244,244)`. The code seems to support it, but there were a couple places where some functions expected size to have just 2 elements if it's a tuple, and other expected 3. 

That resulted in trying to `listify((int,int,int),2` which throws a len mismatch error.

Fixed that: 
- `_get_crop_target()` is only used in `image.py`, so allowing it to receive tuple of len 3 is safe
- Added a helper function to safely cast size to (h,w)
- Added test for that helper
- Added tests for vision.transforms to make sure using that helper doesn't ruin the existing transforms that use size (crop). 

@stas00, @sgugger, please let me know if there's a better way to write and document such small PRs, or if you'd like any changes to my code styles, as well as more / other tests.

### DIFF
--- a/fastai/vision/image.py
+++ b/fastai/vision/image.py
@@ -5,7 +5,7 @@ from io import BytesIO
 import PIL
 
 __all__ = ['PIL', 'Image', 'ImageBBox', 'ImageSegment', 'ImagePoints', 'FlowField', 'RandTransform', 'TfmAffine', 'TfmCoord',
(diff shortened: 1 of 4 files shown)


sgugger:
Code style is great and fits perfectly with ours. Seems fine to me, we'll see if it breaks things by using it a little bit.

PR #1414, opened 2018-12-30
Fix for RuntimeError regarding Metrics based on CM

SvenBecker:
Fixed an issue submitted by Abu Fadl on the fastai forum regarding a RuntimeError message when utilizing gpu. 

Added a RegMetric class to store predictions and targets to allow for computations on epoch end instead of averaging across batches. (TODOs for regression metrics)


sgugger:
Thanks for taking care of the TODOs. Is there any chance you can put those changes on the metrics module that is on master? You're using the one from your last PR so the merge I did is ignored (and I'd rather not to do it again as I may mess something up).
If it's easier to close this PR and start a fresh one, please do so.

PR #1416, opened 2018-12-30
Fix for RuntimeError regarding Metrics based on CM + TODOs

SvenBecker:
Fixed an issue submitted by Abu Fadl on the fastai forum regarding a RuntimeError message when utilizing gpu.

Added a RegMetric class to store predictions and targets to allow for computations on epoch end instead of averaging across batches. (TODOs for regression metrics)


sgugger:
(close and reopen is just to re-trigger the test suite)

stas00:
@sgugger, FYI, you don't need to reopen the issue to rerun the PR build azure check. You can just go to the report on azure and request a rebuild, in this case:
https://dev.azure.com/fastdotai/fastai/_build/results?buildId=2620
clicking on [...] in the right upper corner and choosing Rebuild.
If you don't have the right permissions to do that please ask Jeremy to add you to azure's fastai account managers.

PR #1418, opened 2018-12-30
Update data.py

PSandeepSandy:
"It’s a casting problem. Sometimes pandas dataframe gets confused when adding columns if it’s a number in the data somewhere as the “word” or “token”. We are doing strictly text work in this python file and we always want string not integer operation for “adding”. Here’s the code fix (data.py):"

https://forums.fast.ai/t/textdatabunch-from-file-string/30730/11

credits - gob


stas00:

PR #1419, opened 2018-12-31
Make no_check argument available when creating ImageDataBunch

kiendang:
Currently passing `no_check=True` when creating `ImageDataBunch` using `from_name_re`, `from_name_func` or any other method that calls `create_from_ll` has no effect.


sgugger:
Good catch, thanks!

PR #1422, opened 2019-01-01
Fix _find_coeffs due to change in torch.gesv

kiendang:
Due to https://github.com/pytorch/pytorch/commit/5ac95758e2d9362f21a33c6846da37804184e198 `torch.gesv(B, A)` no longer works if `B` is a 1-D vector


sgugger:
Thanks!

PR #1438, opened 2019-01-08
Fixed an error while importing from 'imports'

paritoshM9:
An error occurred as we were importing from '.imports' rather than 'imports'


sgugger:
No, this is standard python module structure. When importing a module in the same directory, it's always 
.{module_name}

PR #1450, opened 2019-01-10
pred_batch should predict on values returned by on_batch_begin

bearpelican:
I think we should be using the output transformed by on_batch_begin during pred_batch.

This is also more consistent with logic in the validate/fit functions.


sgugger:
Indeed, thanks for catching this!

PR #1453, opened 2019-01-10
Test scripts train

Benudek:
tests for train & basic_train

### DIFF
--- a/tests/test_basic_train.py
+++ b/tests/test_basic_train.py
@@ -1,3 +1,42 @@
 import pytest,fastai
+from utils.fakes import *
+from utils.text import *
 
-#I'm all empty, please fill me with tests about the functions in basic_train.py
+## filename: test_basic_train.py
+## tests code in basic_train.py
+
+## run: pytest tests/test_basic_train.py -s
+
+## Class Learner
+
+## see documentation:  https://docs.fast.ai/basic_train.html
+ 
+## Model fitting methods
+ 
+def test_fit(capsys):
+    learn = fake_learner()
+    learning_rate = 0.01
+    weight_decay = 1e-2
+    learn.fit(epochs=3, lr=learning_rate, wd=weight_decay)
+    assert learn.opt.lr == learn.lr_range(learning_rate)
+    assert learn.opt.wd == weight_decay
+    ## TO CHECK: good idea to assert over this output?    
+    captured = capsys.readouterr()
+    ##match_epoch = re.findall(r'1/3 ', captured.out) ## finds Epoch 1/3
+    ##assert match_epoch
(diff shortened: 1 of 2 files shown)


Benudek:
@stas00 this works locally, but here cannot see the azure pipe with all tests start. There was an error (not sure why) in previous branch on the asserts over UI.

Benudek:
@stas00 this below line fails, even though locally it works.

Any idea, what the issue might be? Happy to use your new class (https://github.com/fastai/fastai/blob/master/tests/utils/text.py#L18) if applicable.

thx

2019-01-10T23:09:37.7319000Z         match_hundperc = re.findall(r'100.00%', captured.out) ## finds 100% progress
2019-01-10T23:09:37.7319120Z >       assert match_hundperc

Benudek:
ok, took out the prints, adjusted the assert and with small change in string to fit: I have a working (?) version without -s option. But there seems to be a different error on one kernel:

https://dev.azure.com/fastdotai/fastai/_build/results?buildId=2751

This here: 

> fit and fit_one_cycle are almost identical, so probably the best to test most of their features together and not duplicate the same test in different files. Here is an example:
> https://github.com/fastai/fastai/blob/master/tests/test_callback.py#L65

looks indeed much better, thx for the tip and I logged as to do (next step). Hopefully, if this (not my code?) error is fixed, we can merge this one and I improve this then incrementally starting with following your example code

thx for your help @stas00 more than I can expect  and can learn

Benudek:
I do not have option [Rerun all] in the upper left corner.

Yes, I will make larger PRs with less commits next time. It's a learning journey but I want to avoid being a burden of course.

stas00:
Thank you for confirming that only those with commit rights can re-run the tests, @Benudek. Perhaps it'd be a good suggestion for github to allow PR submitters to re-run the tests too. Especially when it's not their fault that the test suite has failed. I hit re-run for this one.

All is good wrt to PRs, I was just sharing how every commit to an existing PR commit sends a notice to all, that's why grouping those together is a goodness. No damage done.

Benudek:
still failing this non-determin test, no idea. Take the point on commits & notifications though ;-)

sgugger:
Thanks for your PR. Concerning the tests for fit_one_cycle and lr_find, can we had a little bit more to check the learning rates (and momentums for 1cycle) behaved properly? For now you just test the output which doesn't tell us if everything went well or not.

Benudek:
agree, tests are not yet perfect. lets call it incremental ;-)

Will look at your comments

Benudek:
closing, new follow up PR: https://github.com/fastai/fastai/pull/1524

Issue #1455, opened 2019-01-11
Text classification with float target value: pad_collate cannot create target tensor


tchambon:
When creating a text classifier with float target values, pad_collate calls tensor() to create the target values array with  an array of array. This gives the error:  len() of unsized object

This is a gist of how to reproduce the bug and how to fix the bug for my use case : https://gist.github.com/tchambon/cd9ff7f8851efab1acbadd9207bcb38a

I would like to provide a PR for this but I am afraid this fix could have side effect for other type of target values (ex: for multiclass target values)

sgugger:
Thanks for flagging! I also added a test so that we see if regression suddenly fails in the future.

PR #1456, opened 2019-01-11
Fix issue #1455  Text classification with float target value: pad_col…

tchambon:
Proposal for a fix of the issue #1455: https://github.com/fastai/fastai/issues/1455


sgugger:
This fails in the case of multi-classification since the target is then an array (and not just a tensor). Working on another fix.

PR #1458, opened 2019-01-12
Update core docs (3rd try!)

bfarzin:
Numpy error removed
`help()` call in cells removed as well.


sgugger:
Thanks, it looks good!

Issue #1441, opened 2019-01-08
Overriding image points visual parameters for a better rendering


devforfu:
I am working on face landmarks prediction model and use `DataBunch.show_batch` method to visualize the data I have. Here is an example of the output I am getting (using the most recent version of the library):
![faces](https://user-images.githubusercontent.com/6022009/50816875-c3c95900-1343-11e9-865e-88324935d951.png)

You see, the landmarks points are very very small and poorly visible on some of the images. For example, this styling probably looks a bit better (I mean, markers size and coloring, the titles are irrelevant to my question):

![faces2](https://user-images.githubusercontent.com/6022009/50816957-0be87b80-1344-11e9-980c-79701d721493.png)

There is [a line](https://github.com/fastai/fastai/blob/c1e2a8dafc643a41fe73545f3deb627b64a6544c/fastai/vision/image.py#L319) that uses some constant values to define the rending style. I wonder, would it be more convenient to make these parameters more flexible? Like:
```python
def show(self, ax:plt.Axes=None, figsize:tuple=(3,3), title:Optional[str]=None, 
         hide_axis:bool=True, **kwargs):
    "Show the `ImagePoints` on `ax`."
    if ax is None: _,ax = plt.subplots(figsize=figsize)
    pnt = scale_flow(FlowField(self.size, self.data), to_unit=False).flow.flip(1)
    ax.scatter(pnt[:, 0], pnt[:, 1], 
               s=kwargs.get('s', 10), 
               marker=kwargs.get('marker', '.'), 
               c=kwargs.get('c', 'r'))
    if hide_axis: ax.axis('off')
    if title: ax.set_title(title)
```
Or even: 
```python
# or it could be some function instead that returns default params
params = {'s': 10, 'marker': '.', 'c': 'r'} 
ax.scatter(pnt[:, 0], pnt[:, 1], **params)
```
Would it be reasonable to include something like this into the library?

sgugger:
You can definitely suggest a PR with kwargs that get passed to ax.scatter.

sgugger:
PR was merges so closing this.

PR #1462, opened 2019-01-13
More flexible visual styling for ImagePoints.show

devforfu:
In this PR, an attempt to fix issue #1441 is made. The `ImagePoints.show` method `kwargs` are directly passed into `ax.scatter` (including original default values if nothing is overridden by caller).


sgugger:
Thanks, I believe you wanted to pass **params and not **kwargs, and I refactored a little bit.

PR #1465, opened 2019-01-14
added a nice dataset-builder, utilizing Bing API

lreardon:
This pull contains a script for utilizing Bing API to build datasets, called "search_bing_api.py" . 
I tried to open a new thread on the fastai forums page, but I couldn't figure out how to add a thread.
Sorry this isn't following protocol.

### DIFF
--- a/search_bing_api.py
+++ b/search_bing_api.py
@@ -1,0 +1,112 @@
+# import the necessary packages
+from requests import exceptions
+import argparse
+import requests
+import cv2
+import os
+
+# construct the argument parser and parse the arguments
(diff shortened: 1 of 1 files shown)


xnutsive:
Hey, thanks for the pull request! 

You might want to look at #1382 (or lookup `fastai.widgets.image_downloader`), and see if there's a good way to make those two play together?

jph00:
Thanks for the PR! It's very cool, but not a great fit with the functionality in the library, so it would be best if you put it in a separate repo.

To post on the forum, you simply need to read a few threads and wait a few mins after you first sign up.

PR #1468, opened 2019-01-14
adding timeout handling and re-request of data

bfarzin:
Adding retries to the untar(URL)
Also a more informative message if you try to download a truncated file.

### DIFF
--- a/docs_src/core.ipynb
+++ b/docs_src/core.ipynb
@@ -57,7 +57,7 @@
     {
      "data": {
       "text/markdown": [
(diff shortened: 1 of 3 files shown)


stas00:
I'm not sure this PR should add `IPython.notebook.save_notebook()` - we either need to add it to all docs, or none. This is unrelated to the issue this PR is fixing.

Looking good otherwise, with the only minor tweak required is that the location of `~/.fastai/data` is user configurable, so it should report the actual user configuration and not the default.

PR #1470, opened 2019-01-15
new version af languagePreModelLoader validated by a new test-case

kasparlund:
The LanguagePreModelLoader follows the same callback structure as the existing version in fastai. 
Functional differences are: 
-It is uses all tokens in the user provided tokens in each epoch. drop_last has therefore been removed
-alle arrays are preallocated
-the formate of the user provided tokens (fx uint169 are converted implicitly during the fill of the batch
-about 16% faster than the existing version
-the stop criteria are simple
-if more tokens are required to fill the last batch these tokens are fetch at the head (backwards=False) of tail (backwards=True) of the jagged array

The new text cases verify that the batches are continuous across batches.


fastai/text/data.py
```diff
@@ -13,44 +13,106 @@
...
-        self.dataset,self.bs,self.bptt,self.backwards = dataset,bs,bptt,backwards
-        self.shuffle,self.drop_last = shuffle,drop_last
-        self.lengths = np.array(ifnone(lengths, [len(o) for o in dataset.x.items]))
-        self.n = self.lengths.sum() // self.bs
-        
-    def __len__(self):
-        if self.item is not None:    return 1
-        if (self.n-1)%self.bptt == 0 or self.drop_last: return ((self.n-1) // self.bptt) * self.bs
-        return ((self.n-1) // self.bptt + 1) * self.bs
+    "Create a dataloader with bptt slightly changing."
```
sgugger: Please use the previous description as bptt isn't changing anymore.

fastai/text/data.py
```diff
@@ -13,44 +13,106 @@
...
     
+    class CircularIndex():
```
sgugger: I'm not seeing why this class is so complicated: foward can be a simple attribute here, why are you using a private variable and a setter?

fastai/text/data.py
```diff
@@ -13,44 +13,106 @@
...
+        #When the index exceeds the length of self.idx then it is wrap to start at the head 
+        #or the end if indexing is moving backwards. The shuffled is done in-place, 
+        def __init__(self, length:int, forward:bool): 
+            self.idx      = np.arange(length)
+            self.forward_ = forward
+        def __getitem__(self, i): 
+            idx = self.idx
+            return idx[ i%len(idx) if self.forward_ else len(idx)-1-i%len(idx) ]
+        def __len__(self) -> int: return len(self.idx)
+        def shuffle(self): np.random.shuffle(self.idx)
```
sgugger: Not sure this is done inplace, shouldn't you store the result in self.idx?

fastai/text/data.py
```diff
@@ -13,44 +13,106 @@
...
+        def forward(self, forward:bool=True): self.forward_ = forward
+
+    def __init__(self, dataset:LabelList, bs:int=32, bptt:int=70, backwards:bool=False, shuffle:bool=False):
```
sgugger: I really liked the idea of being able to pass lengths, which would be computed only if not passed (since it can be long on a giant dataset). Could you add this to your PR?

fastai/text/data.py
```diff
@@ -13,44 +13,106 @@
...
+        self.ite_len   = self.bs*int( math.ceil( self.totalToks/(self.bptt*self.bs) )) if self.item is None else 1
+        self.idx = None
+
+    def __len__(self): return self.ite_len
     def __getattr__(self,k:str)->Any: return getattr(self.dataset, k)
-    
+   
+    def allocate_buffers(self):     
+        "allocate the required worth-case batchbuffer"
+        #idx:     used to index into items so that shuffle is used - if idx has been shuffled
```
sgugger: This is all for the documentation of fill_forward I believe, and has nothing to do here.
Can we have better names for ei and eo?

fastai/text/data.py
```diff
@@ -13,44 +13,106 @@
...
+
     def on_epoch_begin(self, **kwargs):
-        self.idxs = np.random.permutation(len(self.dataset)) if self.shuffle else arange_of(self.dataset)
-        self.text_idx = np.concatenate([[0],self.lengths[self.idxs].cumsum()])
-        
+        if self.idx is None: self.allocate_buffers()
+        if self.shuffle:     self.idx.shuffle()
+        self.idx.forward(self.backwards is False) 
+
+        #set up ei and eo to index the data so batches has continuous rows
```
sgugger: This whole part can be vectorized (see the code you're replacing).

fastai/text/data.py
```diff
@@ -13,44 +13,106 @@
...
-        #Grabbing the texts we need in the dataset
-        if self.backwards: concat = np.concatenate([self.dataset.x.items[j][::-1] for j in self.idxs[start:end]])
-        else: concat = np.concatenate([self.dataset.x.items[j] for j in self.idxs[start:end]])
-        #Returning the right portion.
-        start_idx = i-self.text_idx[start]
-        return concat[start_idx:start_idx+seq_len], concat[start_idx+1:start_idx+seq_len+1]
+        if self.idx is None:      self.on_epoch_begin()
+
+        j = k % self.bs
+        if self.backwards: self.ei[j],self.eo[j] = self.fill_backward(  self.dataset.x.items, self.idx, self.batch[j],
```
sgugger: `fill_backward` and `fill_forward` only have two lines of code that differ. They should be put together as `fill` with a `backward` argument, that would help refactor a lot of dupe code.

sgugger:
I'll review the tests later, for now I'd like us to refactor and clean up the main bit.

fastai/text/data.py
kasparlund: I have passed hours on the existing code. The i idea here i to avoid allocations by slicing like so row[ibuf:ibuf+n] = rag[eo:eo+n]. this statement also handle type conversion implicitly if you passe np.uin16. 

I normallý go to great lengths to avoid loops i python and R because they are so slow. In this case using a loop is faster than vectorising. I guess the reason is that in most case the loop i only executed once. 

Let me make som measurements whether that is true

fastai/text/data.py
kasparlund: what about ro and ri
ro being the offset where each batch row starts and ri being the index into the current rag

fastai/text/data.py
kasparlund: ok i will measure whether it makes the code faster

fastai/text/data.py
kasparlund: https://docs.scipy.org/doc/numpy-1.15.0/reference/generated/numpy.random.shuffle.html

fastai/text/data.py
kasparlund: idx = self.idx
using idx is faster than using self.idx even if it is only used twice in the forwards use.
https://stackoverflow.com/questions/12397984/why-is-local-variable-access-faster-than-class-member-access-in-python

%timeit confirms this

fastai/text/data.py
kasparlund: measurements using bptt=70 shows that 48% of the calls to fill_forward pass 1 time through the loop, about 25% pass 2 times and the rest pass through the loop 3-7 times to build a row in the batch. If you have many short sentences it will take more loop than if you have sentence with 70 tokens or more. 
I think this explain why this approach is faster then the vectorisation in this special case

fastai/text/data.py
kasparlund: This code show that we can save about 1.5 seconds for a very long ragged array (25 mio)

`lengths` is an 25 mio long array of lengths
`trainIDS_` is an 25 mio long array of arrays of tokens

def l1(lengths): 
...totalToks=0
...for l in lengths: totalToks += l
...return totalToks
def l2(items): 
...totalToks=0
...for rag in items: totalToks += len(rag)
...return totalToks    
print(f"number of rags:{len(trainIDS_)}")
%timeit l1(lengths)
%timeit l2(trainIDS_)

The result is:
> number of rags:24871686
> 2.45 s ± 55.5 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
> 3.94 s ± 27.9 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

It did make me go back and optimize some loops where i did not yet use local variable.
I have also modified the code to run with lengths=None and with a user provide lengths for the trainingdata.

However do not think the 1.5 seconds pr epoch justify the extra complexity of the argument list and the code.  What do you think ?

fastai/text/data.py
kasparlund: done

kasparlund:
Hi Sylvain. I have implemented all your reguest except use of lengths.

I did make a full implementation using length i have however reverted to not using it. As shown in previous communication we improve, all thing equal, performance by about 1.5 seconds pr epoch. and possibly by 1.7s if there was a length for both training and validation. I do not think this justifies the extra complexity of the interface and the code, but i can reimplement it if you insist.
However, working with lengths led to a couple of extra optimisations that i have now applied so something good did come out of it.

By the way the new testcase "test_text_languagemodelpreloader" has been invaluable at catching error during refactoring. There is just something about testing against randomly varying data:)

sgugger:
For the lengths, I'd really like not to use a functionality that was existing, even if its benefit is slim.
I prefer the version in comments for the `fill_row` function.
Good question for the pytorch dataloader: maybe it would be best to fill the needed row in getitem and get rid of fill_batch altogether?

sgugger:
I'll finish to fix the potential failing tests after merging.

sgugger:
Ok, it was just the batch_first thing that made the merge fail, thanks for your contribution!

PR #1484, opened 2019-01-19
Removed duplicate RNNLearner in fastai.text.learner.__all__

xnutsive:
Minor, shouldn't brake anything. Playing around with `fastai.text` and will fix small things along the way.


sgugger:
Thanks!

PR #1485, opened 2019-01-19
Add docs in torch_core

nvs-abhilash:
Add docs for `Functions that operate conversions` section in docs_src/torch_core.ipynb
Rephrased function doc for `requires_grad` function in `fastai/torch_core.py`

### DIFF
--- a/docs_src/torch_core.ipynb
+++ b/docs_src/torch_core.ipynb
@@ -221,6 +221,66 @@
    ]
   },
   {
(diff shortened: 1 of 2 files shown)


nvs-abhilash:
Please do suggest if any rephrasing is required. I think `to_data` docs seems a overkill, but I was unable to come up with something better. Any feedback is welcome :)

sgugger:
It all looks pretty good. Thanks!

PR #1489, opened 2019-01-20
Implemented cutout transform

xnutsive:
This implements adds cutout to vision.transforms. It was implemented in 0.7, but not in 1.0. 

@sgugger please feel free to discard this if this is a crappy implementation. I had an hour of spare time and I was curious if I can learn how transforms work and implement this. 

The implementation supports default `n_holes` and `length` values (based somewhere around the paper's recommended values, I didn't test their performance myself yet), and also supports randomizing those (they're both `uniform_int` with default values).

The transform is not included by default, but the user can include it via `xtra_tfms` in `get_transforms()`.


sgugger:
Just removed the mask as we can directly 0 the elements of x.

sgugger:
Thanks a lot for implementing it!

PR #1498, opened 2019-01-22
Fixed a regression bug in databunch.show_batch(1) subplots handling

xnutsive:
@sgugger, seems like Jeremy implemented a workaround in `show_xys` to handle a situation when `subplots()` was called for a single subplot, but it stopped working. 

This handles subplots if a single row, column, or a single subplot (1x1) was requested and returns it as `[[ax]]`, and then `show_xys()` flattens it anyway. I also changed `show_xyzs()` to accommodate that. Checked if it brakes any tests or `docs_nbs` full tests: there are a few errors, but they're unrelated (!). Added a few tests for `subplots()` in `core` to make sure the output stays consistent.

With that, `show_batch(1)` should work. 

Also implemented a workaround for a situation when user wants to `show_batch()` for a databunch with small batch size, i.e. 

```
data = src.databunch(bs=4)
data.show_batch(5)
```

will render 2x2 images and won't raise an error.

### DIFF
--- a/fastai/basic_data.py
+++ b/fastai/basic_data.py
@@ -157,12 +157,14 @@ class DataBunch():
     def show_batch(self, rows:int=5, ds_type:DatasetType=DatasetType.Train, **kwargs)->None:
         "Show a batch of data in `ds_type` on a few `rows`."
         x,y = self.one_batch(ds_type, True, True)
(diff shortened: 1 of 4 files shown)


sgugger:
Thanks! Will build up on top of that.

Issue #1504, opened 2019-01-24
Dummy input size in `num_features_model` too small


PPPW:
I tried to use inceptionv4 with fastai v1, however I failed because the `num_features_model` function doesn't works for the `Reduction` layers: the default dummy input size is set to `(64,64)`, which is too small in this case. It will raise an error and fails to get the number of features. 

If I run `model_sizes` with `size=(96,96)`, I got the correct number of features (1536 x 2), and everything works fine now. 

For general cases, I propose we change `num_features_model` to:
```
def num_features_model(m:nn.Module)->int:
    sz = 64
    while True:
        try:            
            return model_sizes(m, size=(sz,sz))[-1][1]    
        except Exception as e:            
            sz *= 2            
            if sz > 2048: raise e
```

It works fine for me, as I tried more different models. Thanks!

PR #1513, opened 2019-01-25
adapt dummy input size for general cases

PPPW:
Issue #1504 : dummy input size in `num_features_model` is too small for some architectures (i.e., inceptionv4). 

Solution: double the input size until it works. Stop if input size is already pretty large, because in this case the error is probably not due to the input size.


sgugger:
Thanks!

PR #1519, opened 2019-01-26
Now SaveModelCallback prints out a message when a better model is saved.

odysseus0:
This is a very minor convenience patch. All it does is that now when training with SaveModelCallback and a better model is found after an epoch, the callback will prints out a message that says something like:

better model found at epoch 4 with accuracy value: 0.975.

It is helpful in the sense that the user can immediately know the best metrics score his saved model has obtained.

Sometimes, we need a lot of epochs. Sometimes, metrics score drops after an epoch. Sometimes, we need to look through all the output to find the best metrics score. With this minor patch, we can just look at the bottom of the output and there the best score will lie. Cheers!


sgugger:
Thanks!

PR #1520, opened 2019-01-26
Text classifier - backwards option

tpietruszka:
In the original ULMFiT paper it is mentioned that ensembling results of a forward and backward model improves accuracy. Training a backwards language model is currently supported in code, but there seems to be no support for classification based on backwards models.

This PR introduces that functionality to TextClasDataBunch. I did not start an issue as this seems to just complete a functionality that is half-present and mentioned as expected.

### DIFF
--- a/fastai/text/data.py
+++ b/fastai/text/data.py
@@ -124,14 +124,17 @@ class SortishSampler(Sampler):
         sort_idx = np.concatenate((ck_idx[0], sort_idx))
         return iter(sort_idx)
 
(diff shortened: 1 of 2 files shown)


sgugger:
Good idea! Can you just amend your test to use a tiny dataframe and not the whole IMDB sample? I'd like it to be as fast as possible.

tpietruszka:
Thanks.

It actually uses the tiny dataframe from text_df(), the imdb_sample path is just to provide a path. I followed the example of other tests in that file

sgugger:
Sorry I misread... Thanks a lot!

PR #1523, opened 2019-01-27
include more torchvision models

PPPW:
add model metadata: 
densenet121
densenet169
densenet201
densenet161
vgg16_bn
vgg19_bn
alexnet


sgugger:
Thanks a lot (failing test is a random one with no link)!

PR #1526, opened 2019-01-27
handle corner case in bb_pad_collate

pouannes:
When the `Image` didn't contain any object to detect (for example if it was croped) the line 

```
bboxes[i,-len(lbls):] = bbs
```
was throwing an error (and the next one would too).
It's ok to do nothing when there's no object to detect.


pouannes:
So I was thinking about this and I'm not sure about something. 
If there's no object in the image at all because of transforms this fix handles it. However I'm not sure of the behaviour if and when the whole image is inside an object BBox. The correct behaviour (IMHO) would be to assign the class of the overall object that got zoomed into to the whole image, and the BBox should then be the whole image. I don't think that's what it's doing right now.
Maybe it's such a corner case it doesn't matter that much.

sgugger:
Seems like a sound suggestion, thanks!

PR #1534, opened 2019-01-29
Fix QRNN layer

mkardas:
After investigating an issue with a near zero train accuracy of ULMFiT with QRNN I've noticed that forget_mult_cuda_* kernels are not run at all on my machine. The problem was that pytorch's CUDA version (9.0) and the system's default CUDA version (9.2) mismatched. Even though pytorch is mostly independent of system's CUDA, `torch.utils.cpp_extension.load` uses the default CUDA installation for JIT-compiling cpp extensions. As a result, the kernels were silently ignored.

This PR:
* adds checks of the last error after a kernel call. On a problematic environment a runtime exception (`CUDA driver version is insufficient for CUDA runtime version`) is raised.
* adds some basic tests of `ForgetMult*`
* restores forget_mult_cuda.cpp and updates path
* adds conditional loading of the cpp extension
* defines `cuda` pytest marker to denote tests that require CUDA to run
* simplifies `can_use_gpu` (checking of `CUDA_VISIBLE_DEVICES` was redundant, as `cuda.is_available()` checks it indirectly by making sure at least one CUDA device is available)

### DIFF
--- a/docs/dev/test.md
+++ b/docs/dev/test.md
@@ -436,6 +436,8 @@ custom options:
    pytest --skipint
    ```
 
+* `cuda` - mark tests as requiring a CUDA device to run (skipped if no such device is present). These tests check CUDA-specific code, e.g., compiling and running kernels or GPU version of function's `forward`/`backward` methods.
+
 
 ### After test cleanup
 
--- a/fastai/text/models/forget_mult_cuda.cpp
+++ b/fastai/text/models/forget_mult_cuda.cpp
@@ -1,0 +1,31 @@
+#include <torch/torch.h>
+
+#include <vector>
+
+// CUDA forward declarations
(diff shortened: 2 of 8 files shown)


sgugger:
Thanks a lot for cleaning the mess I made when moving the models. Just restored the `can_use_gpu` function as it's useful for tests to sometime masks an existing GPU.

mkardas:
Thanks for reviewing. I've tested `can_use_gpu()` locally with this:
```console
$ CUDA_VISIBLE_DEVICES='' python -c "import torch; print(torch.cuda.is_available())"
False
$ python -c "import torch; print(torch.cuda.is_available())"
True
```
and then, to make sure, checked it here: https://github.com/pytorch/pytorch/blob/2235fb256e9e58050585a1bd0342de599570fcce/torch/cuda/__init__.py#L50

sgugger:
You're right, works like a charm. Thanks!

PR #1537, opened 2019-01-29
Adapt ClassificationInterpretation class to work with all data types

jchaykow:


sgugger:
This is a great idea! But there is nothing in your code that limits this to tabular, it would work in all the applications. So I think this should go in train.py, to be accessible from every application, then in vision we can monkey-patch the specific methods (mostly the polot_losses one). Can you make the move to train, and add the most_confused method? Then I can do the monkey-patching for the ` ClassificationInterpration` in `vision.learner` if you're not comfortable with that.

jchaykow:
I took a go at the monkey-patching, not sure if it's exactly what you were imagining. Tests ran ok - here is a [gist](https://gist.github.com/jchaykow/b0e970867ca4e9e51808986a81ea1d2d). One thing I was unsure of is how you want to integrate the TTA functionality in `ClassificationInterpretation` for `vision`, I did not reimplement TTA so may have to consider where to place that in `from_learner`.

jchaykow:
I see the issue. `_learner_interpret` had a TTA call in it and I removed that temporarily from `ClassificationInterpretation`

fastai/vision/learner.py
```diff
@@ -95,119 +96,114 @@ def unet_learner(data:DataBunch, arch:Callable, pretrained:bool=True, blur_final
...
-        else:
-            cm = torch.zeros(self.data.c, self.data.c, dtype=x.dtype)
-            for i in range(0, self.y_true.shape[0], slice_size):
-                cm_slice = ((self.pred_class[i:i+slice_size]==x[:,None])
-                            & (self.y_true[i:i+slice_size]==x[:,None,None])).sum(2)
-                torch.add(cm, cm_slice, out=cm)
-        return to_np(cm)
-
-    def plot_confusion_matrix(self, normalize:bool=False, title:str='Confusion matrix', cmap:Any="Blues", norm_dec:int=2,
+def top_losses(self, k:int=None, largest=True):
```
sgugger: This could go in the general `ClassificationInterpretation` too, it's not specific to vision.

fastai/vision/learner.py
```diff
@@ -95,119 +96,114 @@ def unet_learner(data:DataBunch, arch:Callable, pretrained:bool=True, blur_final
...
+        for clas in infolist[ordlosses_idxs[sampleN]][2]:
+            actualclasses=actualclasses+' -- '+str(classes_ids[clas][1])
+        imag=mismatches_ordered_byloss[sampleN][0]
+        imag=show_image(imag, figsize=figsz)
+        imag.set_title(f"""Predicted: {classes_ids[infolist[ordlosses_idxs[sampleN]][1]][1]} \nActual: {actualclasses}\nLoss: {infolist[ordlosses_idxs[sampleN]][3]}\nProbability: {infolist[ordlosses_idxs[sampleN]][4]}""",
+                        loc='left')
+        plt.show()
+        if save_misclassified: return mismatchescontainer
+
+def confusion_matrix(self, slice_size:int=None):
```
sgugger: It's weird to have two implementations of the same thing. There should only be one `confusion_matrix` in the general `ClassificationInterpretation`.

fastai/vision/learner.py
```diff
@@ -95,119 +96,114 @@ def unet_learner(data:DataBunch, arch:Callable, pretrained:bool=True, blur_final
...
+    thresh = cm.max() / 2.
+    for i, j in itertools.product(range(cm.shape[0]), range(cm.shape[1])):
+        coeff = f'{cm[i, j]:.{norm_dec}f}' if normalize else f'{cm[i, j]}'
+        plt.text(j, i, coeff, horizontalalignment="center", color="white" if cm[i, j] > thresh else "black")
+
+    plt.tight_layout()
+    plt.ylabel('Actual')
+    plt.xlabel('Predicted')
+
+def most_confused(self, min_val:int=1, slice_size:int=None)->Collection[Tuple[str,str,int]]:
```
sgugger: Same for most confused.

fastai/vision/learner.py
```diff
@@ -95,119 +96,114 @@ def unet_learner(data:DataBunch, arch:Callable, pretrained:bool=True, blur_final
...
+    return sorted(res, key=itemgetter(2), reverse=True)
+
+ClassificationInterpretation.top_losses = top_losses
+ClassificationInterpretation.plot_top_losses = plot_top_losses
+ClassificationInterpretation.plot_multi_top_losses = plot_multi_top_losses
+ClassificationInterpretation.confusion_matrix = confusion_matrix
+ClassificationInterpretation.plot_confusion_matrix = plot_confusion_matrix
+ClassificationInterpretation.most_confused = most_confused
+
+def _learner_interpret(learn:Learner, ds_type:DatasetType=DatasetType.Valid):
```
sgugger: I believe it's here we can put back the tta flag.

fastai/vision/learner.py
```diff
@@ -95,119 +96,114 @@ def unet_learner(data:DataBunch, arch:Callable, pretrained:bool=True, blur_final
...
+
+def most_confused(self, min_val:int=1, slice_size:int=None)->Collection[Tuple[str,str,int]]:
+    "Sorted descending list of largest non-diagonal entries of confusion matrix."
+    cm = self.confusion_matrix(slice_size=slice_size)
+    np.fill_diagonal(cm, 0)
+    res = [(self.data.classes[i],self.data.classes[j],cm[i,j])
+            for i,j in zip(*np.where(cm>min_val))]
+    return sorted(res, key=itemgetter(2), reverse=True)
+
+ClassificationInterpretation.top_losses = top_losses
```
sgugger: Only `plot_top_losses` and `plot_multi_top_losses` need to be monkey-patched, they are the only true new methods. And `from_learner` if we want to add TTA.

sgugger:
Thanks! We're almost there I believe, just requested a few more changes.

jchaykow:
[Gist](https://gist.github.com/jchaykow/05b066935120ae71aa242c654b4a8367)

sgugger:
Ok, just made the methods private before the monkey-patch. Thanks a lot!

PR #1541, opened 2019-01-30
Remove nb_conda dependency

dhirschfeld:
AFAICS `fastai` the library doesn't depend on `nb_conda` at all so it shouldn't be specified in the dependencies of the library.

If you want to suggest a default environment for users to run then it should be included in the [`environment.yml`](https://github.com/fastai/fastai/blob/master/environment.yml) rather than directly in the library dependencies.

***xref:*** https://github.com/fastai/fastai/commit/1673cbc964f0faa15a773261ada3ea55cafb3419

I really don't like `nb_conda` and I don't want to be forced to install it in my environment - it makes intrusive changes I don't want to the behaviour of notebook kernel discovery. More to the point it's [currently broken](https://github.com/AnacondaRecipes/nb_conda_kernels-feedstock/issues/1), thereby preventing me from using `fastai`

### DIFF
--- a/conda/meta.yaml
+++ b/conda/meta.yaml
@@ -57,7 +57,6 @@ requirements:
     - pandas
     - bottleneck
     - numexpr
-    - nb_conda
     - nvidia-ml-py3
     - packaging
     - pillow


stas00:

stas00:
I added it since so many people had issues as you can see from https://stackoverflow.com/questions/39604271/conda-environments-not-showing-up-in-jupyter-notebook and several threads on https://forums.fast.ai/ and which this package resolved.

But you're correct that this shouldn't be in the core dependencies since fastai should work w/o jupyter environment. It just didn't seem to be a harmless bonus dependency to add to aid the users. But as you are saying this is not the case. So I'm in agreement with your proposed change, @dhirschfeld.

Instead, we should document that some people may need nb_conda if they have issues with their jupyter/conda env.

dhirschfeld:
Yeah, for new users it can be nice for conda environments to just show up magically but last time I tried it there was no way to prevent it or select environments so it was a non-starter for me.

With the new [enhanced pip-interoperability](https://www.anaconda.com/blog/developer-blog/conda-4-6-release/) it's probably feasible to use pip in the interim.

stas00:
This optional dependency is now documented here:
https://docs.fast.ai/install.html#jupyter-notebook-dependencies

dhirschfeld:
Thanks for the quick response!

PR #1545, opened 2019-01-31
Updated comment in plot_top_losses

ksivasan:
changed to "Show images in `top_losses` along with their prediction, actual, loss, and probability of actual class."


sgugger:
Closing since no news in a week. Fell free to reopen but we can't merge your PR if you don't sign the CLA.

PR #1548, opened 2019-01-31
Distributed code should now work with LanguageLearner

bearpelican:
Fixing edge cases for distributed training with RNNs


fastai/distributed.py
```diff
@@ -5,6 +5,10 @@
 
 __all__ = ['DistributedRecorder', 'DistributedTrainer', 'read_metrics', 'setup_distrib']
 
+def rnn_reset(self):
```
bearpelican: monkey patching DistributedDataParallel so that it calls model.reset

fastai/distributed.py
```diff
@@ -16,7 +20,7 @@ class DistributedTrainer(LearnerCallback):
     def on_train_begin(self, **kwargs):
         self.learn.model = DistributedDataParallel(self.learn.model, device_ids=[self.cuda_id],
                                                    output_device=self.cuda_id)
-        train_sampler = DistributedSampler(self.learn.data.train_ds)
+        train_sampler = DistributedSampler(self.learn.data.train_dl.dataset)
```
bearpelican: Making sure we pass the right dataset to sampler

For language models, data.train_dl.dataset != data.train_ds.
data.train_dl.dataset is a LanguageModelPreLoader
data.train_ds is a LabelList

PR #1556, opened 2019-02-02
enable users to run part of the 1cycle policy

PPPW:
Enable running part of the full 1cycle policy by specifying the start epoch and total epochs.


sgugger:
Looks great, thanks a lot! Can you also change the function in `train.py` to use those new arguments you added?

PPPW:
Hi @sgugger, currently these arguments are passed by the "kwargs" in `fit_one_cycle` of `train.py` so it's working as I tested, I was thinking about making minimal changes... Let me know whether I'd like to add that, thanks for checking!

sgugger:
Oh good point! Although we'd like to stop hiding things in kwargs to make it easier for users to follow what arguments a function takes (in the docs + with tab completion), but that will be for later.

PR #1563, opened 2019-02-03
Replaced "LINK REQUIRED" with link to the referred paper & datasets

BrittonWinterrose:
I believe this is the arXiv paper referenced in the missing link.  "Character-level Convolutional Networks for Text Classification" by Xiang Zhang, Junbo Zhao and Yann LeCun (NIPS 2015 - arXiv:1509.01626v3 [cs.LG] ) is where they utilized eight datasets and stored them each with an associated "test.csv",  "train.csv" and "classes.txt".   This matches @jph00  explanation from the course video when describing the reason for utilizing this newly evolving standard format.  

(Structure of the actual datasets can be seen by the link to Xiang Zhang's google drive link with the associated datasets from the paper. )
screen shot 2019-02-03 at 2 36 43 pm

Lesson 10 video at the time where @jph00  talks about this section:
https://youtu.be/h5Tz7gZT9Fo?t=1451

### DIFF
--- a/courses/dl2/imdb.ipynb
+++ b/courses/dl2/imdb.ipynb
@@ -190,7 +190,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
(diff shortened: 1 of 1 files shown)


sgugger:
Thanks!

PR #1568, opened 2019-02-05
Fix `plot_confusion_matrix` when `normalize=True`

marsbroshok:
When `normalize=True` font color selection between black or white does not work properly - it is not taking into account normalized value, but the original one.
Move confusion matrix value normalization before plotting to fix font color choice behavior.


sgugger:
Thanks!

PR #1569, opened 2019-02-05
TransformerXL update

bfarzin:
I know this is all beta, but in trying to play around with it I noticed that you need to cast `pos` to a float tensor to process the Sine/Cosine functions.  Also, when `mem_len=0` there are a few exceptions or you can't do the `lr_find` and train.

Hopefully helpful.  If not, you can ignore.


bfarzin:
I don't know why this Linux_PyPI Python37 test is failing when others are passing.  Is that something I can fix on my side?  I can't find more details about that failure.

stas00:
The test suite is non-deterministic and occasionally fails on one of the tests. I have re-run the failing CI instance and all is good.

sgugger:
I see the use of your first change (pos.float()) but what about the other two? I don't see how you could have a hidden size if mem_len is set to 0.

bfarzin:
My understanding was `mem_len = 0` means that there is no memory, so you don't have anything to update to the hidden state.  What am I missing there?
When I ran with mem_len = 0 the first batch will run fine, but the second batch will have a "memory" that is not `None` and now you collide with the shape of the input and you throw an error.  Maybe I was calling it wrong?

sgugger:
Yeah but if mem_len is 0 there is no hidden state, so the current tests should be enough.

bfarzin:
I get this error with a simple example (and added this code to "fix" that problem.  Maybe I am going about it wrong or have some strange param setup?  When I stepped into that part of the code, it looks like one has a memory, but the other does not when `mem_len = 0` . What test should I be looking at?

```
from fastai import *
from fastai.text import *
path = untar_data(URLs.IMDB_SAMPLE)
data = TextLMDataBunch.from_csv(path, 'texts.csv')
vocab_size = len(data.vocab.itos)
transformer_model = get_transformerXL_lm(vocab_size,ctx_len=1,mem_len=0)
learn = LanguageLearner(data,transformer_model)
lr_find(learn)
```

Error:

```
~/fast_ai/fastai-fork/fastai/text/models/transformer.py in _apply_attention(self, x, r, u, v, mask, mem)
    112         AC = torch.matmul(wq+u,wk)
    113         BD = _line_shift(torch.matmul(wq+v, wkr))
--> 114         if self.scale: attn_score = (AC + BD).mul_(1/(self.d_head ** 0.5))
    115         if mask is not None:
    116             attn_score = attn_score.float().masked_fill(mask, -float('inf')).type_as(attn_score)

RuntimeError: The size of tensor a (70) must match the size of tensor b (140) at non-singleton dimension 3
```

PR #1570, opened 2019-02-05
Add comma separator for numbers in the thousands for better readability

iamshwin:
This will change: `Total params: 21831525` --> `Total params: 21,831,525` in the `model_summary`.

(I appreciate this is a trivial change, but it does help readability.)

Thanks for reviewing 🙏

### DIFF
--- a/fastai/callbacks/hooks.py
+++ b/fastai/callbacks/hooks.py
@@ -176,11 +176,11 @@ def model_summary(m:Collection[nn.Module], n:int=70):
         total_params += int(params)
         total_trainable_params += int(params) * trainable
         params, size, trainable = str(params), str(list(size)), str(trainable)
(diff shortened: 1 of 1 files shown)


sgugger:
It does indeed, thanks!

sgugger:
So we need to put something else to add this comma. I haven't looked at it in depth, but please make sure that the tests in `test_callbacks_hooks.py` are passing.

In this PR the line `res += f"{layer:<20} {size:<20} {params:<10,} {trainable:<10}\n` throws the error
` ValueError: Cannot specify ',' with 's'.`

It's the good idea though, so if you can fix it I'd definitely merge it!

iamshwin:
Ah - ok have fixed this in #1573 @sgugger

Issue #1575, opened 2019-02-06
DeviceDataLoader doesn't support such Dataset that emits dict.


tamuhey:
The below code raises an error:
```python
# for test dataset class
class DictDataset(Dataset):
    def __getitem__(self, idx):
        return {"a":np.array([1,2,3]),"b":np.array([2,3,4])}
    def __len__(self):
        return 10
dl=DeviceDataLoader.create(DictDataset(), batch_size=3)
next(iter(dl))
```
```
~/work/fastai/fastai/torch_core.py in to_device(b, device)
    109     device = ifnone(device, defaults.device)
    110     if is_listy(b): return [to_device(o, device) for o in b]
--> 111     return b.to(device)
    112 
    113 def data_collate(batch:ItemsList)->Tensor:
AttributeError: 'dict' object has no attribute 'to'
```

While pytorch's `DataLoader` doesn't: 
```python
dl=DeviceDataLoader.create(DictDataset())
next(iter(dl))
```
```
{'a': tensor([[1, 2, 3],
         [1, 2, 3],
         [1, 2, 3]]), 'b': tensor([[2, 3, 4],
         [2, 3, 4],
         [2, 3, 4]])}
```

I think this is because `torch_core.to_device` function doesn't support `dict`.

https://github.com/fastai/fastai/blob/6fe1e1b6e80e4b7e176bbc54ac2d0312b1bdbf1d/fastai/torch_core.py#L107

sgugger:
A `DeviceDataLoader` doesn't take a DataSet but a `DataLoader`, so I don't think there is a bug there. I'm not sure the fastai library handles inputs for models that are dictionaries. What is your use case?

tamuhey:
@sgugger 
Sorry, not `DeviceDataLoader.__init__` but `DeviceDataLoader.create`.
I use this class for those model that I create.

Dictionary is convenient and expository when handling multiple types of data, suitable for graph nn, multitask learning, and so on.

sgugger:
PR was merged.

PR #1576, opened 2019-02-06
ENH: DeviceDataLoader now accepts dict based Dataset

tamuhey:
- from #1575

# Changes

Before this PR, the below code raises an error: 
```python
# for test dataset class
class DictDataset(Dataset):
    def __getitem__(self, idx):
        return {"a":np.array([1,2,3]),"b":np.array([2,3,4])}
    def __len__(self):
        return 10
dl=DeviceDataLoader.create(DictDataset(), batch_size=3)
next(iter(dl))
```
```
~/work/fastai/fastai/torch_core.py in to_device(b, device)
    109     device = ifnone(device, defaults.device)
    110     if is_listy(b): return [to_device(o, device) for o in b]
--> 111     return b.to(device)
    112 
    113 def data_collate(batch:ItemsList)->Tensor:
AttributeError: 'dict' object has no attribute 'to'
```

However, in this commit, the code raises no error.

# Why I create this PR

That's because pytorch's `DataLoader` supports dict based `Dataset`, i.e. the below code raises no error.
```python
dl=DataLoader(DictDataset(), batch_size=3)
next(iter(dl))
```

### DIFF
--- a/fastai/core.py
+++ b/fastai/core.py
@@ -50,6 +50,7 @@ defaults = SimpleNamespace(cpus=_default_cpus, cmap='viridis')
 
 def is_listy(x:Any)->bool: return isinstance(x, (tuple,list))
 def is_tuple(x:Any)->bool: return isinstance(x, tuple)
+def is_dict(x:Any)->bool: return isinstance(x, dict)
 def noop(x): return x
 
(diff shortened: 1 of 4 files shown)


jph00:
Thanks!

PR #1577, opened 2019-02-06
Replace OpenFileProcessor in TextDataBunch.from_folder

daleevans:
I'm fairly certain that removing OpenFileProcessor was a typo.  With that gone, it tries to read the directory list as the text, resulting in a language model dataset that contains just the path name.


sgugger:
Good catch, thanks!

PR #1581, opened 2019-02-06
lr_find and function api descrition

Benudek:
simple test for lr_find.

Suggesting a Documentation for test functions (which API do they test?)for clarity . Potentially to prep feature discussed here: https://forums.fast.ai/t/improving-expanding-functional-tests/32929/65?u=benudek

### DIFF
--- a/tests/test_train.py
+++ b/tests/test_train.py
@@ -12,7 +12,23 @@ def learn():
     learn = fake_learner(50,50)
     return learn
 
+def test_lr_find(learn):
+    "Tests API: lr_find"
+    wd=0.002
+    start_lr = 1e-06
+    num_it = 90
+    end_lr = 10
+    lr_find(learn=learn, start_lr=start_lr, end_lr=end_lr, num_it=num_it, stop_div=True, wd=wd)
+    assert(len(learn.recorder.moms) == len(learn.recorder.lrs))
+    assert(learn.recorder.lrs[0] == start_lr)
+    assert(learn.recorder.moms[0] == 0.9)
+    assert(learn.recorder.lrs[len(learn.recorder.lrs) - 1] < learn.recorder.opt.lr)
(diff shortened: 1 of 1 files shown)


tests/test_train.py
```diff
@@ -12,7 +12,23 @@ def learn():
     learn = fake_learner(50,50)
     return learn
 
+def test_lr_find(learn):
+    "Tests API: lr_find"
+    wd=0.002
```
sgugger: Make all those assignments in one line please (wd,start_lr,num_it,end_lr = ...)

tests/test_train.py
```diff
@@ -12,7 +12,23 @@ def learn():
...
+    start_lr = 1e-06
+    num_it = 90
+    end_lr = 10
+    lr_find(learn=learn, start_lr=start_lr, end_lr=end_lr, num_it=num_it, stop_div=True, wd=wd)
+    assert(len(learn.recorder.moms) == len(learn.recorder.lrs))
+    assert(learn.recorder.lrs[0] == start_lr)
+    assert(learn.recorder.moms[0] == 0.9)
+    assert(learn.recorder.lrs[len(learn.recorder.lrs) - 1] < learn.recorder.opt.lr)
```
sgugger: Just -1 is enough here.

sgugger:
Thanks, I just requested two minor changes before merging.

sgugger:
Thanks!

stas00:
a consistent syntax for assert() needs to be enforced - this PR uses assert(...), whereas we don't use parentheses anywhere else in the tests.

stas00:
also test function names are self-documenting and require no declaration unless it's not so, so let's not have unnecessary noise.

stas00:
@Benudek, perhaps as you receive feedback on your proposed tests you could start compiling guidelines for test writing? How does that sound?

PR #1582, opened 2019-02-06
Update model.py

arpitj07:
I have updated model.py because I was facing error while running lession-1 notebook on Kaggle.


sgugger:
This is more in the data that the problem lies. We can't add this line as there are situations where the target must be float.
Also please note that fastai 0.7 is non longer maintained.

PR #1584, opened 2019-02-06
add sep option to LanguageLearner.predict

daleevans:
For character based models, it's much easier to read the output if you can specify `sep=''`.

### DIFF
--- a/fastai/text/learner.py
+++ b/fastai/text/learner.py
@@ -11,7 +11,7 @@ from .models import get_language_model, get_rnn_classifier
 from .transform import *
 from .data import *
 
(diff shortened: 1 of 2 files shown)


sgugger:
Sounds like a good idea. Can you do the same for beam_search just below?

stas00:
given:

- `sep` (in ImageDataBunch factory methods) is now called `label_delim`

should this option be consistent?

sgugger:
This is different since it's not a label delimiter but a real character separator here.

sgugger:
Perfect, thanks a lot!

PR #1585, opened 2019-02-07
fix split_bn func

dhpollack:
the `split_bn_bias` function currently returns a normal list and not a `ModuleList`.  Additionally, `Sequential` is used instead of `ModuleList`.  Tests pass.  The other approach would to just change the stated output of the function to `->List[ModuleList]` instead of `->ModuleList`.


sgugger:
Mmm the type annotations are the one that are wrong: all those objects are supposed to be list of nn.Modules, not module lists, so the previous implementation was the right one, it's the two type annotations that need to be changed.
Tests pass, but sadly they're not very extensive yet and this specific part isn't really tested yet.

PR #1593, opened 2019-02-08
Minor docs prose improvement

odysseus0:
Minor docs prose improvement to the [training overview](https://docs.fast.ai/training.html).


docs_src/training.ipynb
```diff
@@ -27,29 +27,29 @@
...
     "\n",
-    "[`basic_train`](/basic_train.html#basic_train) contains the definition of this [`Learner`](/basic_train.html#Learner) class along with the wrapper around pytorch optimizer that the library uses. It defines the basic training loop that is used each time you call the [`fit`](/basic_train.html#fit) function in fastai (or one of its variants). This training loop is kept to the minimum number of instructions, and most of its customization happens in [`Callback`](/callback.html#Callback) objects.\n",
+    "[`basic_train`](/basic_train.html#basic_train) defines this [`Learner`](/basic_train.html#Learner) class, along with the wrapper around the PyTorch optimizer that the library uses. It documents the basic training loop that is used each time you call the [`fit`](/basic_train.html#fit) function(or one of its variants) in fastai. A training loop method requires an intentionally minimal number of arguments; if you want to customize your training loop, you could include in the arguments an optional [`Callback`](/callback.html#Callback) object, which provides a rich number of customization options for the training loop.\n",
```
sgugger: I would leave "It defines the basic training loop" (not documents). The last sentence really needs to be "This training loop is kept to the minimum number of instructions, and most of its customization happens in [`Callback`](/callback.html#Callback) objects." "A training loop method requires an intentionally minimal number of arguments" isn't accurate and would be misleading. Then at the end you can add "you can include such Callbacks in the arguments as an optional parameter"

docs_src/training.ipynb
```diff
@@ -27,29 +27,29 @@
...
     "\n",
-    "[`callback`](/callback.html#callback) contains the definition of those, as well as the [`CallbackHandler`](/callback.html#CallbackHandler) that is responsible for the communication between the training loop and the [`Callback`](/callback.html#Callback) functions. It maintains a state dictionary to be able to provide to each [`Callback`](/callback.html#Callback) all the informations of the training loop, easily allowing any tweaks you could think of.\n",
+    "[`callback`](/callback.html#callback) defines these customization options and the [`CallbackHandler`](/callback.html#CallbackHandler) that is responsible for the communication between the training loop and the [`Callback`](/callback.html#Callback)'s methods. The [`CallbackHandler`](/callback.html#CallbackHandler) maintains a state dictionary able to provide each [`Callback`](/callback.html#Callback) object all the information of the training loop it belongs to, making any imaginable tweaks of the training loop within your reach.\n",
```
sgugger: Please remove "[`callback`](/callback.html#callback) defines these customization options" as it's not really true.  Just say it defines this class.
At the end, I'd put "of this training loop" (not the)

docs_src/training.ipynb
```diff
@@ -91,7 +91,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "The most important object for training models is [`Learner`](/basic_train.html#Learner), which needs to know, at minimum, what data to train with and what model to train."
+    "Objects of the [`Learner`](/basic_train.html#Learner) class plays a central role in training models; they needs to know, at the very minimum, what data to train with and what exact kind of model to train."
```
sgugger: There is a typo at plays (no s).

docs_src/training.ipynb
```diff
@@ -154,7 +154,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "To see how our training is going, we can request that it reports various [`metrics`](/metrics.html#metrics) after each epoch. You can pass it to the constructor, or set it later. Note that metrics are always calculated on the validation set."
+    "To see how our training is going, we can request that it reports various kinds of [`metrics`](/metrics.html#metrics) after each epoch. You can pass it to the constructor, or set it later. Note that metrics is always calculated on the validation set."
```
sgugger: It's really are and not is (since metrics is plural).

sgugger:
Thanks. It's not ideal with the jupyter notebook format, but this needs a few changes before merging if you can.

odysseus0:
Certainly, I will incorporate your comments in another commit. Many thanks for the detailed comments; these really help me understand the library better.

odysseus0:
@sgugger In terms of the current difficulty reviewing Jupyter Notebook on Github, this Github app called [ReviewNB](https://www.reviewnb.com) might help.

odysseus0:
[Jupytext](https://github.com/mwouts/jupytext) is another library that might be potentially useful in resolving the same problem.

stas00:
> @sgugger In terms of the current difficulty reviewing Jupyter Notebook on Github, this Github app called [ReviewNB](https://www.reviewnb.com) might help.

This app is not so useful as it could be, since you can't do it direclty from github, at least last time I tried it. nbdiff does that beautifully already. https://forums.fast.ai/t/jupyter-notebook-enhancements-tips-and-tricks/17064/16, either way you have to pull the PR locally to use it. github should really install nbdime and have it used internally.

odysseus0:
I see. So currently the best way to go is to pull the PR locally and use nbdiff inside the Jupyter Notebook, at least until Github installs nbdime and uses it by default.

@stas00 Do you find the Jupytext thing to be a potentially useful interim solution? Love to hear your words of wisdom on that matter.

stas00:
> I see. So currently the best way to go is to pull the PR locally and use nbdiff inside the Jupyter Notebook, at least until Github installs nbdime and uses it by default.

Probably users need to start asking github support to include it - otherwise it won't happen, why would MS bother to provide an enhancement for just a few millions of jupyter notebooks hosted on github. It has a bigger fish to fry.

> @stas00 Do you find the Jupytext thing to be a potentially useful interim solution? Love to hear your words of wisdom on that matter.

I'm yet to try Jupytext, but from the description of it, it solves some problems but introduces other problems, e.g. editing the same content from 2 different places (jupyter and editor) is surely not going to go smoothly. And then it adds even more complexity for users to figure out which files to edit and how to check they did it right. And if you want to have the final rendering of the notebook with outputs in github (.e.g as we have them for the lesson notebooks), then you have to source control the jupytext source and the synced .ipynb, and telling users not to edit .ipynb (kind of what we already with autogenerated .html). All that is doable, but overall IMHO it doesn't make things easier.

I think the best solution would be for github to make an effort to better support jupyter notebooks by providing custom tools for diffing, commenting, etc. But it probably will take a lot of support requests by users.

odysseus0:
I see. Jupytext solves the diff problem on Github but adds another layer of complexities that might make users and future contributors' lives even harder. Totally with you. For now, I will try to get familiar with the workaround workflow.

docs_src/training.ipynb
odysseus0: Thanks.

docs_src/training.ipynb
odysseus0: Thanks.

docs_src/training.ipynb
odysseus0: Thanks!

docs_src/training.ipynb
odysseus0: Thanks for the detailed reply. I get the last part but could you please elaborate a bit on "This training loop is kept to the minimum number of instructions"? Do you mean that the training loop is very bare bone and generic? Or do you intend to mean that `fit` takes very few arguments to call? Or maybe both? Sorry for the back-and-forth but I really want to make sure that I understand exactly what you mean.

odysseus0:
@sgugger Hey Sylvain, I just updated the `train` notebook according to your feedback. There is one remaining question and it is asked in the comment on the specific block.

I also started working on adding content to the `basic_train` notebook. There is one specific thing that I am not sure how to get to work under the current build workflow of the notebook,  but still, I believe it to be a good feature to have on the documentation.

Currently, I believe that we rebuild the notebooks and the website after each modification from the contributors. So the exact outputs after the rebuild might be different from the output I see when running the notebook locally. However, there are situations where explaining the exact output might be helpful when writing docs. Here is an example:

image

Personally I believe that being able to point to the concrete exact number in the output makes the explanation more straightforward.

However, I am not sure how to do it. Is there a way to make the entire code execution deterministic, so that I can rest assured that the final outoput after the notebook rebuild will be the same as the one I see in my local computer? Or maybe there is a way to embed Python code inside the Markdown cell, so that I can simply refer to the variable? Need your wisdom.

sgugger:
In your specific example, I'd leave it interactive: just mentions that the first number is the probability of being a 3 and the second the probability of being a 7, then let's check if it's right or wrong, without using the actual probabilities or the actual class.
In general, any code we don't want to be run should be embedded in a Mardown cell.

docs_src/training.ipynb
sgugger: Yes, I mean the training loop has very few lines of code. The number of arguments isn't what Callbacks are for, they are there to avoid having thousands of lines of code in the basic training loop.

odysseus0: Ah, I see. Thank you again for the explanation!

odysseus0:
Hey @sgugger, I just updated the notebook according to your requested changes. Could you please review it again?

sgugger:
I can't merge the 14 conflicts on basic_train.html and basic_train.ipynb. Can you pull them from master again?

odysseus0:
It turned out that somehow I forgot to run `tools/run-after-git-clone` in my VM. That is why two commits failed the CI. I simply followed the instruction here in the [docs](https://docs.fast.ai/dev/develop.html#stripping-out-jupyter-notebooks) and it fixed my problem.

sgugger:
Thanks!

PR #1594, opened 2019-02-08
Added axis labels for Recorder.plot_lr.

odysseus0:
Added axis labels for `Recorder.plot_lr` according to the style used by Sylvain Gugger in his [blog post](https://sgugger.github.io/the-1cycle-policy.html). The improvement makes the output of `Recorder.plot_lr(show_moms=True)` more understandable for new users.

It used to look like this:
unknown

Now it looks like this:
unknown


sgugger:
Thanks!

PR #1595, opened 2019-02-08
Added try except AttributeError

us:
Added to "'NoneType' object has no attribute 'group'" error.

### DIFF
--- a/fastai/data_block.py
+++ b/fastai/data_block.py
@@ -266,7 +266,8 @@ class ItemList():
             s = str((os.path.join(self.path,o) if full_path else o).as_posix())
             res = pat.search(s)
             assert res,f'Failed to find "{pat}" in "{s}"'
-            return res.group(1)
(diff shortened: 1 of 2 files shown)


sgugger:
Good point. Maybe we can use the same assert as [here](https://github.com/fastai/fastai/blob/8a54d18ab21f9f60473a39927dd9c3b3397e8e95/fastai/data_block.py#L268) to make the behavior consistent?

us:
@sgugger thank you  I fixed it.

sgugger:
Thanks!

sgugger:
Oops, I misunderstood what you were trying to achieve. We don't want to go with the error being silent here, since it'll make things fail later on. We want the user to clearly know the pattern wasn't found in res so they can adjust their regex (or use another labelling method, or get rid of the samples that don't work).
I reverted.

PR #1597, opened 2019-02-08
Adding optional plot and return of min gradient 

bfarzin:
Attempt to return a suggestion (optionally) from `lr_find`
Also updated LR_find notebook to show usage

Happy to correct, change or add more examples if needed.  Just let me know!


stas00:
I like the printing part under the plot, but not sure whether plot() should return any values, since it's a plot function. perhaps let's just start with just printing it out? and making it the default? if we don't return anything do we even need the option?

Also, returning "random" values would make things complicated down the road if we have other arguments we want to return. For getting the min gradient perhaps there should be an accessor instead? what do you think @sgugger?

bfarzin:
thanks for the note about .html files.  I will update just the .ipynb

I agree that returning a value from a `.plot()` call seems strange. If it is just printing/plotting, then making it default seems natural.  

I can make it a property of the `Recorder` and then it can be used if you wish.  

I will make those updates now and await any other suggestions.  Happy to make more changes.

fastai/basic_train.py
```diff
@@ -382,7 +382,10 @@ def __init__(self, learn:Learner):
         self.opt = self.learn.opt
         self.train_dl = self.learn.data.train_dl
         self.no_val,self.silent = False,False
+        self.min_g_lr = None
```
sgugger: No need to define this here, just define it inside LR Finder.

fastai/basic_train.py
```diff
@@ -382,7 +382,10 @@ def __init__(self, learn:Learner):
...
 
+    def min_g_lr(self): return self.min_g_lr
```
sgugger: An attribute is enough, no need to make a method out of it.

fastai/basic_train.py
```diff
@@ -454,7 +457,11 @@ def plot(self, skip_start:int=10, skip_end:int=5)->None:
         ax.set_xlabel("Learning Rate")
         ax.set_xscale('log')
         ax.xaxis.set_major_formatter(plt.FormatStrFormatter('%.0e'))
-
+        mg = (np.gradient(np.array([x.item() for x in losses]))).argmin()
+        print(f"Min numerical gradient: {lrs[mg]:.2E}")
+        ax.plot(lrs[mg],losses[mg],markersize=10,marker='o',color='red')
+        self.min_g_lr = lrs[mg]
```
sgugger: I think `self.min_grad_lr` would be a better name.

sgugger:
Thanks! It looks like a useful feature, just requested a few changes.

bfarzin:
Thanks for the feedback!  Very easy changes to make.

sgugger:
Thanks!

emaror:
Small question regarding the `np.gradient()`. 

When you pass just the function values (in this case `self.losses`), `np.gradient()` assumes that these are equally spaced. With lr_find(), that is not the case. Shouldn't we also pass `self.lrs`, i.e. `np.gradient(np.array([x.item() for x in losses]),lrs)` ?

sgugger:
No, we want to take these gradients with respect to the log scale of the learning rates.

emaror:
I see.Thanks!

PR #1598, opened 2019-02-09
Expanding initial beam search input to work with transformerxl

bearpelican:
Get a invalid tensor dimension when running beam search with transformerxl.

Seems like it's because the memory 0th dimension is a size of 1 on the first word and a size of top_k from thereafter.


sgugger:
I don't know why you had an error, but this works too.

PR #1601, opened 2019-02-09
Fix environment.yml

ARMATAV:
This fixes the environment for newer people installing from the environment.yml file.


stas00:
Thank you, @ARMATAV

PR #1603, opened 2019-02-10
fix to extra row being generated by plot_top_losses. 

sjdlloyd:
Before it was adding an extra row for some values of k. For example k = 10, there would be an extra row.
Tested this manually for a number of desired grid sizes.
It's a pretty minor bug, but was also a simple fix.


sgugger:
Thanks!

PR #1605, opened 2019-02-10
Fix exceptions in text tokenization

PiotrCzapla:
The tokenization was joining empty lines together and was throwing an exception on an example that ends with whitespace.


sgugger:
Thanks!

PR #1607, opened 2019-02-10
Added heatmap boolean variable to plot_top_losses. By default this va…

at110:
Added heatmap boolean variable to plot_top_losses. By default this variable is True.

When true, plot_top_losses will overlay heat-maps on the top of images. Otherwise, plot_top_losses will display only images associated with top losses.

I am not sure how to write a test case. 
But here is two scenarios in which test worked well with and without my code.
(I assumed that passing of test is equivalent of displaying images associated with top losses).
I am looking forward to   learn more about it.

#####with my code 
path = untar_data(URLs.PETS); 
path_anno = path/'annotations'
path_img = path/'images'
np.random.seed(2)
pat = re.compile(r'/([^/]+)_\d+.jpg$')
data = ImageDataBunch.from_name_re(path_img, fnames, pat, ds_tfms=get_transforms(), size=224, bs=bs).normalize(imagenet_stats)
learn = create_cnn(data, models.resnet34, metrics=error_rate)
interp = ClassificationInterpretation.from_learner(learn)
losses,idxs = interp.top_losses()
len(data.valid_ds)==len(losses)==len(idxs)
interp.plot_top_losses(9, figsize=(15,11),heatmap=True)

###without my code
path = untar_data(URLs.PETS); 
path_anno = path/'annotations'
path_img = path/'images'
np.random.seed(2)
pat = re.compile(r'/([^/]+)_\d+.jpg$')
data = ImageDataBunch.from_name_re(path_img, fnames, pat, ds_tfms=get_transforms(), size=224, bs=bs).normalize(imagenet_stats)
learn = create_cnn(data, models.resnet34, metrics=error_rate)
interp = ClassificationInterpretation.from_learner(learn)
losses,idxs = interp.top_losses()
len(data.valid_ds)==len(losses)==len(idxs)
interp.plot_top_losses(9, figsize=(15,11))

### DIFF
--- a/fastai/train.py
+++ b/fastai/train.py
@@ -100,12 +100,13 @@ Learner.clip_grad = clip_grad
 
 class ClassificationInterpretation():
     "Interpretation methods for classification models."
-    def __init__(self, data:DataBunch, probs:Tensor, y_true:Tensor, losses:Tensor, ds_type:DatasetType=DatasetType.Valid):
-        self.data,self.probs,self.y_true,self.losses,self.ds_type = data,probs,y_true,losses,ds_type
+    def __init__(self, data:DataBunch, learn:Learner, probs:Tensor, y_true:Tensor, losses:Tensor, ds_type:DatasetType=DatasetType.Valid):
+        self.data,self.probs,self.y_true,self.losses,self.ds_type, self.learn= data,probs,y_true,losses,ds_type,learn
         self.pred_class = self.probs.argmax(dim=1)
+        
 
     @classmethod
-    def from_learner(cls, learn:Learner, ds_type:DatasetType=DatasetType.Valid):
+    def from_learner(cls, learn: Learner,  ds_type:DatasetType=DatasetType.Valid):
         "Create an instance of `ClassificationInterpretation`"
         preds = learn.get_preds(ds_type=ds_type, with_loss=True)
         return cls(learn.data, *preds)
--- a/fastai/train.py
+++ b/fastai/train.py
@@ -109,7 +109,7 @@ class ClassificationInterpretation():
     def from_learner(cls, learn: Learner,  ds_type:DatasetType=DatasetType.Valid):
         "Create an instance of `ClassificationInterpretation`"
         preds = learn.get_preds(ds_type=ds_type, with_loss=True)
-        return cls(learn.data, *preds)
+        return cls(learn, *preds)
 
     def confusion_matrix(self, slice_size:int=1):
         "Confusion matrix as an `np.ndarray`."
--- a/fastai/train.py
+++ b/fastai/train.py
@@ -100,8 +100,8 @@ Learner.clip_grad = clip_grad
 
 class ClassificationInterpretation():
     "Interpretation methods for classification models."
-    def __init__(self, data:DataBunch, learn:Learner, probs:Tensor, y_true:Tensor, losses:Tensor, ds_type:DatasetType=DatasetType.Valid):
-        self.data,self.probs,self.y_true,self.losses,self.ds_type, self.learn= data,probs,y_true,losses,ds_type,learn
+    def __init__(self, learn:Learner, probs:Tensor, y_true:Tensor, losses:Tensor, ds_type:DatasetType=DatasetType.Valid):
+        self.data,self.probs,self.y_true,self.losses,self.ds_type, self.learn= learn.data,probs,y_true,losses,ds_type,learn
         self.pred_class = self.probs.argmax(dim=1)
         
 
--- a/fastai/vision/learner.py
+++ b/fastai/vision/learner.py
@@ -6,7 +6,7 @@ from .image import *
 from . import models
 from ..callback import *
 from ..layers import *
-from ..callbacks.hooks import num_features_model
+from ..callbacks.hooks import *
 from ..train import ClassificationInterpretation
 
 __all__ = ['create_cnn', 'create_body', 'create_head', 'unet_learner']
@@ -100,9 +100,9 @@ def unet_learner(data:DataBunch, arch:Callable, pretrained:bool=True, blur_final
 def _cl_int_from_learner(cls, learn:Learner, ds_type:DatasetType=DatasetType.Valid, tta=False):
     "Create an instance of `ClassificationInterpretation`. `tta` indicates if we want to use Test Time Augmentation."
     preds = learn.TTA(ds_type=ds_type,with_loss=True) if tta else learn.get_preds(ds_type=ds_type, with_loss=True)
-    return cls(learn.data, *preds, ds_type=ds_type)
+    return cls(learn.data,learn, *preds, ds_type=ds_type)
 
-def _cl_int_plot_top_losses(self, k, largest=True, figsize=(12,12)):
+def _cl_int_plot_top_losses(self, k, largest=True, figsize=(12,12),heatmap:bool=True):
     "Show images in `top_losses` along with their prediction, actual, loss, and probability of predicted class."
     tl_val,tl_idx = self.top_losses(k,largest)
     classes = self.data.classes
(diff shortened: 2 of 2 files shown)


fastai/train.py
```diff
@@ -100,12 +100,13 @@ def clip_grad(learn:Learner, clip:float=0.1)->Learner:
 
 class ClassificationInterpretation():
     "Interpretation methods for classification models."
-    def __init__(self, data:DataBunch, probs:Tensor, y_true:Tensor, losses:Tensor, ds_type:DatasetType=DatasetType.Valid):
-        self.data,self.probs,self.y_true,self.losses,self.ds_type = data,probs,y_true,losses,ds_type
+    def __init__(self, data:DataBunch, learn:Learner, probs:Tensor, y_true:Tensor, losses:Tensor, ds_type:DatasetType=DatasetType.Valid):
```
sgugger: So we need to had the `Learner` for this to work, which is fine. Let's remove `DataBunch` then since we can access it with `learn.data`. This needs a bit of adjustment after (and in the `from_learner` method, don't forget to pass this `Learner`), so let me know if you'd rather have me doing it.

sgugger:
Thanks a lot! I added a comment are there is a little bit to refactor.

at110:
@sgugger Thank you.I agree with your suggestion about refactoring. I will fix it. I want to do it by myself because submitting PR and code refactoring is very good learning experience.

Thanks a lot!

fastai/train.py
```diff
@@ -100,12 +100,13 @@ def clip_grad(learn:Learner, clip:float=0.1)->Learner:
...
+        self.data,self.probs,self.y_true,self.losses,self.ds_type, self.learn= learn.data,probs,y_true,losses,ds_type,learn
         self.pred_class = self.probs.argmax(dim=1)
+        
 
     @classmethod
-    def from_learner(cls, learn:Learner, ds_type:DatasetType=DatasetType.Valid):
+    def from_learner(cls, learn: Learner,  ds_type:DatasetType=DatasetType.Valid):
         "Create an instance of `ClassificationInterpretation`"
         preds = learn.get_preds(ds_type=ds_type, with_loss=True)
         return cls(learn.data, *preds)
```
sgugger: Be careful, you need to pass the `Learner` here., not the data.

fastai/vision/learner.py
```diff
@@ -114,6 +114,18 @@ def _cl_int_plot_top_losses(self, k, largest=True, figsize=(12,12)):
         cl = int(cl)
         im.show(ax=axes.flat[i], title=
             f'{classes[self.pred_class[idx]]}/{classes[cl]} / {self.losses[idx]:.2f} / {self.probs[idx][cl]:.2f}')
+        if(heatmap):
```
sgugger: No parenthesis, just heatmap (style guide).

fastai/vision/learner.py
```diff
@@ -114,6 +114,18 @@ def _cl_int_plot_top_losses(self, k, largest=True, figsize=(12,12)):
...
+            xb,_ = self.data.one_item(im)
+            xb = xb.cuda()
```
sgugger: This is dangerous as you may break environments where people are on cpu only.  On the line above, you need to pass `detach=False, denorm=False` and xb won't be put on the CPU (on won't be denormalized, which is also needed), then this line won't be necessary anymore.

fastai/vision/learner.py
```diff
@@ -114,6 +114,18 @@ def _cl_int_plot_top_losses(self, k, largest=True, figsize=(12,12)):
...
+            m = self.learn.model.eval()
+            with hook_output(m[0]) as hook_a:
+                with hook_output(m[0], grad= True) as hook_g:
```
sgugger: Are you using `hook_g`?

sgugger:
Thanks for the changes! There is one typo you left, then I dug a little bit more in the code to actually create the heatmap and there are a few things to change there.

at110:
Thank you for feedback. I will work on these suggested changes as soon as possible.

fastai/train.py
at110: Thanks. This is fixed now.

fastai/vision/learner.py
at110: This is fixed.

fastai/vision/learner.py
at110: This is also fixe now.

fastai/vision/learner.py
at110: I must confess that I did not understand code for heat-map in details. I just used Jeremy's code from class as it is. But now I looked at  Lesson 6: pets revisited, I started to understand use of hook_g. I think grad-cam is  more informative than heat-map.  When I used grad-cam with plot_loss (largest = True/Fasle), it gave information about why we got largest loss for some of the  validation images. 
plot_top_losses(largest=True)
interp plot_top_losses 9 figsize 15 11 largest true heatmap true
plot_top_losses(largest=False)
interp plot_top_losses 9 figsize 15 11 largest false heatmap true

at110:
I was reading grad-cam paper http://openaccess.thecvf.com/content_ICCV_2017/papers/Selvaraju_Grad-CAM_Visual_Explanations_ICCV_2017_paper.pdf

and they have applied relu to final computations
So do you think 

mult = F.relu(((acts*grad_chan[...,None,None])).mean(0))

is more appropriate?

fastai/vision/learner.py
sgugger: Ok, using grad Cam makes more sense, yes!

sgugger:
Yes, please try and unless it gives horrible results, we'll use the approach of the paper.
Also, if you can then update the doc notebook (vision.learner) with this new argument heatmap and explains it shows the grad CAM with a link to the paper, that would be awesome. (This can be a separate PR.)

at110:
Definitely, I will like to contribute for documentation. As you suggested, I will create separate PR for updating doc notebook.

at110:
With RELU, visualization is  even better . 

There is one more change from paper, I am pushing.
mult = F.relu(((acts*grad_chan[...,None,None])).**mean**(0))
to
mult = F.relu(((acts*grad_chan[...,None,None])).**sum**(0))

sgugger:
Looks all good, thanks a lot for your help!

at110:
Thank you!

yeldarby:
This is a really neat feature! Unfortunately it looks like it's not quite working properly. I've been trying to debug why and it looks like sometimes `mult` is ending up as a tensor of size [1, 1] containing just `[[0.]]` so it's just putting a black overlay over the image.

The `lesson1-pets.ipynb` notebook can demonstrate the issue. On the `PETS` dataset the heatmap shows up (`mult` is of size [10, 10]).

If you go to the end of the notebook and do `plot_top_losses` on the  model created for the `MNIST_SAMPLE` dataset the heatmap doesn't work (`mult` is of size [1, 1]).

sgugger:
Yes, on MNIST, the images are too tiny and end up with just 1x1 feature after going through the pretrained model.

stas00:
Perhaps the feature should be activated only on a reasonable image size? e.g. at least 4x4 feature?

at110:
Yes. That makes sense. I liked idea. 
Does it need new PR?

stas00:
If @sgugger is in agreement, I'd say yes. 
We just need to decide on the threshold.

sgugger:
Yes maybe the default for heatmap should be None, then we set it properly depending on the size. 4 by 4 seems like a good idea so let's test for the numel of the features to be >= 16 (since images can be rectangular)

at110:
How about 3x3 or even 2x2?
I am struggling with kaggle Histopathology competition and they have image size of 96. I have feature matrix size of numberofChannelsx3x3 with resnet34 and resnet50. Having heat-map displayed with top losses was helpful. I guess same must be true for size of 64 as well.  
Even with 2x2, it will show very rough approximate location. 
If we turn it off below 4x4 then we may not be able use it with progressive resizing of image until we cross certain size threshold.


Also, @sgugger you caught one more issue. I did not anticipate rectangular images,
I will need to change following code as well 
```
 sz = im.shape[-1]
 axes.flat[i].imshow(mult, alpha =0.6, extent= (0,sz,sz,0), interpolation='bilinear', cmap='magma')

```

yeldarby:
Neat. I'm actually running a customized ResNet model so I was able to modify it to get it to work by changing `self.avgpool = nn.AdaptiveAvgPool2d((1, 1))` to `self.avgpool = nn.AdaptiveAvgPool2d((7, 7))`.

(Unfortunately making that change hurt the accuracy of the model quite a bit)

heatmap

stas00:
Perhaps the default should be on, with a very reasonable (high?) default threshold and also make the threshold configurable? Otherwise, if it's off by default, users will miss out on the feature that they may not even know exists?

at110:
Thank you @stas00 and @sgugger  for your suggestions. I submitted new PR for this change.

hbsmith:
Is it common that PRs without tests get merged into the master? `plot_top_losses` is the first example of diving into the docs when going through the fast.ai course, so perhaps it is not the best function to allow changes to without tests.

sgugger:
Well in an ideal world, we only have perfect PRs, with perfect bug-free code, prefect tests and perfect updates to the docs. This is an open-source project though, so people who make PRs do their best and often, we first merge that we find to be a cool new functionality even if it doesn't have tests or complete doc.
You are more than welcome to fix whatever is missing with another PR that would be deeply appreciated. Nonconstructive criticism is less so.

hbsmith:
@sgugger Sorry, I am definitely not trying to be non-constructive, and understand this is open-source. I was just trying to figure out why calling `doc(interp.plot_top_losses)` for me in the first lesson (through colab) was returning a "No tests found" above the docstring, and why the "Show in docs" hyperlink was taking me to github instead of the docs (maybe related to #1660 ?). The "contribute a test" also seems to point to a localhost. Will look into opening PRs for these issues. Thanks for quick response.

sgugger:
A lot of functions aren't tested yet, and the functionality that shows the tests is a new one, to highlight where help would be welcome ;)
If the link was broken, it might be an issue, so double-check and maybe file one so we can fix this.

at110:
I am also learning about unit testing . Sincere apologies if I have broken things unintentionally. If  new PR is not created for this issue then I am willing contribute for writing test for this one.

PR #1608, opened 2019-02-10
Add Dynamic Loss Scaling to MixedPrecision Callback

flpeters:
New Feature: Dynamic Loss Scaling when training in `fp16` (Mixed Precision).

Every `max_noskip` steps `loss_scale` is doubled, until it has reached `max_loss_scale`. If at any point during training, after the backward pass, any gradient is `inf`, `-inf` or `nan`, the weight update will be skipped and `loss_scale` is halved.
This prevents a model from being destroyed (the parameters turning `nan`), because updates that would cause this are never performed.
The code is heavily inspired by [NVIDIA amp](https://github.com/NVIDIA/apex/tree/master/apex/amp) which is licensed under the [BSD 3-Caluse License](https://github.com/NVIDIA/apex/blob/master/LICENSE). I did write everything myself, meaning no line of code was copied, but the idea on how to do it is entirely theirs.
Testing this feature can be done by training a model in fp16 and setting `max_noskip` to something low like 20. That way `loss_scale` will increase every 20 steps until the gradients overflow, at which point an update is skipped and `loss_scale` is reduced again.
I haven't changed [this line](https://github.com/fastai/fastai/blob/b935a205758456d3b1f0a665ba118b9d150614e6/fastai/train.py#L37) yet, because there are probably more intuitive names for the arguments than what I came up with.

Using this I was able to successfully train a model whose parameters had previously always turned `nan` at some point. 
capture

Todo:
- Add the new arguments to `to_fp16`
- If there are any other sources where the loss could possibly become `nan`, something where decreasing `loss_scale` doesn't help, then it will get stuck trying to decrease `loss_scale` as much as possible. A fix would be to keep track of how many times `loss_scale` was reduces in a row and to throw an error if it exceeds some number.

### DIFF
--- a/fastai/callbacks/fp16.py
+++ b/fastai/callbacks/fp16.py
@@ -55,9 +55,11 @@ def master2model(model_params:Sequence[Tensor], master_params:Sequence[Tensor], 
 
 class MixedPrecision(LearnerCallback):
     "Callback that handles mixed-precision training."
-    def __init__(self, learn:Learner, loss_scale:float=512., flat_master:bool=False):
+    def __init__(self, learn:Learner, loss_scale:float=512., max_noskip:int=2000, max_loss_scale:float=2.**24, flat_master:bool=False):
         super().__init__(learn)
         self.loss_scale,self.flat_master = loss_scale,flat_master
+        self.max_noskip, self.max_loss_scale = max_noskip, max_loss_scale
+        self.noskip = 0
         self.not_min += ['model_params', 'master_params']
         assert torch.backends.cudnn.enabled, "Mixed precision training requires cudnn."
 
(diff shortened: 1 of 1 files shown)


sgugger:
It's funny I was working on this specifically just yesterday! This looks great but I have a few things I'd like refactored and I also wanted to start high then diminish instead of rising the loss scale, providing the option to do a fix or dynamic loss scale.
Depending on the time you have to put on this, I can either merge and work from what you've done, or suggest changes in a code review, let me know what you prefer.

flpeters:
Thank you for your feedback! Since you were already working on this, you probably know exactly where to go from here so feel free to continue from where I left of. I'm glad I could be of some help :)

sgugger:
Noted. I'll merge and take over then, thanks a lot!

sgugger:
If you want to have a look at my changes, it's all in [this commit](https://github.com/fastai/fastai/commit/d8cdcc21c4edef9fc6369e232073b6c3ce3051dd).

Issue #1606, opened 2019-02-10
top_k_accuracy does not work with language models


juriwiens:
**Describe the bug**
The `top_k_accuracy` metric throws the following error when used with a language model.

I already have a fix for this problem. I'm going to write some tests this afternoon and then submit a PR.

```
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-4-b8d7093fffc8> in <module>
----> 1 learn.validate(metrics=[accuracy, top_k_accuracy])

/opt/anaconda3/lib/python3.7/site-packages/fastai/basic_train.py in validate(self, dl, callbacks, metrics)
    298         cb_handler = CallbackHandler(self.callbacks + ifnone(callbacks, []), metrics)
    299         cb_handler.on_epoch_begin()
--> 300         val_metrics = validate(self.model, dl, self.loss_func, cb_handler)
    301         cb_handler.on_epoch_end(val_metrics)
    302         return cb_handler.state_dict['last_metrics']

/opt/anaconda3/lib/python3.7/site-packages/fastai/basic_train.py in validate(model, dl, loss_func, cb_handler, pbar, average, n_batch)
     55             if not is_listy(yb): yb = [yb]
     56             nums.append(yb[0].shape[0])
---> 57             if cb_handler and cb_handler.on_batch_end(val_losses[-1]): break
     58             if n_batch and (len(nums)>=n_batch): break
     59         nums = np.array(nums, dtype=np.float32)

/opt/anaconda3/lib/python3.7/site-packages/fastai/callback.py in on_batch_end(self, loss)
    256         "Handle end of processing one batch with `loss`."
    257         self.state_dict['last_loss'] = loss
--> 258         stop = np.any(self('batch_end', not self.state_dict['train']))
    259         if self.state_dict['train']:
    260             self.state_dict['iteration'] += 1

/opt/anaconda3/lib/python3.7/site-packages/fastai/callback.py in __call__(self, cb_name, call_mets, **kwargs)
    197     def __call__(self, cb_name, call_mets=True, **kwargs)->None:
    198         "Call through to all of the `CallbakHandler` functions."
--> 199         if call_mets: [getattr(met, f'on_{cb_name}')(**self.state_dict, **kwargs) for met in self.metrics]
    200         return [getattr(cb, f'on_{cb_name}')(**self.state_dict, **kwargs) for cb in self.callbacks]
    201 

[...]

PR #1609, opened 2019-02-10
Fix top_k_accuracy

juriwiens:
The current top_k_accuracy implementation causes errors when used in the context of language models (#1606).

This PR adds:
- some contextual tests that allow the errors to be reproduced
- an alternative implementation that passes the tests

I know that the fastai style is to use very short abbrevations, please let me know if my names are fine / short enough.


sgugger:
Seems all great, thanks a lot!

PR #1614, opened 2019-02-11
Added instructions on local site testing for Mac OS users.

odysseus0:
Added instructions on local site testing for Mac OS users.


sgugger:
Thanks!

PR #1616, opened 2019-02-11
Added instructions for nbdime git integration. Other minor fixes.

odysseus0:
Current Docs Authoring guide on the documentation website does not include an instruction to setup `nbdime` git integration. As a result, when there is conflict in merging notebook changes, the notebook format will be broken and the post-merge git hook will throw exceptions. Here is what happened:

```bash
Auto-merging docs_src/training.ipynb
Auto-merging docs_src/basic_train.ipynb
CONFLICT (content): Merge conflict in docs_src/basic_train.ipynb
Auto-merging docs/training.html
Auto-merging docs/basic_train.html
CONFLICT (content): Merge conflict in docs/basic_train.html
  File "tools/fastai-nbstripout", line 74, in <module>
    s = json.load(input_stream)
  File "/Users/georgezhang/anaconda3/envs/fastai-dev/lib/python3.7/json/__init__.py", line 296, in load
    parse_constant=parse_constant, object_pairs_hook=object_pairs_hook, **kw)
  File "/Users/georgezhang/anaconda3/envs/fastai-dev/lib/python3.7/json/__init__.py", line 348, in loads
    return _default_decoder.decode(s)
  File "/Users/georgezhang/anaconda3/envs/fastai-dev/lib/python3.7/json/decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/Users/georgezhang/anaconda3/envs/fastai-dev/lib/python3.7/json/decoder.py", line 353, in raw_decode
    obj, end = self.scan_once(s, idx)
json.decoder.JSONDecodeError: Expecting property name enclosed in double quotes: line 422 column 1 (char 34860)
error: external filter 'tools/fastai-nbstripout -d' failed 1
error: external filter 'tools/fastai-nbstripout -d' failed
fatal: docs_src/basic_train.ipynb: clean filter 'fastai-nbstripout-docs' failed
```

I just added instruction to set up `nbdime` git integration in the guide. Also fixed some other minor markdown issues, primarily adding language annotation to the code block so that there will be code syntax highlight.


sgugger:
Thanks!

stas00:
Well, 2 things:

1. this approach is problematic since it overrides our custom git diff settings, so now `git diff` will be showing `nbdime diff` and not the diff we need. Not good. 

merge conflict shouldn't happen if you did the setup right and nobody committed an unstripped out notebook. and when things break we have: https://docs.fast.ai/dev/develop.html#unstripped-notebook-repair.

So we can have an nbdime installed but not act globally.

2. in the future please don't mix doc enhancements with "functional" changes, it makes it much more difficult to review PRs. So kindly stick to separating the two into separate PRs. Thank you.

stas00:
And about doc PRs in general:

1. in the future please don't mix doc enhancements with "functional" changes, it makes it much more difficult to review PRs. So kindly stick to separating the two into separate PRs. Thank you.

2. please don't delete the vertical new lines, it's much easier to tell sections apart when editing when there is some breath space. it doesn't impact the final rendering in html. Thanks.

odysseus0:
Hey @stas00, I just realized that I did not run `tools/run-after-git-clone` on my remote VM after cloning the repo. 

The reason why I did not is because of the following passage in the [Doc Maintenance](https://docs.fast.ai/gen_doc_main.html#step-2-setup) article.

image

Somehow it gave me the illusion that you have now embedded it as a hook and skipped running `tools/run-after-git-clone`. For some noob like me, I did not manage to catch exactly what exactly are the 'specfic instructions'. So when I tried to merge commmits from the remote VM, naturally the post-merge hook broke. My fault.

---

We definitely should continue our discussion on the forum.

> 
> May be paste this comment there and follow up to it?
> 

I am not sure exactly which part of the comment should go to the forum. Since you are the author, would you mind pasting the part you desire to the forum?

stas00:
I'm not quite following you - if you were to follow the link and run `./fastai-make-pr-branch` to create the PR branch, it runs `tools/run-after-git-clone` in the checked out branch - hence the note.

> I am not sure exactly which part of the comment should go to the forum. Since you are the author, would you mind pasting the part you desire to the forum?

I said that because I wasn't sure what we were trying to fix. But as you're clarifying there is nothing to fix, other than perhaps clarifying the docs.

You can always check whether you're set up by running:

```
$ tools/trust-origin-git-config -t
Executing: git config --list --show-origin
Check: repo's .gitconfig is trusted
```

to test it. Perhaps it's the easiest to just say to run `tools/run-after-git-clone` anyway?

odysseus0:
Let me clarify a bit. Sorry for the confusion.

1. The problem is on my part because I misunderstood the docs.

2. I definitely agree with you that just tell the users what to run would save all of us some troubles. More words leave more space for misunderstanding.

---

I did not realize that there is such a thing as `./fastai-make-pr-branch`. My bad. I just followed the [Doc Maintenance](https://docs.fast.ai/gen_doc_main.html) instructions, which lead me to [submit your pr](https://docs.fast.ai/dev/git.html#step-7-submit-your-pr) instruction; the latter did not mention running `./fastai-make-pr-branch`. For the previous few attempts where I did remember to run `tools/run-after-git-clone` after clone, there seems no issue.

I do have a suggestion. Could we separate the instructions to follow and the explanations of the details involved into separate documents? Maybe it is just me being noob, but sometimes I always get mixed when reading both together.

stas00:
But even if you don't use `./fastai-make-pr-branch`, and just followed the the Doc Maintenance instructions, they include `tools/run-after-git-clone`, so really all bases are covered. 

Perhaps what's is needed is to add a note that if you do get a merge conflict, check that you have your repo setup? since that's where your initial trouble started.

> I do have a suggestion. Could we separate the instructions to follow and the explanations of the details involved into separate documents? Maybe it is just me being a noob, but sometimes I always get mixed when reading both together.

I'm not sure how that would work, since you can't have explanations w/o instructions they explain. I think that perhaps what you needing is to keep everything as it is, but also add a section with just one block of bash commands to run.

We have something similar here: https://docs.fast.ai/dev/release.html#quick-release-process
There are just the commands, and then they are repeated again and explained.

Yes?

On the other hand `./fastai-make-pr-branch` is all of those commands in one block and you don't even need to know what they are. I suppose we can have both.

odysseus0:
> Perhaps what's is needed is to add a note that if you do get a merge conflict, check that you have your repo setup correctly?

Totally agree. Technically, the part about stripping notebook will do it. We just need to add some notes to point people with merge conflict to that section. If you are okay with it, I will add it.

> I'm not sure how that would work, since you can't have explanations w/o instructions they explain. I think that perhaps what you are needing is to keep everything as it is, but also add a section with just one block of bash commands to run.
>
> We have something similar here: https://docs.fast.ai/dev/release.html#quick-release-process
There are just the commands in that section w/o explanations, and then they are repeated again and explained.
>
> Yes?

Exactly. That is what I mean. Make the explanations as detailed as we like, but create a separate section of just-type-these-bash-commands-in-order guide would be perfect. 

> On the other hand `./fastai-make-pr-branch` is all of those commands in one block and you don't even need to know what they are. I suppose we can have both.

Oh, I am totally with you. Don't get me wrong. It looks wonderful now I know what it does. I just somehow missed it. We definitely should keep it!

stas00:
OK, I tried to do that, but making a PR branch is a very different situation from https://docs.fast.ai/dev/release.html#quick-release-process.  The latter is a straight line with no conditional branches, the former has many branches, plus it can't be copy-n-pasted since it requires user-specific entries (ssh vs https, username, branchname, etc.). So while I'm with you that a condensed list of commands would be very useful, but I fail to see how we could provide that in this case.

and, yes, please feel free to make any suggestions you feel would improve things. These are always very helpful from a user who is new, because the old timers often know too much to see what's  missing or not obvious. As you can tell some I'm happy to integrate some of those verbatim, others with changes, other don't quite fit and so we try to find a better way as this discussion shows. Slowly, slowly, things will get even better. 

Thank you for caring and brainstorming with me on how to improve our documentation, @odysseus0.

stas00:
@odysseus0, and as I mentioned earlier, it'd be helpful to add to git.md instructions on how to use nbmerge - in particular for users who edited their lessons and an update is breaking them. (w/o configuring nbdime to act globally). probably both, its normal method and the web one. I have only used it a few times so currently am not an expert.

and probably should migrate this entry as well https://forums.fast.ai/t/jupyter-notebook-enhancements-tips-and-tricks/17064/16 - I find that integration with jupyter notebook to be so helpful, especially with lesson notebooks. The trick is to configure it correctly to show you just the right bits as that section explains.

odysseus0:
@stas00 No problem. I will try to find a place in `git.md` to put down instructions about `nbdime`. Your forum post looks wonderful and I will certainly migrate it there.

odysseus0:
Hey @stas00, here is my understanding of what you intend, please let me know if you find any misunderstanding:

- For library and docs developers, only integrate the merge driver of nbdime locally as we customized diff outselves.
- For course notebook users and class takers, integrate both the merge and diff drivers, as there will be no customized diff.

stas00:
Thank you for asking for clarification.

I think let's start with first no integration in either situation. and just straightforward usage instructions.

For example, once nbdime is installed, you don't need to configure it to use it with git, because you can call it directly, e.g.  call `nbdiff` instead of `git diff`. we already have `git diff` using `fastai-nbstripout` for diff. 

sidenote: I like and dislike CLI `nbdiff` - like it for not dumping inlined images like normal diff does, but disliking that its diff shows cells separately so you can't really tell what's different. But it's web nbdiff is fantastic, especially if you configure it as I shared in that forum post.

The same can be done for `nbmerge`, but I have only used it once long time ago, so I don't remember how to use it. Perhaps, it's not practical trying to use it w/o integration, unlike `nbdiff`, and then the user has to integrate it - but only for git merging and not diff. Once you will have that knowledge then we can figure out the answer to your question. But the bottom line it shouldn't interfere with the setup we have - which is only diff and filters, but I'm not sure how merge is impacted by filters, we only have them on the way out (git push), so perhaps it's not a problem.

Also we need to figure out what does it do if the files in question aren't notebooks, does it delegate them to something else? Most likely it uses the same gitattributes to be only run for `*.ipynb` and doesn't get invoked otherwise? and then it has the CLI and the web interface. Given that web nbdiff is a way more user-friendly, perhaps the same applies for nbmerge. 

Finally, remember we can't integrate anything on behalf of users due to git's security, so we can only give them instructions on how to do it and they have to activate them - well we can prep .gitconfig file and they still need to activate it. which will lead to another problem that while fastai-nbstripout is in the repo it'll always work, and if the user doesn't have nbdime installed things would break badly if it's "integrated" - you will see what I mean once you start playing with it.

That's why my first sentence was - let's just give instructions on what one could do if they wanted to use nbmerge....

odysseus0:
Hey @stas00, thanks for the detailed reply. Here is what I will include in the instructions based on the feedback:

- Install `nbdime` out of the box with no `git` integration.
- Configure the `nbdime` setting according to your forum post and explains what it does.
- Briefly go over CLI `nbdiff`, then focus on `nbdiff-web`.
- Go over `nbmerge`, specifically the `nbmerge-web`.

Below are what I am going to figure out:

- Whether `nbmerge` is practical to use without git integration.
- Whether `nbmerge`, even without integration, break our setup.
- What does `nbmerge` do if the files in question aren't notebooks.

Below is what I can guarantee you:

- `nbmerge-web` is pretty fantastic, especially when used only for the source code like your forum post has shown

Below is what I unfortunately (again) failed to understand:

>Finally, remember we can't integrate anything on behalf of users due to git's security, so we can only give them instructions on how to do it and they have to activate them - well we can prep .gitconfig file and they still need to activate it. which will lead to another problem that while `fastai-nbstripout` is in the repo the integration rules will always work, and if the user doesn't have `nbdime` installed things would break badly if it's "integrated" - you will see what I mean once you start playing with it.

stas00:
It's most likely because we haven't discussed what both of us mean by "integration". Let's not worry about it for now - we will sort it out on the way - it's just docs, so we can improve those gradually as we find a common language.

The rest looks perfect.

And yes, I am sure nbmerge-web is fantastic, how can it not be when nbdiff-web is :) Thank you for your vote of confidence, @odysseus0

odysseus0:
Hey @stas00, here are what I have figured out so far:

* Whether `nbmerge` is practical to use without git integration.
  * It is really not. Unlike `nbdiff`, which in some sense looks for source control information automatically without git integration, `nbmerge` is oblivious of the source control context and would require three arguments to run:

   ```txt
  nbmerge: error: the following arguments are required: base, local, remote
   ```

   That is too much a hassle for the users.

  * However, `nbdime mergetool` is certainly practical to use even without git integration. Basically, it would show a nice web-based visualization of the merge conflicts, similar to `nbdiff`.

* Whether `nbdime mergetool`, even without integration, breaks our setup.
  * It does not from my limited experience. Last time when I broke our setup by running on `jupyter-lab`, I fixed the problem using `nbdime mergetool`.

* What does `nbmerge` do if the files in question aren't notebooks.
  * They simply ignore them and print out

PR #1617, opened 2019-02-11
Removed redundant information about stripping notebook.

odysseus0:
There are two places that mentioned solutions about fixing commit of unstripped-out notebooks. I removed the less complete one and replaced it by moving in the more complete one from an earlier part of the docs.

The rest are minor markdown improvements.

### DIFF
--- a/docs/dev/develop.md
+++ b/docs/dev/develop.md
@@ -16,7 +16,6 @@ Note: windows users, not using bash emulation, will need to invoke the command a
 
     python tools\run-after-git-clone
 
-
 ### after-git-clone #1: a mandatory notebook strip out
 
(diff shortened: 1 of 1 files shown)


sgugger:
@stas00 I'll let you check this one since you were the one to write those.

stas00:
Thank you for catching the overlap and proposing a better way, @odysseus0. I think new users will benefit from explicit instructions, so I merged the first section into the second and not the other way around, as your PR proposed, but otherwise the outcome is similar to what you suggested. there was also incorrect/oudated info in the sections which is now fixed.

Your style changes can't be easily merged, as I suggested in the other PR, please send such changes separately. Thank you.

PR #1619, opened 2019-02-11
Update transforms.py

vinotharjun:


stas00:

sgugger:
In any case, `import cv2` is done in `imports.py`, so the first line (`from .imports import *`) will give it to you.

PR #1621, opened 2019-02-12
Updated Numericalize

ouimet51:
Previously, the if statement in numericalize would check if the column type is a NOT a numeric type, this allowed us to still pass through a datetime dtypes and attempt to run categorical methods on those datetypes resulting in an assertion error. Using the is_categorical_dtype method allows us to explicitly only pass in columns that are categorical.


sgugger:
Please sign the [Contributor Licence Agreement](https://www.clahub.com/agreements/fastai/fastai) before we can merge your PR, thanks!

ouimet51:
> Please sign the [Contributor Licence Agreement](https://www.clahub.com/agreements/fastai/fastai) before we can merge your PR, thanks!

This is done :)

sgugger:
Thanks!

PR #1625, opened 2019-02-12
update beam_search

bfarzin:
From my reading, `beam_search` seemed to be mixing log probs and probs of each step. 
I moved it over to use log_probs everywhere and confirmed I get the same output.
I would like to add a test for this, but kind of difficult since it depends on the entire arch working, state, etc. Any ideas about how to do that?  I do want beam search to have some kind of test coverage.

Before the change, I would get this error when trying to beam search (complete example):
```
from fastai import *
from fastai.text import *

path = untar_data(URLs.IMDB_SAMPLE)
data = TextLMDataBunch.from_csv(path, 'texts.csv')
learn = language_model_learner(data, AWD_LSTM, drop_mult=0.5,pretrained=True)
defaults.device = torch.device('cpu')
learn.beam_search("this is a test",4,top_k=10,beam_sz=10)

<ipython-input-17-02df7b44939e> in <module>
      1 ## something not right with this implementation  get's negative probs.
----> 2 learn.beam_search("this is a test",4,top_k=10,beam_sz=10)

~/fast_ai/fastai-fork/fastai/text/learner.py in beam_search(self, text, n_words, top_k, beam_sz, temperature, sep, decoder)
    151                 xb = nodes[:,-1][:,None]
    152         if temperature != 1.: scores.div_(temperature)
--> 153         node_idx = torch.multinomial(1-scores, 1).item()
    154         return text + sep + sep.join(decoder(self.data.vocab.textify([i.item() for i in nodes[node_idx]], sep=None)))
    155 

RuntimeError: invalid argument 2: invalid multinomial distribution (encountering probability entry < 0) at /pytorch/aten/src/TH/generic/THTensorRandom.cpp:298
```

I attempted to fix the output which was duplicating the input string and the xxbos indicator. 

Happy to adjust/revise as needed.

### DIFF
--- a/fastai/text/learner.py
+++ b/fastai/text/learner.py
@@ -125,8 +125,8 @@ class LanguageLearner(RNNLearner):
             xb = xb.new_tensor([idx])[None]
         return text + sep + sep.join(decoder(self.data.vocab.textify(new_idx, sep=None)))
 
-    def beam_search(self, text:str, n_words:int, top_k:int=10, beam_sz:int=1000, temperature:float=1., sep:str=' ',
-                    decoder=decode_spec_tokens):
+    def beam_search(self, text:str, n_words:int, no_unk:bool=True, top_k:int=10, beam_sz:int=1000, temperature:float=1.,
+                    sep:str=' ', decoder=decode_spec_tokens):
         "Return the `n_words` that come after `text` using beam search."
         ds = self.data.single_dl.dataset
         self.model.reset()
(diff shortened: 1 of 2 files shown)


bfarzin:
I added a test in there.  I am not totally sure that it will be stable even though I am setting the numpy and torch random seed.  I am concerned that if weights are updated for the model, we will get a different prediction and the tests will all fail.

Open to any ideas about how to deal with this instability.  Should I mock the learner rather than using a pre-defined arch and weights?

sgugger:
You can't  test beam search like this for several reasons:
- we can't download the pretrained model in the tests because it takes too much time
- so it will never be deterministic, and you can't predict that will come out.

Good catch for the logs, I forgot to adapt the rest when I added this for numerical stability.

bfarzin:
Bummer. Ok, I will remove the test.

sgugger:
Good to go, thanks!

PR #1626, opened 2019-02-12
Toc parenthesis

bearpelican:
Documenting CLI for gen docs
No longer removing parenthesis from headers in TOC

@stas00


stas00:
Thank you, @bearpelican!

PR #1628, opened 2019-02-12
Don't assume categorical type. Explicit type conversion

waxter013:
Series.cat property is only available to Categorical data types, but is used where any type can be passed in. The code assumes that the data will be a Categorical type. As it stands now, passing in a DataFrame or Series will result in an error. Explicitly converting to a Categorical gets rid of this assumption.


sgugger:
Seems reasonable, thanks!

PR #1630, opened 2019-02-13
Keep git note updated with current progress.

odysseus0:
Now since `tools/fastai-make-pr-branch-py` is official and included in the repo, we can remove the words of precaution about its experimental status in the git notes.

### DIFF
--- a/docs/dev/git.md
+++ b/docs/dev/git.md
@@ -34,29 +34,18 @@ https://github.com/fastai/fastai/tree/master/fastai
 
 Below you will find detailed 5 steps towards creating a PR.
 
-### Helper Program
+### One-Click Solution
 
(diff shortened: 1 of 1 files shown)


stas00:
In general please don't change header names w/o first checking that some other documentation is using its anchor and there are a few of those in this case. And then there are the fastai forums where those links are kind of stuck forever, so ideally header names shouldn't be changes after they have been set for a while - totally ok to change a lot when it's new. 

And we are a way behind of broken anchors in the documentation, see:
https://dev.azure.com/fastdotai/fastai/_build?definitionId=5

The reason it's experimental is also because we are trying to port it to python:
https://forums.fast.ai/t/port-fastai-make-pr-branch-from-bash-to-python/29853/29
and your proposed change is a bit too soon.

odysseus0:
Gotcha. I will find some place to add these tips. Thanks again, @stas00.

PR #1631, opened 2019-02-13
Added more documentation for basic_train.

odysseus0:
Added more documentation for basic_train


odysseus0:
I am not sure if it was just my illusion, but when I was using fast.ai 1.0.43.1, somehow the `fit` method does not work anymore. I updated the library and somehow it works again.

To fix the output returned by the malfunctioning `fit` method, I rerun the notebook to update the outputs.

odysseus0:
I figured out why the second last and my previous commits failed the CI.

It turned out that the current Jupyter Lab is not compatible with the workflow we have setup for the notebooks. GCP uses Jupyter Lab by default for its Deep Learning VM, so I used it when rerunning the notebook remotely. It failed the CI again. So I rerun the notebook again with Jupyter Notebook. That fixed the `nbstripout_config` CI like a charm.

sgugger:
Looking good. There was a problem in v1.0.43 indeed that was fixed yesterday night.

PR #1632, opened 2019-02-13
Add warning not to use Jupyter Lab when contributing to the notebooks.

odysseus0:
Changes made in Jupyter Lab will break the current build workflow even after running `tools/run-after-git-clone`. Please refer to the change in the docs' markdown file and [another pull request](https://github.com/fastai/fastai/pull/1631#issuecomment-463167143) for more detail.

### DIFF
--- a/docs/gen_doc_main.md
+++ b/docs/gen_doc_main.md
@@ -6,6 +6,16 @@ title: Doc Maintenance
 
 Here is how you can contribute to the `fastai` documentation:
 
+### WARNING: Use Jupyter Notebook, NOT Jupyter Lab
+
+The current version of Jupyter Lab, as of February 16, 2019, is incompatible with our source control and build process of the notebooks. Changes from Jupyter Lab, when committed, will cause all sorts of problems, including
+
+- Failed CI test, specifically the `nbstripout_config` test. You will see a red cross next to your commit in Github.
+- Errors when merging the changes in another repository.
+- When you run `git status`, the notebook is always shown as modified. Even after running `git reset --hard` to reset your repository, it will not go away.
+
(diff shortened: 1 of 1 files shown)


stas00:
Thank you for this suggestion, @odysseus0. We already have a dedicated section for it, so I amended it with the warning: https://github.com/fastai/fastai/commit/ae7210515decda31530078d6ef72bc2da707656b
The many ways it breaks do not matter, there is no point listing them all, since none of them works.

I haven't experimented with jupyterlab, but most likely it generates a different json file, that is still compliant with nbformat, but is different from the one generated by jupyter notebook. Our `fastai-nbstripout` is written to bypass `nbformat` and to manipulate json format directly, which made it
about 10-15 times faster, than the same [nbstripout](https://github.com/kynan/nbstripout) which follows the rules and was too slow for our needs. So chances are that someone needs to look at the format generated by jupyterlab and adjust `fastai-nbstripout` to support that variation, and then most likely it should just work. If you'd like to have a look, `fastai-nbstripout` is a simple script that just traverses and manipulates the json file.

Gokkulnath:
@odysseus0 Can you please share more details on what problem you faced when running fastai-nbstripout  script on jupyterlab notebooks ?
 I am currently working on making fastai-nbstripout compatible with jupyter lab notebook as well. Can you provide a sample notebook where you had faced the issue.

odysseus0:
@Gokkulnath Ah, I believe that it is pretty general. Cannot recall exactly how it happened.

Here is my tip on how to recreate it:

- Clone two copies of your forked repositories.
- In cloned repo No. 1, make a change in a cell you choose in Jupyter Lab that will lead to output change. Commit and push it.
- In cloned repo No. 2, make a different change in the same cell you choose in Jupyter Notebook that will lead to a different output change. Commit.
- Now when you check `git status` in repo No. 2, you should be able to see something like that it has diverged from the remote.
- Then try to pull from remote and see what happens. What happened to me is that everything breaks apart.

PR #1637, opened 2019-02-14
Change the variable name in lr_find to match the corresponding one in

anirudhsundar:
Changed variable name in lr_find to make it match the calling function so that its easier to understand the code


sgugger:
Thanks!

PR #1638, opened 2019-02-14
Link checker installation guide for Mac OS users

odysseus0:
Added link checker installation guide for Mac OS users.

### DIFF
--- a/tools/checklink/README.md
+++ b/tools/checklink/README.md
@@ -9,15 +9,15 @@ This is https://github.com/w3c/link-checker with some custom tweaks.
    ```
    sudo apt install w3c-linkchecker
    ```
-
-   See "More on Prerequisites" below for other ways to install this prerequisite if the above doesn't work.
-
+   
 * Mac OS:
 
    ```bash
    brew install perl # Make sure that we have the most updated perl version installed
    perl -MCPAN -e 'install W3C::LinkChecker' # Install the link checker
    ```
+
+See "More on Prerequisites" below for other ways to install this prerequisite if the above doesn't work.
 
 To run this link checker again a local checkout, you also need to install `jekyll`:
 
--- a/tools/checklink/README.md
+++ b/tools/checklink/README.md
@@ -4,6 +4,8 @@ This is https://github.com/w3c/link-checker with some custom tweaks.
 
 ## Prerequisites
 
+### For Ubuntu/Debian Users
+
 Install:
 
 ```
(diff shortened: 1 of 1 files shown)


stas00:
This is goodness, but let's not create new headers unnecessarily, see how I changed the last similar PR of yours I believe to this approach:
https://github.com/fastai/fastai/blob/master/docs/gen_doc_main.md#testing-site-locally

and in the case of this PR these headers don't fit with the following header, because they are 1 or 2 and 3 - see what I mean? and the comment after the first section applies to both sections, and you left it out in the second - hence merging them is better.

stas00:
Actually, looking at this again:
https://github.com/fastai/fastai/blob/master/docs/gen_doc_main.md#testing-site-locally
does it not ask for a password on mac-os, like in the ubuntu entry?

odysseus0:
> Actually, looking at this again:
https://github.com/fastai/fastai/blob/master/docs/gen_doc_main.md#testing-site-locally
does it not ask for a password on mac-os?

Mac OS is a bit strange. As long as you didn't run `sudo` and only tries to install the package on your local user, it would not prompt for a password.

odysseus0:
> This is goodness, but let's not create new headers unnecessarily, see how I changed the last similar PR of yours I believe to this approach:

Gotcha. I thought that only changing header is not a good practice. If I understand you correctly, below is what I am going to do starting now:

- For minor change, add bullet points instead of new headers.

stas00:
> > This is goodness, but let's not create new headers unnecessarily, see how I changed the last similar PR of yours I believe to this approach:
> 
> Gotcha. I thought that only changing header is not a good practice. If I understand you correctly, below is what I am going to do starting now:
> 
>     * For minor change, add bullet points instead of new headers.

No, that's not it.

I guess I failed to convey what I was trying to say, that wasn't it. I was talking about the logical placement of content and headers, and your suggestion was conflicting with several bits of already existing text. 

Your suggestion was creating a situation where:

1. Choice A foo
2. Choice B foo
3. Unrelated topic bar

see how item 3 isn't fitting there? A more logical way is to:

1. foo
  * choice A
  * choice B
2. bar

--------------

Your suggestion was taking:

1. Choice A foo
   ...
  and also see BAR
   ...
BAR

and turning it into:

   1. Choice A foo
      ...
      and also see BAR
   2. Choice B foo
      ...

BAR

do you see that 'and also see BAR' is missing from item 2, so instead of replicating it, we make bullets:

foo:
   * Choice A
      ...

   * Choice B
     ...

    and also see BAR

BAR

-----------

I hope that was a better demo of the logic I was talking about.

stas00:
Thank you, @odysseus0

odysseus0:
Another lesson for me: how to better organize the logic in markdown without creating unnecessary headers.

As always, thank you, @stas00!

stas00:
It's not really about markdown or headers, it's more about trying to look at words as data and try to organize them in logical way. and it's like the foundation of the grammar rules that they teach us in school, where to put or, where and and commas, etc. and first they teach you how to group parts of the sentence by similarity, opposition, sequence, etc. then md and headers are just the grammar equivalent. 

See you in the next PR.

PR #1639, opened 2019-02-14
support other csv delimiters for text

jtbr:
Adds support for non-comma delimiters in csv files, as well as for naming columns in csv files without headers. This is especially important for text, since commas are often found in text.
Documentation is also updated.


fastai/data_block.py
```diff
@@ -121,10 +121,11 @@ def from_df(cls, df:DataFrame, path:PathOrStr='.', cols:IntsOrStrs=0, processor:
         return res
 
     @classmethod
-    def from_csv(cls, path:PathOrStr, csv_name:str, cols:IntsOrStrs=0, header:str='infer', 
-                 processor:PreProcessors=None, **kwargs)->'ItemList':
-        "Create an `ItemList` in `path` from the inputs in the `cols` of `path/csv_name` opened with `header`."
-        df = pd.read_csv(Path(path)/csv_name, header=header)
+    def from_csv(cls, path:PathOrStr, csv_name:str, cols:IntsOrStrs=0, delimiter:str=None, header:str='infer', 
+                 col_names:Optional[Collection[str]]=None, processor:PreProcessors=None, **kwargs)->'ItemList':
```
sgugger: The `cols` argument can already either be a collection of strings that are the column names, so I don't want to add another one that would be redundant. You can check if `cols` is a list of strings with `is_listy(cols) and isinstance(cols[0],str)` for instance.

fastai/data_block.py
```diff
@@ -121,10 +121,11 @@ def from_df(cls, df:DataFrame, path:PathOrStr='.', cols:IntsOrStrs=0, processor:
...
+        """Create an `ItemList` in `path` from the inputs in the `cols` of `path/csv_name` opened with `header`.
```
sgugger: Let's leave it a the first line (fastai style guide) and the second line of this doc string should go in the doc notebook).

fastai/text/data.py
```diff
@@ -204,15 +204,15 @@ def from_df(cls, path:PathOrStr, train_df:DataFrame, valid_df:DataFrame, test_df
 
     @classmethod
     def from_csv(cls, path:PathOrStr, csv_name, valid_pct:float=0.2, test:Optional[str]=None,
-                 tokenizer:Tokenizer=None, vocab:Vocab=None, classes:Collection[str]=None, header = 'infer', text_cols:IntsOrStrs=1,
-                 label_cols:IntsOrStrs=0, label_delim:str=None, chunksize:int=10000, max_vocab:int=60000,
-                min_freq:int=2, mark_fields:bool=False, **kwargs) -> DataBunch:
+                 tokenizer:Tokenizer=None, vocab:Vocab=None, classes:Collection[str]=None, delimiter:str=None, header='infer',
+                 col_names:Optional[Collection[str]]=None, text_cols:IntsOrStrs=1, label_cols:IntsOrStrs=0, label_delim:str=None,
```
sgugger: No `col_names` here either
You should also change `ImageDataBunch.from_csv` to add the `delimiter` argument.

sgugger:
Thanks. There is one problem with this PR outlined in the comments I left and a few things to tweak, but it seems like a useful functionality.

fastai/data_block.py
jtbr: The case where col_names is useful is where the csv has no header, so pandas doesn't know the column names. I don't think cols can overcome this (of course you can work around it by using only indices, which might be enough if there aren't many columns). Your call. 

I'll work on the other changes you mention.

sgugger: Oh right, cols would only contain the names of the inputs columns, not the outputs. If there are no col names, though, you have to know in advance what indexes your inputs are, so let's use integers in cols in this case.
I just don't want to have yet another argument if it's not necessary (`delimiter` is useful however).

jtbr:
I've made the changes you requested.

sgugger:
Looking good, thanks a lot!

PR #1641, opened 2019-02-14
Update dataset URLs to datasets.py

pnlph:
Dataset URLs added to the fastai registry. Image Classification and NLP.


stas00:
Thank you, @pnlph!

PR #1643, opened 2019-02-14
Optionally plot values inside the confusion matrix

herrmann:
Hi. I'm adding a flag to optionally skip plotting the text inside the confusion matrix cells, because when there's a high number of classes it takes a while to render and it's usually not visible unless the figsize is too big. I'm not sure how to add an automated test for it, so I'm submitting the PR directy. Thanks in advance!


sgugger:
Sounds like a good idea, thanks!

PR #1647, opened 2019-02-15
Added a save method and load classmethod to Vocab Class

StatisticDean:
Since 1.0.44, saving TextDataBunch has been changed. As a result, the itos.pkl file containing the Vocab of the language model has to be saved on its own. This PR adds a save(self, path) method and a load(clas, path) classmethod that allows you to save or load a vocab directly from the vocab class as suggested by sgugger in https://forums.fast.ai/t/saving-vocab-in-1-0-44/38313.

### DIFF
--- a/fastai/text/transform.py
+++ b/fastai/text/transform.py
@@ -140,6 +140,11 @@ class Vocab():
         self.itos = state['itos']
         self.stoi = collections.defaultdict(int,{v:k for k,v in enumerate(self.itos)})
 
+    def save(self, path):
+        "Save the itos in path"
+        with open(path, 'wb') as f:
+            pickle.dump(self.itos, f)
+
     @classmethod
     def create(cls, tokens:Tokens, max_vocab:int, min_freq:int) -> 'Vocab':
         "Create a vocabulary from a set of `tokens`."
@@ -148,4 +153,11 @@ class Vocab():
         for o in reversed(defaults.text_spec_tok):
             if o in itos: itos.remove(o)
             itos.insert(0, o)
+        return cls(itos)
+    
+    @classmethod
+    def load(cls, path):
+        "Load the Vocab contained in path"
+        with open(path, 'rb') as f:
+            itos = pickle.load(f)
         return cls(itos)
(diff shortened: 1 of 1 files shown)


fastai/text/transform.py
```diff
@@ -140,6 +140,11 @@ def __setstate__(self, state:dict):
         self.itos = state['itos']
         self.stoi = collections.defaultdict(int,{v:k for k,v in enumerate(self.itos)})
 
+    def save(self, path):
+        "Save the itos in path"
+        with open(path, 'wb') as f:
```
sgugger: This can all go on the same line with `pickle.dump(self.itos, open(path,'wb'))`.

fastai/text/transform.py
```diff
@@ -140,6 +140,11 @@ def __setstate__(self, state:dict):
         self.itos = state['itos']
         self.stoi = collections.defaultdict(int,{v:k for k,v in enumerate(self.itos)})
 
+    def save(self, path):
+        "Save the itos in path"
```
sgugger: Maybe `self.itos` and path with backticks (fastai style guide).

fastai/text/transform.py
```diff
@@ -149,3 +154,10 @@ def create(cls, tokens:Tokens, max_vocab:int, min_freq:int) -> 'Vocab':
             if o in itos: itos.remove(o)
             itos.insert(0, o)
         return cls(itos)
+    
+    @classmethod
+    def load(cls, path):
+        "Load the Vocab contained in path"
```
sgugger: backticks for Vocab and path.

fastai/text/transform.py
```diff
@@ -149,3 +154,10 @@ def create(cls, tokens:Tokens, max_vocab:int, min_freq:int) -> 'Vocab':
...
+        with open(path, 'rb') as f:
```
sgugger: Same, as before, this can go all on the same line.

sgugger:
Thanks! Just a few cosmetic changes and we should be good to merge.

sgugger:
Perfect, thanks!

Issue #1644, opened 2019-02-14
Potential Statistical Bug: _dihedral(...) and _dihedral_affine(...) in fastai.vision


kechan:
**Describe the bug**
Potential statistics/probabilistic bug for data augmentation transforms of dihedral(...) and dihedral_affine(...). I said potential because I have yet to dig into how uniform distribution are being sampled in fastai. For all I know, it could be doing it weirdly?

**Provide your installation details**
```text
=== Software === 
python       : 3.7.1
fastai       : 1.0.42
fastprogress : 0.1.18
torch        : 1.0.0
torch cuda   : None / is **Not available** 

=== Hardware === 
No GPUs available 

=== Environment === 
platform     : Darwin-17.7.0-x86_64-i386-64bit
conda env    : Unknown
python       : /Users/kelvinchan/tmp/python3_env/deeplearning/bin/python3
sys.path     : 
/Users/kelvinchan/Documents/LearnFastai
/Users/kelvinchan/tmp/python3_env/deeplearning/lib/python37.zip
/Users/kelvinchan/tmp/python3_env/deeplearning/lib/python3.7
/Users/kelvinchan/tmp/python3_env/deeplearning/lib/python3.7/lib-dynload
/usr/local/Cellar/python/3.7.1/Frameworks/Python.framework/Versions/3.7/lib/python3.7
/Users/kelvinchan/tmp/python3_env/deeplearning/lib/python3.7/site-packages
/Users/kelvinchan/tmp/python3_env/deeplearning/lib/python3.7/site-packages/IPython/extensions
/Users/kelvinchan/.ipython
no supported gpus found on this system
```

**To Reproduce**
(1) Put this in your notebook
```
from fastai.vision import *
def get_ex(): return open_image('imgs/cat_example.jpg')
tfms = [dihedral_affine()]
[get_ex().apply_tfms(tfms, size=224, padding_mode='zeros').show(ax=ax) 
 for i, ax in enumerate(plt.subplots(rows, cols, figsize = (width, height))[1].flatten())
];
```

(2) 
In your local install (pip), edit this in fastai/vision/transform.py:

```
def _dihedral_affine(k:partial(uniform_int,0,8)):
    "Randomly flip `x` image based on `k`."
    print("k={}".format(k))          // add this line to report the sampled k
    x = -1 if k&1 else 1
```

[...]

PR #1651, opened 2019-02-16
Fixed _dihedral and _dihedral_affine uniform_int sampling

kechan:
Fixes in fastai/vision/transform.py

Refer to https://github.com/fastai/fastai/issues/1644 for more details. 

The fix will allow sampling between 0-7 (excl. 8) resulting in equal probability for all orientations (incl. no op).


sgugger:
Thanks!

PR #1665, opened 2019-02-18
Fixed empty classes attribute

ambernardino:
Added classes attribute to avoid break when using TextDataBunch.load


sgugger:
Note that `TextDataBunch.load` is deprecated now, but it could also break with `from_ids`

PR #1668, opened 2019-02-18
Spline Fit for Recorder.plot() and AccumulateStepper(LearnerCallback)

KeremTurgutlu:
**Recorder.plot()** 

Here is the discussion thread for this PR:

https://forums.fast.ai/t/autolrfinder/38331/22

So far this method seems to be working pretty well for fixing outlier problems with lr vs loss plot.
Although, probably it's best to see if there is enough need for this feature, so please thumbs up if you see a fit here. It doesn't have drastic changes in `.plot()` method code. Now it optionally takes argument `k` as the spline order. If no k if given original smoothen loss is plotted.

**AccumulateStepper(LearnerCallback)**

Here is the discussion thread for this PR:

https://forums.fast.ai/t/accumulating-gradients/33219/26

Further experiments can be made...

### DIFF
--- a/fastai/basic_train.py
+++ b/fastai/basic_train.py
@@ -469,18 +469,27 @@ class Recorder(LearnerCallback):
             axs[1].set_xlabel('Iterations')
             axs[1].set_ylabel('Momentum')
         else: plt.plot(iterations, self.lrs)
+    
+    @staticmethod
+    def smoothen_by_spline(xs, ys, **kwargs):
+        xs = np.arange(len(ys))
+        spl = scipy.interpolate.UnivariateSpline(xs, ys, **kwargs)
+        ys = spl(xs)
+        return ys
 
-    def plot(self, skip_start:int=10, skip_end:int=5)->None:
+    def plot(self, skip_start:int=10, skip_end:int=5, **kwargs)->None:
         "Plot learning rate and losses, trimmed between `skip_start` and `skip_end`. Optionally plot and return min gradient"
         lrs = self.lrs[skip_start:-skip_end] if skip_end > 0 else self.lrs[skip_start:]
         losses = self.losses[skip_start:-skip_end] if skip_end > 0 else self.losses[skip_start:]
+        losses = [x.item() for x in losses]
+        if 'k' in kwargs: losses = self.smoothen_by_spline(lrs, losses, **kwargs)
         _, ax = plt.subplots(1,1)
         ax.plot(lrs, losses)
         ax.set_ylabel("Loss")
         ax.set_xlabel("Learning Rate")
         ax.set_xscale('log')
         ax.xaxis.set_major_formatter(plt.FormatStrFormatter('%.0e'))
-        mg = (np.gradient(np.array([x.item() for x in losses]))).argmin()
+        mg = (np.gradient(np.array(losses))).argmin()
         print(f"Min numerical gradient: {lrs[mg]:.2E}")
         ax.plot(lrs[mg],losses[mg],markersize=10,marker='o',color='red')
         self.min_grad_lr = lrs[mg]
--- a/fastai/train.py
+++ b/fastai/train.py
@@ -5,7 +5,7 @@ from .basic_data import *
 from .basic_train import *
 
 __all__ = ['BnFreeze', 'GradientClipping', 'ShowGraph', 'ClassificationInterpretation', 'fit_one_cycle', 'lr_find', 'one_cycle_scheduler', 'to_fp16', 'to_fp32',
-           'mixup']
+           'mixup', 'AccumulateStepper', 'turn_on_accumulation', 'turn_off_accumulation']
 
 def one_cycle_scheduler(lr_max:float, **kwargs:Any)->OneCycleScheduler:
     "Instantiate a `OneCycleScheduler` with `lr_max`."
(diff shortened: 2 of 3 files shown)


fastai/basic_train.py
```diff
@@ -469,18 +469,27 @@ def plot_lr(self, show_moms=False)->None:
...
+        losses = [x.item() for x in losses]
+        if 'k' in kwargs: losses = self.smoothen_by_spline(lrs, losses, **kwargs)
         _, ax = plt.subplots(1,1)
         ax.plot(lrs, losses)
         ax.set_ylabel("Loss")
         ax.set_xlabel("Learning Rate")
         ax.set_xscale('log')
         ax.xaxis.set_major_formatter(plt.FormatStrFormatter('%.0e'))
-        mg = (np.gradient(np.array([x.item() for x in losses]))).argmin()
+        mg = (np.gradient(np.array(losses))).argmin()
```
sgugger: I think you need the `[x.item() for x in losses]` since the element of `losses` are rank-0 tensors.

fastai/train.py
```diff
@@ -94,8 +94,52 @@ def clip_grad(learn:Learner, clip:float=0.1)->Learner:
     "Add gradient clipping of `clip` during training."
     learn.callback_fns.append(partial(GradientClipping, clip=clip))
     return learn
-
 Learner.clip_grad = clip_grad
+     
+class AccumulateStepper(LearnerCallback):
+    """
```
sgugger: On one line with just " " (fastai style guide).

fastai/train.py
```diff
@@ -94,8 +94,52 @@ def clip_grad(learn:Learner, clip:float=0.1)->Learner:
...
+    Does accumlated step every nth step by accumulating gradients
+    """
+    def __init__(self, learn:Learner, n_step:int = 1):
+        super().__init__(learn)
+        self.n_step = n_step
+ 
+    def on_train_begin(self, **kwargs):
+        "check if loss is reduction"
+        if hasattr(self.loss_func, "reduction") and (self.loss_func.reduction == "mean"):
+             print("For better gradients consider 'reduction=sum'")
```
sgugger: Make that a warning instead.

fastai/train.py
```diff
@@ -94,8 +94,52 @@ def clip_grad(learn:Learner, clip:float=0.1)->Learner:
...
+        
+    def on_epoch_begin(self, **kwargs):
+        "init samples and batches, change optimizer"
+        self.acc_samples = 0
```
sgugger: The two init on one line bli,bla = 0,0 (fastai style guide).

fastai/train.py
```diff
@@ -94,8 +94,52 @@ def clip_grad(learn:Learner, clip:float=0.1)->Learner:
...
+        if (self.acc_batches % self.n_step) == 0:
+            for p in (self.learn.model.parameters()):
+                if p.requires_grad: p.grad.div_(self.acc_samples)
+            self.acc_samples = 0
+            return False
+        else: return True
+    
+    def on_step_end(self, **kwargs):
+        "zero gradients after stepping"
+        if (self.acc_batches % self.n_step) == 0: return False
```
sgugger: Just returning the result of the test is enough ;)

fastai/train.py
```diff
@@ -94,8 +94,52 @@ def clip_grad(learn:Learner, clip:float=0.1)->Learner:
...
+        else: return True
+    
+    def on_epoch_end(self, **kwargs):
+        "step the rest of the accumulated grads if not perfectly divisible"
+        for p in (self.learn.model.parameters()):
+                if p.requires_grad: p.grad.div_(self.acc_samples)
+        self.learn.opt.step()
```
sgugger: Maybe add a `drop_last` argument similar to the one in the pytorch `DataLoader`? If it's `True`, we won't do the step, just zeroing.

sgugger:
Thanks a lot! There are a few style change to make and I made a suggestion to add an argument, then we'll be able to merge this.

KeremTurgutlu:
Thanks for suggestions :) I will update the PR as soon as possible, probably tonight.

fastai/train.py
KeremTurgutlu: I didn't understand this part, can you elaborate please :)

fastai/basic_train.py
```diff
@@ -469,18 +469,27 @@ def plot_lr(self, show_moms=False)->None:
...
+        xs = np.arange(len(ys))
+        spl = scipy.interpolate.UnivariateSpline(xs, ys, **kwargs)
+        ys = spl(xs)
+        return ys
+
+    def plot(self, skip_start:int=10, skip_end:int=5, **kwargs)->None:
         "Plot learning rate and losses, trimmed between `skip_start` and `skip_end`. Optionally plot and return min gradient"
         lrs = self.lrs[skip_start:-skip_end] if skip_end > 0 else self.lrs[skip_start:]
         losses = self.losses[skip_start:-skip_end] if skip_end > 0 else self.losses[skip_start:]
+        losses = [x.item() for x in losses]
```
KeremTurgutlu: That's true that's why I am getting item here.

sgugger: Eh, how did I miss this line?

fastai/train.py
sgugger: No I was saying, `return (self.acc_batches % self.n_step) != 0` (we need `False` when we should do the step + zero, `True` if we skip)

Instead of testing if bool return False else return True, just return not book ;)

fastai/train.py
```diff
@@ -94,8 +94,49 @@ def clip_grad(learn:Learner, clip:float=0.1)->Learner:
...
     return learn
-
 Learner.clip_grad = clip_grad
+     
+class AccumulateStepper(LearnerCallback):
+    "Does accumlated step every nth step by accumulating gradients"
+
+    def __init__(self, learn:Learner, n_step:int = 1, drop_last:bool = False):
+        super().__init__(learn)
+        self.n_step = n_step
```
sgugger: Both on one line (fastai style guide)

sgugger:
Two changes (one cosmetic, one more important in on_step_end) and we'll be good to go, thanks again!

sgugger:
Thanks! Now we'll have to address that issue in BatchNorm.

PR #1673, opened 2019-02-19
Improvedoctest

Benudek:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -172,7 +172,7 @@ test-fast: ## run tests in parallel (requires pip install pytest-xdist)
 	pytest -n 3
 
 test-full: ## run all tests, including slow ones, print summary
-	pytest --runslow -ra
+	pytest --runslow -ra -regtestapidb
 
 test-cpu: ## run tests with the default python and CUDA_VISIBLE_DEVICES=""
 	CUDA_VISIBLE_DEVICES="" python setup.py --quiet test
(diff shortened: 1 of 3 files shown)


Benudek:
context: 
https://forums.fast.ai/t/doc-test-project/38344/155?u=benudek
https://forums.fast.ai/t/doc-test-project/38344/154?u=benudek

stas00:
I haven't tried running it yet, but looking at the code - looks excellent, @Benudek.

I'd ask to only change:
```
  apiTestsMap = dict()
    oneOrMoreTestsfailed = False
```
we don't use CamelCase other than for class names.

So apiTestsMap => api_tests_map
oneOrMoreTestsfailed => some_tests_failed (some = one or more)

and later once it's all happy and working, we would need to go over and use consistent naming for the map/db/file, as now it's a bit all over. But will worry about it later.

So please make these 2 case renames and this is good to good.

Thank you.

stas00:
And looking again why `RegisterTestsPerAPI.oneOrMoreTestsfailed`, when it can just be a local variable in `conftest.py`?

Benudek:
think RegisterTestsPerAPI.some_tests_failed could go to conftest - but needs to be a class var, since is called on every test.

I am actually fine to go through these little improvements like naming conventions, if we make a list.

Some small things I saw (and we can extend the list and I can correct)
-  I use inspect.currentframe while I seem to remember we want to just say currentframe and import inspect accordingly
- in this I use fastaimodule = re.match(r'^fastai\..*',fq_apiname) to exclude non fastai APIs that were registered. Since we really do not want that in the code, could in full_name_with_qualname just raise an exception if that match isnt given. So, we force developers to revisit their code.

stas00:
Ah, ok, I couldn't tell from looking at the diff. Will look closer once it's in. 

Let's do just FooBar to foo_bar, commit and then deal with the finetuning later.

PR #1674, opened 2019-02-19
Add **dl_kwargs to .databunch() which is passed to DataLoader()

xeTaiz:
Hi there!
As stated in the [forums](https://forums.fast.ai/t/comfortable-pin-memory-option/38599), I was looking for a comfortable way to set pin_memory in dataloaders. I added some `**dl_kwargs` to `Databunch.create()` and as suggested by @sgugger, I did the same for `TextClasDataBunch.create()` and `TextLMDataBunch.create()`.

Works nicely for me and tests do pass. I added a note `**dl_kwargs` are passed to `DataLoader()` to the correspondig docstrings, which should automatically add it to the docu htmls right?


sgugger:
Thanks a lot!

PR #1675, opened 2019-02-19
Fix replace caps error.

illcat:
Fix an error.


sgugger:
Not exactly: we don't want all the I to become TK_UP i for instance, it's not the same thing as when a person is writing in all caps. On the other hand I should be coded TK_MAJ i

illcat:
t = 'I'
'''
if t[0].isupper() and t[1:].islower(): res.append(TK_MAJ)
'''
t[1:]Array may be out of bounds.

PR #1677, opened 2019-02-19
Optionally run AWD-LSTM from pre-computed embeddings

herrmann:
This patch allows someone using AWD-LSTM to calculate the forward pass directly from the embedded input. The alternative would be copying-pasting the model and splitting the embedder and encoder, but that would complicate code maintenance and break compatibility with saved models. This is needed in case someone needs to calculate the sequential Jacobian (I'll send another patch for this) or re-use the embeddings somewhere else.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -24,7 +24,6 @@ of that change.
 ### New:
 
 - `Learner.destroy`: completely free up `learn`, leaving an empty shell (to replace `gc.collect` eye-sore)
-- `Learner.hibernate`: `learn.export` + `learn.destroy` shortcut method
 - support `pynvx` instead of `pynvml` on OSX
 
 ### Changed:
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -25,6 +25,7 @@ of that change.
 
 - `Learner.destroy`: completely free up `learn`, leaving an empty shell (to replace `gc.collect` eye-sore)
 - support `pynvx` instead of `pynvml` on OSX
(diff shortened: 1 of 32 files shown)


sgugger:
Mmm, can you try to see if setting 
```
model.encoder_dp = noop
```
doesn't solve your issue? I'd rather not add a layer of complexity to an already fairly complex model.

herrmann:
Not quite... for calculating the intrinsic attention (or input sensitivity, or part of the sequential Jacobian) I need to calculate the partial derivatives of the embedded sentence w.r.t. the particular classification's softmax score. To do that, I embed, detach and set require_grad to True, then forward the model up to classification, call backward() and then proceed calculating the intrinsic attention from the embedding's grad.

Do you think I should I copy this thread in the forums as well? Thanks in advance!

sgugger:
Yes please, I'd like to make sure more people would benefit from this.

stas00:
That doesn't look right. You probably need to rebase your fork.

stas00:
Also: Please sign this CLA agreement https://www.clahub.com/agreements/fastai/fastai as explained [here](https://github.com/fastai/fastai/blob/master/CONTRIBUTING.md) before we can proceed.  Thank you.

herrmann:
Oops, I messed up the pull request... I'll open another one with both changes (AWD-LSTM patch and the TextClassificationInterpretation class)

herrmann:
New pull request: https://github.com/fastai/fastai/pull/1688

PR #1682, opened 2019-02-19
Doctestimprovements

Benudek:
follow up https://github.com/fastai/fastai/pull/1673

var names adjusted (previous PR was ouy of sync)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -172,7 +172,7 @@ test-fast: ## run tests in parallel (requires pip install pytest-xdist)
 	pytest -n 3
 
 test-full: ## run all tests, including slow ones, print summary
-	pytest --runslow -ra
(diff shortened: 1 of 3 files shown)


Benudek:
Some potential improvements to follow up:

* RegisterTestsPerAPI.some_tests_failed could go to conftest - but needs to be a class var, since is called on every test.
* I use inspect.currentframe while I seem to remember we want to just say currentframe and import inspect accordingly
* in this I use fastaimodule = re.match(r'^fastai..*',fq_apiname) to exclude non fastai APIs that were registered. Since we really do not want that in the code, could in full_name_with_qualname just raise an exception if that match isnt given. So, we force developers to revisit their code.

stas00:
Thank you, @Benudek.

PR #1685, opened 2019-02-20
check for existing entry

Benudek:


stas00:
That fixed the dups, thanks @Benudek

PR #1689, opened 2019-02-20
Fix spacy tokenizer no longer instantiated at each prediction

VictorXunS:
Issue:
Each time a new prediction is done using the text Tokenizer, a new Spacy blank object is instantiated.

Fix:
The tokenizer object is instantiated once and we then use the cached version

Result:
Saves 35 ms of CPU prediction time on my laptop (i7-7700)


sgugger:
I wish we could, but we can't have that without breaking the multiprocessing part of the tokenizer. 
Also, please note you need to sign the CLA for us to merge your PR.

PR #1690, opened 2019-02-20
flag missing this_tests

Benudek:
warning doesn't spit out filename, only test function but check the log, immediately the line before gives the filename :-)

### DIFF
--- a/fastai/gen_doc/doctest.py
+++ b/fastai/gen_doc/doctest.py
@@ -15,6 +15,7 @@ class TestAPIRegistry:
     "Tests register which API they validate using this class."
     api_tests_map     = defaultdict(list)
     some_tests_failed = False
+    failed_test = None
 
     @staticmethod
     def this_tests(*funcs):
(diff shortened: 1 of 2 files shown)


Benudek:
After thought: we do this exercise only to throw a warning, when a test is not registered with this - should we not register the API name in the json db also and an entry to test registered with this_tests?

Note, there is a slight complication: In

### @pytest.hookimpl(tryfirst=True, hookwrapper=True)
### def pytest_runtest_makereport(item, call):

we just have the testmethodname as a string. I havent found an obvious way to get from that to the fullly qualified name, which we would need to enter this into the DB.

Actually, think the warning is sufficient ... entering all APIs even if not registered with this_tests would mean cluttering the json with irrelevant entries.

Benudek:
@bearpelican we are mostly done with this json.db. Pls call pytest -sv or make test_full to get the json.

stas00:
Almost right.

- except it's not `failed_test`, but `has_this_test`.
- you don't really need to assign anything to `has_this_test`, other than True/False.
- and we don't want 100s of warnings, but to collect all missing ones and report them at the end in one go.
- and for now this feature should be off by default since it'll interfere with dev. once we instrument the remainder of the tests then yes, we will turn it on by default - i.e. the check will always run, regardless of how pytest is run, to catch newly written tests w/o `this_tests` early on.

Benudek:
Imho: I would give the warning right after the test ran (so my 1st version) ]. Since we should have very little warnings soon, that should be fine and is simpler for an end report, see my 2 below points.

However, made a small PR following your suggestion. The var failed_test was left over from * failed_tests where I was collecting all tests to deliver them once. But this set would need to be extended to a dictionary or rather we use some pytest object

1. when we collect missing this_tests for one go, we should identify the file name. As described above, how-to wasn't obvious even though there should be some way via inspect to go there from item.name (that's all we have in this event unless you want to dissect the item string, which could be instable over releases).
2. we could make a simple pretty print, but supposedly we should add such things to some pytest report I suppose, which gets listed automatically at the very end

Actually, I need to take a small fast.ai break for some days as I am trying to push this through here a little next to my normal live obligations. I hope this is good to merge, maybe with small additions, and we wait for Andrew and then resume?

stas00:
I think it's a bit overcomplicated, but let's get it in and then improve from there.

Thank you for letting me know that you'll be doing some other things for a while, @Benudek. Please take your time.

Thank you for this collaboration so far.

Benudek:
yeeah, its not yet great but I guess a start... I know. did it after 23h local time here or so ;-)

I want to help registering these tests with this_tests as soon as we are there. Other than, that dont get me wrong my preference would be to code all along but my timebudget is somewhat limited for now.

thx for all the tips, good learning curve

PR #1692, opened 2019-02-21
Uppercase class name

yang-zhang:


stas00:
Thank you, @yang-zhang

PR #1696, opened 2019-02-21
final report on this_tests calls missing

Benudek:

### DIFF
--- a/fastai/gen_doc/doctest.py
+++ b/fastai/gen_doc/doctest.py
@@ -16,7 +16,7 @@ class TestAPIRegistry:
     api_tests_map     = defaultdict(list)
     some_tests_failed = False
     has_this_tests = None
-    failed_tests = set()
+    missing_this_tests = set()
 
     @staticmethod
     def this_tests(*funcs):
@@ -35,33 +35,34 @@ class TestAPIRegistry:
             else:
                 raise Exception(f"'{func}' is not in the fastai API")
         try: 
-            TestAPIRegistry.failed_tests.remove(TestAPIRegistry.has_this_tests) 
+            missing_this_test = 'file: '+ relative_test_path(filename) + ' / test: ' + test_name
+            TestAPIRegistry.missing_this_tests.remove(missing_this_test) 
         except:
             None
         TestAPIRegistry.has_this_tests = None
 
-    def this_tests_flag_on(item):
-        TestAPIRegistry.has_this_tests = item.name
-        ## to do: get fully qualifed name from item.name & then find filename. place in dictionary
-        TestAPIRegistry.failed_tests.add(TestAPIRegistry.has_this_tests)
+    def this_tests_flag_on(filename, test_name):
+        TestAPIRegistry.has_this_tests = test_name
+        missing_this_test = 'file: '+ filename + ' / test: ' + test_name
+        TestAPIRegistry.missing_this_tests.add(missing_this_test)
        
     def tests_failed(status=True):
         TestAPIRegistry.some_tests_failed = status
     
-    def this_tests_flag_check(item):
-        if TestAPIRegistry.has_this_tests == item.name:
+    def this_tests_flag_check(filename, test_name):
+        if TestAPIRegistry.has_this_tests == test_name:
             TestAPIRegistry.has_this_tests = None
 
     def registry_save():
+        if TestAPIRegistry.missing_this_tests:
+            print('\n*** Warning: Pls register the following tests with TestAPIRegistry.this_tests:\n')
+            for missing_this_test in TestAPIRegistry.missing_this_tests:
+                print(str(missing_this_test) + '\t')
         if TestAPIRegistry.api_tests_map and not TestAPIRegistry.some_tests_failed:
             path = Path(__file__).parent.parent.resolve()/DB_NAME
             print(f"\n*** Saving test api registry @ {path}")
             with open(path, 'w') as f:
                 json.dump(obj=TestAPIRegistry.api_tests_map, fp=f, indent=4, sort_keys=True, default=_json_set_default)
-        # to do: this should probably go to a test report object
-        #print('\n*** Warning: Pls register the following tests with TestAPIRegistry.this_tests')
-        #pp = pprint.PrettyPrinter(indent=4)
-        #pp.pprint(TestAPIRegistry.failed_tests)   
 
 def this_tests(*funcs): TestAPIRegistry.this_tests(*funcs)
 
--- a/fastai/gen_doc/doctest.py
+++ b/fastai/gen_doc/doctest.py
@@ -43,8 +43,6 @@ class TestAPIRegistry:
 
     def this_tests_flag_on(filename, test_name):
         TestAPIRegistry.has_this_tests = test_name
-        missing_this_test = f"file: {filename} / test:  {test_name}"
-        TestAPIRegistry.missing_this_tests.add(missing_this_test)
        
     def tests_failed(status=True):
         TestAPIRegistry.some_tests_failed = status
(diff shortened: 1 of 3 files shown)


Benudek:
@stas00 fails on ( I think) a maybe different reporting format of pytest - locally it works for me:

conftest.py / pytest_runtest_makereport
test_name =  filenameandmethod.split('::::', 1)[1]

It is a bit hacky to parse the test report like that. Other than that, this code should essentially be fine (report at the very end on missing this_tests calls)

Any other good ideas (as we need the filename) here?

stas00:
Sorry, I'm not understanding what you're saying. Your wrote in the new code:
```  
    test_name =  filenameandmethod.split('::::', 1)[1] 
```
the test suite failed telling you this is where the problem is, 
```
INTERNALERROR>     test_name =  filenameandmethod.split('::::', 1)[1]
INTERNALERROR> IndexError: list index out of range
```
That means that the split has failed, and there is no index [1].

FWIW, the CI is using:

pytest==4.3.0
pytest-runner==4.4

(from looking at its install logs)

Perhaps you're using a different version and the format of those strings isn't the same in those different versions?

Benudek:
So this works locally for me when running make test-full. How do you test on github? Could try to reproduce locally. 

But actually this issue shows parsing that report is too hacky. Rather from the string of the function we need to get to the object. Like you do in your str2obj function. Only, you need the fully qualified name and we just have the simple one here. 

Ideas welcome

stas00:
> So this works locally for me when running make test-full. How do you test on github? Could try to reproduce locally.

You add some debug statements I guess. And I told you already the pytest versions azure CI (not github) is running, so you can match the version and check.

> But actually this issue shows parsing that report is too hacky. Rather from the string of the function we need to get to the object. Like you do in your str2obj function. Only, you need the fully qualified name and we just have the simple one here.

I've never done this before with pytest - I'll have a look when I get a chance.

Benudek:
ok, I look at it when I find time too. 

As fallback: I would still regard report line per test as an option, as soon as we registered with this_tests.

Benudek:
Now seems fine :-) pytest version 4.3.0 locally as here btw

Maybe we should have the string parse in try except and then call functions only when file and test name are found. Then again, in case pytest changes test report format it is probably best to know and hot hide the exception.

So, hope its good t merge now ;-)


@pytest.hookimpl(tryfirst=True, hookwrapper=True)
def pytest_runtest_makereport(item, call):
    outcome = yield
    res = outcome.get_result()
    #individualtests = 0
    individualtests = [s for s in set(sys.argv) if re.match(r'.*test_\w+\.py',s)]    
    if pytest.config.getoption("--testapireg") and not individualtests:
        **filenameandmethod = str(res).split('\'', 1)[1].split('\'', 1)[0]  
        filename =  filenameandmethod.split('::::', 1)[0]   
        test_name =  filenameandmethod.split('::::', 1)[1]**       
        if res.when == "setup": TestAPIRegistry.this_tests_flag_on(filename, test_name)
        if res.when == "call" and res.failed:  TestAPIRegistry.tests_failed()
        if res.when == "teardown":  TestAPIRegistry.this_tests_flag_check(filename, test_name)

stas00:
Here is a much simpler implementation:
```
@pytest.hookimpl(tryfirst=True, hookwrapper=True)
def pytest_runtest_makereport(item, call):
    outcome = yield
    res = outcome.get_result()
    filename, lineno, testname = res.location
    if res.when == "setup":    TestAPIRegistry.this_tests_flag_on(filename, testname)
    if res.when == "teardown": TestAPIRegistry.this_tests_flag_check(filename, testname)
    individualtests = [s for s in set(sys.argv) if re.match(r'.*test_\w+\.py',s)]
    #individualtests = 0
    if pytest.config.getoption("--testapireg") and not individualtests:
        if res.when == "call" and res.failed:  TestAPIRegistry.tests_failed()
```
Please test.

I got the object info here: https://docs.pytest.org/en/latest/_modules/_pytest/reports.html

Thank you.

Benudek:
wow, why did I overlook that location attribute? That's much better indeed.

fastai/gen_doc/doctest.py
```diff
@@ -35,33 +35,34 @@ def this_tests(*funcs):
...
             None
         TestAPIRegistry.has_this_tests = None
 
-    def this_tests_flag_on(item):
-        TestAPIRegistry.has_this_tests = item.name
-        ## to do: get fully qualifed name from item.name & then find filename. place in dictionary
-        TestAPIRegistry.failed_tests.add(TestAPIRegistry.has_this_tests)
+    def this_tests_flag_on(filename, test_name):
+        TestAPIRegistry.has_this_tests = test_name
+        missing_this_test = 'file: '+ filename + ' / test: ' + test_name
```
stas00: no + style please f"" instead.

fastai/gen_doc/doctest.py
```diff
@@ -35,33 +35,34 @@ def this_tests(*funcs):
...
     
-    def this_tests_flag_check(item):
-        if TestAPIRegistry.has_this_tests == item.name:
+    def this_tests_flag_check(filename, test_name):
+        if TestAPIRegistry.has_this_tests == test_name:
             TestAPIRegistry.has_this_tests = None
 
     def registry_save():
+        if TestAPIRegistry.missing_this_tests:
+            print('\n*** Warning: Pls register the following tests with TestAPIRegistry.this_tests:\n')
```
stas00: too complicated - please use join() here - all of it should be one command.

fastai/gen_doc/doctest.py
```diff
@@ -35,33 +35,34 @@ def this_tests(*funcs):
             else:
                 raise Exception(f"'{func}' is not in the fastai API")
         try: 
-            TestAPIRegistry.failed_tests.remove(TestAPIRegistry.has_this_tests) 
+            missing_this_test = 'file: '+ relative_test_path(filename) + ' / test: ' + test_name
```
stas00: no + style please f"" instead.

fastai/gen_doc/doctest.py
Benudek: done

fastai/gen_doc/doctest.py
Benudek: dont understand

fastai/gen_doc/doctest.py
Benudek: done

Benudek:
think this is fine now, imho the report per line was good enough ;-)

fastai/gen_doc/doctest.py
Benudek: I suppose you mean sth like this: print('\n'.join(url.format(year) for year in range(2000, 2016))) ?

Benudek: print('\n'.join(str(missing_this_test) for missing_this_test in TestAPIRegistry.missing_this_tests))

tests/test_basic_data.py
```diff
@@ -16,6 +17,7 @@
 ## Class DataBunch
 
 def test_DataBunch_Create():
+    this_tests(DataBunch.create)
```
stas00: I remind again that we need to use live objects, so all these should be `object.foo`, not `class.foo` - `data.foo` in this case. and all of the remaining changes in this file.

Also please use a separate PR, otherwise this PR will never finish.

stas00:
Thanks, @Benudek.

Sorry this was dragging too long, I made the last fixes and merged it - please review what I did so if you get inspired to send more PRs you will see what's required.

Benudek:
You will only see other PRs from me if I have considerable timebudget at home to prep them and probably discussed all details on the forum :-)

PR #1698, opened 2019-02-21
Add cyclic date/time features to dataframes

herrmann:
Forum post: https://forums.fast.ai/t/fastai-v1-adding-features/23041/109

This patch adds the `add_cyclic_datepart` function to `fastai.tabular` as an alternative (or complementary) form of representation of date/time.

### DIFF
--- a/fastai/tabular/transform.py
+++ b/fastai/tabular/transform.py
@@ -1,9 +1,11 @@
 "Cleaning and feature engineering functions for structured data"
 from ..torch_core import *
 from pandas.api.types import is_numeric_dtype
+from datetime import date, datetime
+import calendar
 
 __all__ = ['add_datepart', 'cont_cat_split', 'Categorify', 'FillMissing', 'FillStrategy', 'Normalize', 'TabularProc',
-           'add_elapsed_times', 'make_date']
+           'add_elapsed_times', 'make_date', 'add_cyclic_datepart']
 
 def make_date(df:DataFrame, date_field:str):
     "Make sure `df[field_name]` is of the right date type."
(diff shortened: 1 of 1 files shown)


fastai/tabular/transform.py
```diff
@@ -13,8 +15,45 @@ def make_date(df:DataFrame, date_field:str):
...
+        rs = tt.tm_hour/24, tt.tm_hour%12/12, tt.tm_min/60, tt.tm_sec/60
+        feats += [f(r * 2 * np.pi) for r in rs for f in fs]
+    if add_linear:
+        if type(d) == date: feats.append(d.year + rs[-1])
+        else:
+            secs_in_year = (datetime(d.year+1, 1, 1) - datetime(d.year, 1, 1)).total_seconds()
+            feats.append(d.year + ((d - datetime(d.year, 1, 1)).total_seconds() / secs_in_year))
+    return feats
+
+def add_cyclic_datepart(df:DataFrame, field_name:str, prefix:str=None, drop:bool=True, time:bool=False, **kwargs):
```
sgugger: We're trying to avoid using unnecessary kwargs now, so that users can see every argument in the documentation. Here you're only hiding add_linear in those kwargs, so I'd explicitly name it.

sgugger:
Thanks! Just one change requested to make the generated docs clearer.
Btw don't forget to add another PR that document your new functionality by updating the doc notebooks ;)

fastai/tabular/transform.py
herrmann: Updated (I also prefer it that way).

herrmann:
Thanks! I'll check how to update the docs and send usage examples in another PR.

sgugger:
It's very well documented [here](https://docs.fast.ai/gen_doc_main.html)

PR #1699, opened 2019-02-21
Add nb examples for SaveModelCallback class

jamesdellinger:
It took some searching through the fast.ai forums for me to figure out how to properly use `SaveModelCallback()`, so I thought I'd add some notebook examples to the two pages in the docs that reference this class.

Hope these examples help future folks who encounter an error when they make the same mistake I did (specifying this callback when creating a `Learner()` object, as opposed to the correct way: doing so inside the `fit()` or `fit_one_cycle()` methods).

### DIFF
--- a/docs_src/callbacks.ipynb
+++ b/docs_src/callbacks.ipynb
@@ -531,6 +531,71 @@
    ]
   },
   {
+   "cell_type": "code",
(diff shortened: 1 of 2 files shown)


sgugger:
Looks nice, thanks a lot!

PR #1704, opened 2019-02-23
Add HTTPS command examples to PR How-to Step 1a

jamesdellinger:
I noticed that some of the commands in Step 1a of the `How to Make a Pull Request (PR)` section in `git.md` only had an SSH example. When following these instructions I unwittingly created an upstream using SSH (even though I had intended to use HTTPS).  I was later momentarily confused as to why git was complaining about a certificate when I went to fetch. 

I hope that with HTTPS  examples for setting upstream and the expected appearance of a proper setup, future users will be less likely to make the same mistake I did.


stas00:
Thank you, @jamesdellinger!

PR #1707, opened 2019-02-24
Additional description about the batch_stats function

shashankhalo7:
The output of 'ImageDataBunch.batch_stats' is a list of two tensors, first being the mean for all individual channels and the other tensor corresponds to the standard deviation for those channels


sgugger:
Thanks a lot! A tiny thing to remember for the feature: please use backticks (see my additional commit) and not ' ' for function names.

shashankhalo7:
OK thanks I will remember that

PR #1712, opened 2019-02-24
Added threshold value for heatmap to be computed and displayed(overla…

at110:
…id) to prevent it when using very small images.


at110:
Also, this feature does not work with to_fp16().
I found a work around like following

 ```
xb,_ = self.data.one_item(im, detach=False, denorm=False)
 xb = xb.float()
 model = self.learn.model.float()
 m = model.eval()
```

in place of 

 ```
xb,_ = self.data.one_item(im, detach=False, denorm=False)
 m = self.learn.model.eval()
```

Do you agree with this? Any better way of doing it?
Please suggest.

sgugger:
Thanks! Don't forget the type annotation when adding an argument in a function(I added it in an additional commit).
For FP16, your fix puts everything in FP32 silently so I'd rather not do it this way. The user can always put it's `Learner` back to FP32 with `learn = learn.to_fp32()`.

at110:
Thank you!!

PR #1713, opened 2019-02-24
Use pynvx as pynvml wrapper on OSX

phenomax:
For full functionality, please wait for https://github.com/1132719438/pynvx/pull/3 being merged.

### DIFF
--- a/fastai/utils/collect_env.py
+++ b/fastai/utils/collect_env.py
@@ -2,8 +2,10 @@
 from ..imports.torch import *
 from ..core import *
 from ..script import *
+from .pynvml_gate import *
 import fastprogress
 import subprocess
+import platform
 
 __all__ = ['show_install', 'check_perf']
 
@@ -152,7 +154,6 @@ def check_perf():
 
     from PIL import features, Image
     from packaging import version
-    import pynvml
 
     print("Running performance checks.")
 
@@ -189,8 +190,8 @@ def check_perf():
     }
     print("\n*** CUDA status")
     if torch.cuda.is_available():
-        pynvml.nvmlInit()
-        nvidia_ver = pynvml.nvmlSystemGetDriverVersion().decode('utf-8')
+        pynvml = load_pynvml_env()
+        nvidia_ver = (pynvml.nvmlSystemGetDriverVersion().decode('utf-8') if platform.system() != "Darwin" else "Cannot be determined on OSX yet")
         cuda_ver   = torch.version.cuda
         max_cuda = "8.0"
         for k in sorted(nvidia2cuda.keys()):
--- a/fastai/utils/pynvml_gate.py
+++ b/fastai/utils/pynvml_gate.py
@@ -1,0 +1,22 @@
+"""Get OS specific nvml wrapper. On OSX we use pynvx as drop in replacement for pynvml"""
+
+import platform
+from ..script import *
+
+def load_pynvml_env():
+    is_osx = platform.system() == "Darwin"
+    if is_osx:
+        try:
+            from pynvx import pynvml
+        except:
+            print("please install pynvx on OSX: pip install pynvx")
+            sys.exit(1)
+
+        pynvml.nvmlInit()
+        return pynvml
+
+    # use nvidia-ml-py3 module
+    import pynvml
+    pynvml.nvmlInit()
+
+    return pynvml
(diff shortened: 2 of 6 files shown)


stas00:
something is wrong with your PR, @phenomax.

phenomax:
@stas00 You committed faster than me, thus the merge conflict :)
Do you want to squash our changed together? I also fixed `utils/collect_env.py` and `callbacks/mem.py`

fastai/utils/collect_env.py
```diff
@@ -189,8 +189,9 @@ def check_perf():
     }
     print("\n*** CUDA status")
     if torch.cuda.is_available():
+        pynvml = get_pynvml()
         pynvml.nvmlInit()
-        nvidia_ver = pynvml.nvmlSystemGetDriverVersion().decode('utf-8')
+        nvidia_ver = pynvml.nvmlSystemGetDriverVersion()
```
stas00: you removed the utf-8 decode call - was it not needed? I don't remember why I put it there in first place

docs/callbacks.html
```diff
@@ -652,7 +652,7 @@ <h3 id="SaveModelCallback"><a href="/callbacks.tracker.html#SaveModelCallback"><
 <div class="output_area">
 
 <div class="output_subarea output_stream output_stdout output_text">
-<pre>bestmodel_1.pth  bestmodel_2.pth  bestmodel_3.pth
+<pre>bestmodel_1.pth  bestmodel_2.pth  bestmodel_3.pth
 </pre>
 </div>
 </div>
```
stas00: this one doesn't belong here, correct? What happens if hit [delete] - will it just remove it from the PR, but leave the file in place?

fastai/utils/get_pynvml.py
```diff
@@ -0,0 +1,20 @@
+"""Get OS specific nvml wrapper. On OSX we use pynvx as drop in replacement for pynvml"""
+
+import platform
+from ..script import *
+
+def get_pynvml():
```
stas00: probably best not to call the module and function by the same name, as both share the same symbol.

In ipyexperiments I called the module pynvml_gate.py, but I'm not attached to it.
https://github.com/stas00/ipyexperiments/blob/master/ipyexperiments/utils/pynvml_gate.py

fastai/utils/collect_env.py
phenomax: I did it, because the pynvx `cudaSystemGetDriverVersion` returns an `int`. In https://github.com/1132719438/pynvx/pull/3 i proposed to change it's return value to string.

docs/callbacks.html
phenomax: Yes, I don't know how it got there. I'll clean up the commit.

fastai/utils/get_pynvml.py
phenomax: This looks like a useful terminology. I'll adapt to that.

fastai/utils/collect_env.py
stas00: But nvmlSystemGetDriverVersion returns a string. so yes, pynvx should do the same. So let's not remove that decode call.

stas00:
And we don't know if and when your PR 1132719438/pynvx#3 gets merged, and that's not enough it has to be released on pypi too. So perhaps we hold off on those parts for when that happens?

fastai/utils/pynvml_gate.py
```diff
@@ -0,0 +1,22 @@
...
+        try:
+            from pynvx import pynvml
+        except:
+            print("please install pynvx on OSX: pip install pynvx")
+            sys.exit(1)
+
+        pynvml.nvmlInit()
+        return pynvml
+
+    # use nvidia-ml-py3 module
```
stas00: this import should probably come first, since pynvx loads it anyway.

stas00:
ok, we shouldn't init inside the wrapper, since it loads w/o nvidia libs, but the init should only happens `if use_gpu`. thank you, CI

phenomax:
Ok. I have squashed our commits together. Would you like to move the init out of the `load_pynvml_env` function to avoid merge conflits?

stas00:
can you git pull my changes in? otherwise every fix you commit, it recreates the conflict. thank you.

stas00:
Looking good on CI - will retest shortly on gpu setup.

Thank you, @phenomax

phenomax:
👍 Thanks for your quick reactions!

PR #1716, opened 2019-02-25
Note for load_learner failure on breaking change

jeffrwells:


stas00:
Thank you, @jeffrwells.

But please remove the other 2 changes, they aren't right.

and yes, we should pull 1.0.45 from pypi, it wasn't released...

PR #1717, opened 2019-02-25
Added __eq__ and __hash__ methods to the ItemBase class and child cla…

jimypbr:
…sses so that they can be compared and counted with collections.Counter.


sgugger:
It's a good feature to add, but it's a bit of a bummer if we have to reimplement for some new types of items. Maybe the basic method should check if the data is array-like or not? Also I think we should use torch for the test as it will likely be tensors by that time.
Also in hash, we should check if the obj attribute exists: a user that defines its custom ItemList might forget to implement that, and we could still fall back on the data attribute in that case.

sgugger:
Just waiting for the tests and it should be good!

sgugger:
Thanks a lot!

PR #1719, opened 2019-02-26
Support for DataBunches with no Validation DL

edave:
Reworks `Databunch.dls()` to not return the empty `valid_dl` (if it is set to `None`). `ImageDataBunch. batch_stats()` is  updated to fallback to using the Training DL if there is no `valid_dl` to use for calculating batch stats.

Added some testing to verify this all works.

### DIFF
--- a/fastai/basic_data.py
+++ b/fastai/basic_data.py
@@ -132,8 +132,12 @@ class DataBunch():
                 self.fix_dl)
 
     @property
-    def dls(self):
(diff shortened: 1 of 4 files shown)


edave:
See [this fast.ai forum post](https://forums.fast.ai/t/a-custom-itemlist-for-a-conditional-gan/37294/5?u=edave) for more background

sgugger:
It seems the test pass so I don't think this breaks any existing behavior. Thanks for the additional tests!

PR #1721, opened 2019-02-26
Correction of incorrect docstrings for freezing of layers

sibowi:
In the docstring for fastai.basic_train.freeze() it was written that the function will “freeze up to last layer”. What actually happens is that the function freezes up to last layer group.
In the docstring for fastai.basic_train.freeze_to(n:int) it was written that the function will “freeze layers up to layer n”. What actually happens is that the function freezes up to layer group n.
Both have been corrected.


sgugger:
Perfect, thanks!

PR #1723, opened 2019-02-27
WIP Add `split_subsets()` to ItemList

sotte:
I tend to work with a small subset of the data first, then increase the dataset size. The data block api offers the `use_partial_data` so select a subset of the data (independent of train and valid). I however want to use full valid ds and only train on a smaller train ds. With the current API your can achieve this by combining `use_partial_data`, some manual math, and `random_split_by_pct`, but this is a bit cumbersome.

I wrote a little function `split_subsets` which allows you to define the size of the train and valid ds independently. It looks something like this:
```python
# do the the normal split like random_split_by_pct
ImageList.from_csv(...).split_subsets(train_size=.8, valid_size=.2)
# only work with a tenth of the train ds but keep the valid ds size
ImageList.from_csv(...).split_subsets(train_size=.08, valid_size=.2)
```

I really think that the name of the method is not ideal. Any suggestions? `random_split_by_pct` has the word `random` in there, maybe this method should as well?

Once the name is fix I'll mention the method in the docs docs_src/data_block.ipynb.

### DIFF
--- a/fastai/data_block.py
+++ b/fastai/data_block.py
@@ -197,6 +197,17 @@ class ItemList():
         cut = int(valid_pct * len(self))
         return self.split_by_idx(rand_idx[:cut])
 
+    def split_subsets(self, train_size:float, valid_size:float, seed=None) -> 'ItemLists':
+        "Split the items into train set with size `train_size * n` and valid set with size `valid_size * n`."
+        assert 0 < train_size < 1
+        assert 0 < valid_size < 1
+        assert train_size + valid_size <= 1.
+        if seed is not None: np.random.seed(seed)
+        n = len(self.items)
+        rand_idx = np.random.permutation(range(n))
+        train_cut, valid_cut = int(train_size * n), int(valid_size * n)
+        return self.split_by_idxs(rand_idx[:train_cut], rand_idx[-valid_cut:])
+
     def split_by_valid_func(self, func:Callable)->'ItemLists':
         "Split the data by result of `func` (which returns `True` for validation set)."
         valid_idx = [i for i,o in enumerate(self.items) if func(o)]
--- a/docs_src/data_block.ipynb
+++ b/docs_src/data_block.ipynb
@@ -1294,6 +1294,22 @@
   {
    "cell_type": "code",
    "execution_count": null,
+   "metadata": {},
+   "outputs": [],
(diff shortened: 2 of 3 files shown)


fastai/data_block.py
```diff
@@ -197,6 +197,20 @@ def random_split_by_pct(self, valid_pct:float=0.2, seed:int=None)->'ItemLists':
         cut = int(valid_pct * len(self))
         return self.split_by_idx(rand_idx[:cut])
 
+    def split_subsets(self, train_size:float, valid_size:float, seed=None) -> 'ItemLists':
+        """Split the items into train set with size `train_size * n` and valid set with size `valid_size * n`.
```
sgugger: Please limit the docstring to one line, then add any other information in the corresponding doc notebook (fastai style guide).

fastai/data_block.py
```diff
@@ -197,6 +197,20 @@ def random_split_by_pct(self, valid_pct:float=0.2, seed:int=None)->'ItemLists':
...
+
+        The total size can be smaller than 1 to work with a subset of the data.
+        """
+        assert 0 < train_size < 1
+        assert 0 < valid_size < 1
+        assert train_size + valid_size <= 1.
+        if seed is not None: np.random.seed(seed)
+        n = len(self)
```
sgugger: Small detail, but `len(self.items)` is safer since `len(self)` pretends to be 1 when it should be 0 sometimes (a hack to go with pytorch DataLoader).

sgugger:
Thank you so much for doing this! If you can, there are two little changes to add (one style and one that could be an issue for empty ItemLists).

fastai/data_block.py
sotte: Done. (I think, the notebook documentation system is still a bit strange to me. Do I have to strip the output? Does that happen automatically? CONTRIBUTING.md is not too clear on this.)

fastai/data_block.py
sotte: That's interesting. Done.

sotte:
I also added the docs to the notebook.

sgugger:
Eh, you write it, you name it ;) Thanks a lot, looks perfect!

sotte:
It was a pleasure! Keep up the good work!

PR #1725, opened 2019-02-27
All plot functions return their figure

sotte:
All plot functions return the `plt.Figure`. This makes it easy to save the figures or tweak them to ones liking.

### DIFF
--- a/fastai/basic_train.py
+++ b/fastai/basic_train.py
@@ -468,18 +468,21 @@ class Recorder(LearnerCallback):
         "Add `names` to the inner metric names."
         self._added_met_names = names
 
-    def plot_lr(self, show_moms=False)->None:
+    def plot_lr(self, show_moms=False)->plt.Figure:
         "Plot learning rate, `show_moms` to include momentum."
         iterations = range_of(self.lrs)
         if show_moms:
-            _, axs = plt.subplots(1,2, figsize=(12,4))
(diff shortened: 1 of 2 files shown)


sgugger:
Looks good, thanks for adding this!

stas00:
Nope, this is not good at all.
e.g. this now prints 2 plots:
learn.recorder.plot()

sgugger:
Ah, it didn't do what I thought it would. Yes returning the figure without a new flag won't work in notebooks, so I'll revert.
We can do this with an extra arg `return_fig:bool=None` that would default to `defaults.return_fig`, `False` by default. That way the user can easily control the default behavior globally but still override it when they want too. @sotte let me know if you want to do it.

sotte:
Of course you're right, the current implementation breaks the workflow of many people. I'll create a new PR adjust the following:
- add a global `return_fig` to `defaults`
- add `return_fig` flag to all plot functions
- use the global `default.return_fig` if `return_fig is None` 

I did a test in a notebook and saw two plots, but added an `;` at the end of the line immediately without even thinking, just habit in these cases. Some people argue that showing every plot automatically in the notebook environment is a great feature, some people would prefer an explicit return :) I mostly work in IDEs/editors and the problem/the second plot is not visible there. 

Thanks @sgugger and @stas00!

PR #1728, opened 2019-02-28
Add `return_fig` argument to all plot functions

sotte:
Fixed all remarks from #1725.

* add a global `return_fig` to `defaults`
* add `return_fig` flag to all plot functions
* use the global `default.return_fig` if `return_fig is None`

Here is a demo in a notebook :)
https://gist.github.com/sotte/633dc6f46367da5e0dbd38afd929e9a3

There is one more plot function `_cl_int_plot_multi_top_losses`. I didn't touch that one because the behaviour of `save_misclassified` is a bit special.

### DIFF
--- a/fastai/basic_train.py
+++ b/fastai/basic_train.py
@@ -468,18 +468,21 @@ class Recorder(LearnerCallback):
         "Add `names` to the inner metric names."
         self._added_met_names = names
 
-    def plot_lr(self, show_moms=False)->None:
+    def plot_lr(self, show_moms=False, return_fig:bool=None)->Optional[plt.Figure]:
         "Plot learning rate, `show_moms` to include momentum."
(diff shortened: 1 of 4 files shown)


sgugger:
Thanks a lot, that's exactly what I had in mind! 
The example notebook is great, if you don't mind, you should take a few examples in it and add them to the docs notebook so that this new argument is properly documented.

sotte:
@sgugger where exactly do you want more documentation?
I think the feature is small enough, the argument name `return_fig` good enough that it doesn't need more docs. I think it would actually hamper the readability of the docs (depending on where you put it).

sgugger:
I think it can be documented in `core` when documenting `defaults` without confusing anyone. People won't know the option is available otherwise.

PR #1732, opened 2019-03-01
Turning on nbtest links and including test_api.json.

bearpelican:
Remaining steps to show test links in docs.fast.ai.

As stas suggested, there's a convenience command:
`make docs`
This will update both tests and docs source links


stas00:
Thank you, @bearpelican! Will be experimenting with it shortly and send feedback.

PR #1733, opened 2019-03-01
Throwing errors in nbtest instead of printing to notebook

bearpelican:
Using relative test folder path instead of depending on user checkout folder name


stas00:
Let's try this one.

PR #1740, opened 2019-03-03
Change formatting of CSV path

richlipkin:
The assignment of csv_path in the Image Cleaner widget was generating an error on certain setups. The line:

csv_path = self._path/'cleaned.csv'

was generating the error:

TypeError: unsupported operand type(s) for /: 'str' and 'str'

So it thought that the / was an operand for some reason. Instead I  used os.path.join, which is much more universally compatible than the code used before.

### DIFF
--- a/fastai/widgets/image_cleaner.py
+++ b/fastai/widgets/image_cleaner.py
@@ -208,7 +208,9 @@ class ImageCleaner():
 
     def write_csv(self):
         # Get first element's file path so we write CSV to same directory as our data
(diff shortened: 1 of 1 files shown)


sgugger:
It's more an issue of `self._path` not being a Path object. Will fix that.

sgugger:
Fixed [here](https://github.com/fastai/fastai/commit/d42de385f29c71423304d1f65653d8eccd816a13).

PR #1744, opened 2019-03-04
update stepper with new cb_handler

KeremTurgutlu:

### DIFF
--- a/fastai/train.py
+++ b/fastai/train.py
@@ -98,7 +98,7 @@ Learner.clip_grad = clip_grad
      
 class AccumulateStepper(LearnerCallback):
     "Does accumlated step every nth step by accumulating gradients"
-
+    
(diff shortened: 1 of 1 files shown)


KeremTurgutlu:
Updating AccumulateStepper based on changes in CallbackHandler.

sgugger:
Looks like I forgot to update this one in my big changes yesterday! Thanks for flagging. Note that there is no need to return `False` since it's the default, only `True` is necessary. Will refactor later today.

sgugger:
If you're interested, [here](https://github.com/fastai/fastai/commit/d7bd9c8e6ae0de03e646eb82f24d725a94ec0062) is the bit of refactor I added. You can do both in a single event (since they are only reset at `on_batch_begin`) and no need to return `False`.

Issue #1553, opened 2019-02-01
Add skip_start argument to all loss plots.


gcp:
Currently, only the learning rate finder has a skip_start option, to ignore the first (few) epochs.

I think this would be useful for all loss plots. 

Currently plot_loss() has a "last" argument, which is vaguely similar, but probably less useful (how does it interact with early stopping?) and in any case a somewhat inconsistent API.

sgugger:
Sounds right. Would you be willing to suggest a PR to address this?

TheAnig:
I can work on it. I'm assuming this is the function [plot_loss()](https://github.com/fastai/fastai/blob/master/old/fastai/sgdr.py#L100)

PR #1746, opened 2019-03-04
[MinorFix] Add skip_start to plot_losses 

krishnakalyan3:
The `plot_loss` signature has been changed to include `skip_start:int=None, skip_end:int=None` instead of `last`.

Ref:
#1553

### DIFF
--- a/fastai/basic_train.py
+++ b/fastai/basic_train.py
@@ -522,15 +522,16 @@ class Recorder(LearnerCallback):
             self.min_grad_lr = lrs[mg]
         if ifnone(return_fig, defaults.return_fig): return fig
 
-    def plot_losses(self, skip_start:int=None, skip_end:int=None, return_fig:bool=None)->Optional[plt.Figure]:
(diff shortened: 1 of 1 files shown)


krishnakalyan3:
@sgugger I am not sure if the Validation plot correct. Could you please review my PR and share your feedback.

Thank you so much.

sgugger:
Thanks for doing this! There is a little problem with `skip_end`: you should aim for the same behavior as `learn.recorder.plot()` so it should be `-skip_end` if `skip_end>0`. Check [here](https://github.com/fastai/fastai/blob/94666ebe0b9f9b8c36d7cbcfc9376465650d56f2/fastai/basic_train.py#L505) for more details.

There is going to be a mismatch with the validation loss indeed, but I can deal with that once we have fixed the `skip_end` part and merged your PR.

krishnakalyan3:
@sgugger thanks for catching this. Please let me know if there are any changes to be made.

sgugger:
Thanks! I'll take it from there.

krishnakalyan3:
@sgugger thanks for the review!.

PR #1747, opened 2019-03-04
Patch

kdorichev:
1. Tests for `replace_all_caps` and `deal_caps` added.
2. `replace_all_caps` was slightly changed to reflect the name -- now it does replace tokens in CAPS.
3. Functions description was clarified.

Tests results on tests/test_text_* 
BEFORE: 1 failed, 21 passed, 8 skipped
AFTER: 22 passed, 8 skipped


sgugger:
Thanks a lot!

PR #1749, opened 2019-03-04
Fix cut parameter

xeTaiz:
Hi there,
in my experiments with completely custom models I came across the following:
The `create_cnn()` method takes a `cut` parameter (either int or callable)
which is passed to `create_body()`. `create_body()` also takes a `body_fn` to support custom cuts (basically just doing `body_fn(model)`).
I found that the `create_cnn()` method only passes the first 3 parameters of the `create_body()` method without ever making use of that optionally callable nature of `cut`.

I changed the `create_body()` function so that its `cut` parameter can be a callable as well and check for its type inside there to take the appropriate action. 

In this same investigation I found that the `num_features_model()` method (which is called right after `create_body()`) assumes the model that is passed in to be an iterable. This is no problem for most archs, as their root module is a `nn.Sequential` (which is iterable). However, if the model is not a `nn.Sequential` it will fail. I basically changed this part so that every model that is passed into this method is converted to a `nn.Sequential` first.
Since this method has caused me even more trouble with fancier custom bodies, I have rearranged the checks for a custom head, so that it is not called when a custom head is supplied.

### DIFF
--- a/fastai/vision/learner.py
+++ b/fastai/vision/learner.py
@@ -46,7 +46,7 @@ def cnn_config(arch):
     return model_meta.get(arch, _default_meta)
 
 def create_body(arch:Callable, pretrained:bool=True, cut:Optional[Union[int, Callable]]=None):
-    "Cut off the body of a typically pretrained `model` at `cut` or as specified by `body_fn`."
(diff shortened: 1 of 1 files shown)


sgugger:
This all sounds great, and having `num_features` executed only if there is no custom head passed seems like a very good idea.
Thanks a lot!

PR #1750, opened 2019-03-04
Make dummy batches size 2

xeTaiz:
Hi there,
I encountered issues with the `dummy_batch()` method in models where I use PyTorch's `squeeze()` method, as it removes the batch dimension for all dummy batches. Changing the dummy batches size to 2 fixes the problem. While this can also be resolved by not using `squeeze()` in that way, I believe it would be nice if the fastai methods are not affected by using `squeeze()`.


sgugger:
Discussed it with Jeremy and we feel it's up to the user to squeeze according to the proper dim (you might get a batch size of 1 in your model independently of `learn.summary`), so closing this.

PR #1753, opened 2019-03-05
add Cadene CNN pretrained models

PPPW:
As discussed in [this forum post](https://forums.fast.ai/t/more-cnn-archs-in-fastaiv1/35836/5), various Cadene pretrained CNN models are added in one file, without introducing an additional dependency to Cadene. 

Users not using these models are not affected; users who want to use these models must install "pip install pretrainedmodels" by themselves, then import the model from this file. 

It's tested in [here](https://github.com/PPPW/deep-learning-random-explore/blob/master/CNN_archs/PR_test.ipynb) (as the tests are too specific so they are not added into repo).

### DIFF
--- a/fastai/vision/models/cadene_models.py
+++ b/fastai/vision/models/cadene_models.py
@@ -1,0 +1,89 @@
+try:
+    import pretrainedmodels
+    _ = pretrainedmodels.inceptionv4
+except (ImportError, AttributeError) as e:
+    print('Error: you must install library \"pretrainedmodels\" to use the model. Try:\n' + 
+          '> pip install pretrainedmodels \n')
+    raise
+
+from torch import nn
+from ..learner import model_meta
+
+__all__ = ['inceptionv4', 'inceptionresnetv2', 'nasnetamobile', 'dpn92', 'xception_cadene', 'se_resnet50', 
+           'se_resnet101', 'se_resnext50_32x4d', 'senet154', 'pnasnet5large']
+
+def inceptionv4(pretrained=False):
+    pretrained = 'imagenet' if pretrained else None
+    model = pretrainedmodels.inceptionv4(pretrained=pretrained)
+    all_layers = list(model.children())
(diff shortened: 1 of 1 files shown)


sgugger:
Thanks! I'll refactor the code a little but this is great!

sgugger:
Ok, refactored all the duplicates. I checked every model could still load, but I may have done some mistake, so if you don't mind too much, can you double-check I didn't break anything? Thanks a lot!

PPPW:
Hi, thanks for the cleaning up, it looks way better and fastai-ish now! The code itself looks good, however some recent changes in the “vision/learner.py” may affect the cases where I deliberately set cut to None (I should set to “noop” probably), I’ll test those cases more.

But one thing I notice is, if I uninstall “pretrainedmodels” and run it, it won’t raise error during importing (never reach line 9). Not sure why this is happening, but that’s why in the original code I checked whether it has inceptionv4 (or any model) in the import part. I wonder was this the case on your side as well, or something weird in my environment? Thanks

sgugger:
Yes cut can be a function or an int now, and the old way of doing None should correspond to `noop`.
I'll test more the import issue and report.

PPPW:
Hi, I have done more checking, they’re all good if:
1. Change the cut of “None” into “noop” (nasnetamobile and pnasnet5large)
2. “pmodels” is not defined, change that into model name string (6 places in the file)

Thanks!

sgugger:
Oh pmodels is my bad, it was when I had imported pretrainedmodels with a shorter name. Will fix with the noop!

sgugger:
And I just made several tries, can confirm I see the message properly displayed if I try
```
from fastai.vision.models.cadene_models import *
```
without pretrainedmodels installed.

PPPW:
You’re right, I used pip uninstall and I thought it was uninstalled, but I found there’s still an empty folder called pretrainedmodels, after deleting it manually now the import error shows. So we’re all good!

PR #1754, opened 2019-03-05
adapt early stopping to new cb_handler

KeremTurgutlu:
Just a minor fix that I came across following the breaking change in CallbackHandler.


sgugger:
Ooops, forgot this one, thanks a lot!

PR #1755, opened 2019-03-05
Reading flow improvements

velaia:
While reading the docs I came upon some sections where the reading flow can be improved.


sgugger:
Sounds good, thanks a lot!

PR #1756, opened 2019-03-05
reduce coupling between DataBunch and Learner

radekosmulski:
replacing an `ifnone` with `or` so that this will no longer be necessary when using a custom `DataBunch`:

```
    data = DataBunch(...l)
    data.train_ds.loss_func = lambda: None
```


sgugger:
Thanks!

PR #1759, opened 2019-03-05
Separate creating cnn arch from cnn learner

radekosmulski:
This decouples creating a cnn arch from cnn learner. Also, changes arg name from `arch` to `base_arch` which improves communicating the intention and readability.

This breaks public API (don't think this is an issue at this point) and quite aggressively aligns naming (`cnn_learner` now aligned with `unet_learner`)

A second, more conservative PR to follow plus additional information.

### DIFF
--- a/fastai/vision/learner.py
+++ b/fastai/vision/learner.py
@@ -9,7 +9,7 @@ from ..layers import *
 from ..callbacks.hooks import *
 from ..train import ClassificationInterpretation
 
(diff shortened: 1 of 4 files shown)


radekosmulski:
I see the failing tests, will fix.

sgugger:
Ok, this is going to require a lot of adjusting, but I've been bugging Jeremy to change that name since... everything else was called learner :)

PR #1761, opened 2019-03-05
Separate creating cnn arch from cnn learner

radekosmulski:
This is a more conservative version of https://github.com/fastai/fastai/pull/1759.

I do not think the naming works as well though as in the other PR, I would be in support of making a more aggressive change. Not fixing the tests for now but if you chose to go with this PR, will gladly do so (like I did for the other PR)

I have seen some conversation on the forum and I think the library is in too early of a state to maintain a stable public API, but those are just my two Polish groszes.


sgugger:
I prefer the other one and Jeremy agreed.

PR #1764, opened 2019-03-06
fix model summary output format

gurvindersingh:
Current code returns summary as string and notebook renders it as such without treating `\n` as newline char. Code need to call `print()` and return its output to be rendered by notebook to correctly format. 

Before

image

After

image


sgugger:
No, the fact the function doesn't print anything and returns text is intended. It's like when you type `learn.model`: PyTorch doesn't print the model, it returns it, so if you have a later statement in the same cell, you don't see the model.
We do have to find a way to make the result show more prettily in jupyter, that's a given.

gurvindersingh:
I was not sure if that was intended or not. So now I just have to run like this `print(learn.summary()`

sgugger:
You may not have to. problem comes from the representation of the strings and the way they are displayed in jupyter. I hacked a little thing to fix this.

PR #1766, opened 2019-03-06
fix tensorboard when running model to find learning rate.

gurvindersingh:
I get exception as `valid_ds` attribute missing when running `learner.lr_find()` with Tensorboard integration. This PR fixes this.


sgugger:
Thanks a lot!

PR #1771, opened 2019-03-07
Update data_block.py

RealLankinen:
This line of code will make string extensions work. Previously if someone gave a string extension instead of array it caused a nasty problem that was hard to debug.


sgugger:
As you can see, this makes the tests fail because sometimes a set of extensions is passed and not a list. There is a convenience function in fastai that will make the thing you pass a list, it's `listify`. Be careful to just leave `extensions` as `None` if it was, as it's tested afterward.

RealLankinen:
Thank you for giving feedback. Now it is better.

sgugger:
Thanks!

PR #1773, opened 2019-03-07
Update basics.py

illcat:


sgugger:
Thanks, but callbacks should be imported on a need basis, not in basics.

PR #1774, opened 2019-03-07
fix __all__ in vision.learner

radekosmulski:
When `create_cnn` got changed to `create_cnn_model`, the name was not changed in `__all__` and cannot be imported now.

Would love a quick post1 release or something like that to be able to use this when installing a specific version :slightly_smiling_face:


stas00:
Thank you, @radekosmulski!

Should be doable, but let's see if there is anything else that needs to go in

sgugger:
FYI removing `create_cnn` broke the deprecation warning (it's now throwing an error since it's not in all).I put it back.

PR #1779, opened 2019-03-07
fix callback ShowGraph not matching new callback system

egonuel:
callback ShowGraph was still returning False instead of a dict.
The PR changes that to return an empty dict, which is now also returned regardless of existing last_metrics.


jph00:
Thanks!

Issue #1480, opened 2019-01-18
load_learner: unsupported operand type(s) for /: 'str' and 'str'


RomRoc:
When I export trained model everything is ok:
`learn.export(fname=MODEL_PKL)`

Then when I load it:
`learn = load_learner(path=COLAB_DIR)`

I get:
```
/usr/local/lib/python3.6/dist-packages/fastai/basic_train.py in load_learner(path, fname, test)
    459 def load_learner(path:PathOrStr, fname:PathOrStr='export.pkl', test:ItemList=None):
    460     "Load a `Learner` object saved with `export_state` in `path/fn` with empty data, optionally add `test`."
--> 461     state = pickle.load(open(path/fname, 'rb'))
    462     model = state.pop('model')
    463     src = LabelLists.load_state(path, state.pop('data'))

TypeError: unsupported operand type(s) for /: 'str' and 'str'


```
It works using a Path instead of a string:
`learn = load_learner(path=Path(COLAB_DIR))`

https://github.com/fastai/fastai/blob/f68ae0e4b0bd7d644d42f53ac2886b48e72224b5/fastai/basic_train.py#L460

sgugger:
Thanks for flagging!

PR #1780, opened 2019-03-07
fix(load): fix load_empty unsupported operand type(s) for /: 'str' an…

relax94:
The same bug like #1480 with code style like **load_state** method.


stas00:
thank you, @relax94

PR #1788, opened 2019-03-08
Minor improvement to Data block docs

odysseus0:
Minor improvement to Data block docs


sgugger:
Thanks! Please remember not to run the script to build docs as it makes reading the diff very hard (and would be prone to merge conflicts) though.

odysseus0:
Got it! I will remember it next time.

PR #1792, opened 2019-03-08
Extended documentation for Wide ResNet

RikTikTav:
New notes were added for the WideResNet class to help clarify its use.


sgugger:
Thanks a lot!

PR #1797, opened 2019-03-09
Fix gan to adapt callback API changes

PPPW:
Fix GAN to adapt the changes in commit "Change in the Callback API". Works fine if validation set is not empty. 

If the validation set is empty, although training works fine, the `gen_loss` and `disc_loss` still won't be added to `last_metrics` and won't show in the progress bar. The `gen_loss` and `disc_loss` are added in `on_epoch_end` of `GANTrainer`, if `last_metrics` is not `None`, (with my changes here) the two losses will be added to it and recorded (the `Recorder` is called after `GANTrainer`); But when the validation set is empty, `self.state_dict['last_metrics']` is set to None in `on_epoch_end` of `CallbackHandler`, then I'm not sure how to add the GAN metrics to it. 
Perhapse we shouldn't set `self.state_dict['last_metrics']` to `None` (but set to an empty list instead), so all callbacks can still share and modify it?

### DIFF
--- a/fastai/vision/gan.py
+++ b/fastai/vision/gan.py
@@ -99,7 +99,7 @@ class GANTrainer(LearnerCallback):
         self.switch(self.gen_mode)
         self.closses,self.glosses = [],[]
         self.smoothenerG,self.smoothenerC = SmoothenValue(self.beta),SmoothenValue(self.beta)
-        self.recorder.no_val=True
+        #self.recorder.no_val=True
         self.recorder.add_metric_names(['gen_loss', 'disc_loss'])
         self.imgs,self.titles = [],[]
 
@@ -111,7 +111,7 @@ class GANTrainer(LearnerCallback):
         "Clamp the weights with `self.clip` if it's not None, return the correct input."
         if self.clip is not None:
             for p in self.critic.parameters(): p.data.clamp_(-self.clip, self.clip)
-        return (last_input,last_target) if self.gen_mode else (last_target, last_input)
+        return {'last_input':last_input,'last_target':last_target} if self.gen_mode else {'last_input':last_target,'last_target':last_input}
 
     def on_backward_begin(self, last_loss, last_output, **kwargs):
         "Record `last_loss` in the proper list."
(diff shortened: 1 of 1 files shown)


sgugger:
You should use `add_metrics`that will directly do that check for you and return an updated list.
Thanks in any case, I had forgotten those and this PR is really needed!

PPPW:
Hi @sgugger, right, I saw `add_metrics` in the `Recorder` was removed which causes this error. Do you mean we can now use the `add_metrics` in `torch_core`? It checks for `None`, yes. But if `last_metrics` is `None` in `on_epoch_end` of `GANTrainer` (only happens when valid set is empty), my solution here does't work, and I don't see how to make it work without further changes. May I missed something? 

If `last_metrics` is an empty list, I can modify it in `GANTrainer` and the `Recorder` will use the updated list, it's like pass by reference; but if it's `None` and I reset it to empty list, in the `Recorder` it's still `None`, so this trick won't work anymore...

sgugger:
The `GANTrainer` callback happens before the `Recorder` (see the [order](https://github.com/fastai/fastai/blob/b6986e1dbf8b2a52e19e6c12e5f90131cdb90825/fastai/vision/gan.py#L74)) so if it returns an updated `last_metrics` list, this is what `Recorder` will receive.
I'm not sure of the problem you think there is here?

PPPW:
Sorry for the confusion. In `_call_and_update` of the `CallbackHandler`, the `state_dict` is passed into each callback, in the `order` as you mentioned. But the updates of `last_metrics` won't be got by later callbacks if it's `None`. In the following example: 

Untitled

My understanding may be wrong, there may be better ways to handle this, but this is what the concern is, hope I made it clearer, as there're quite a few changes going on... 

Anyway, the problem I'm concerned about is, if I train a GAN with empty validation set, the `gen_loss` and `disc_loss` are not showing in the console, because they are added in the `GANTrainer` but it's still `None` in the `Recorder`, for the reason explained above. I wonder could you replicate this issue?

sgugger:
It'll be easier to show you. Let me merge this and I'll update the line.

sgugger:
Done in [this commit](https://github.com/fastai/fastai/commit/3305fd33dd5e7cc4e07806929396e776e25f934d). By returning the result of `add_metrics` (which checks for None), we make sure the metrics are updated and sent to the recorder.

PPPW:
😅... I see, that's how it works... Things get updated in the `_call_and_update` of the dict `new`. I'm still thinking in the old mechanism. Thank you for the explanation! 

Then I think my line in the `on_epoch_end` of `GANTrainer` should be removed (first line), it's redundant and will add metrics twice

sgugger:
Oh yes, will do!

PPPW:
One minor issue again: it seems “add_metrics” only handles None “mets”, not None “last_metrics”... When the validation set is empty we got None “last_metrics”, so we need to check for that somewhere as well...

PR #1799, opened 2019-03-09
just a reminder to do pip install -e ...

Benudek:
name of branch misleading, this is just adding a print to not forget pip install ... 

I often forget and need some clicks into the docs and even raised an error for that, so we can avoid that maybe

### DIFF
--- a/tools/fastai-make-pr-branch
+++ b/tools/fastai-make-pr-branch
@@ -154,6 +154,7 @@ if [[ "$clone_is_needed" -eq 1 ]]; then
 else
     echo -e "cd .\n"
 fi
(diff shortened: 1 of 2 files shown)


jph00:
Thanks!

stas00:
This is incorrect. ".[dev]" applies only to the fastai repo.

Benudek:
ok, sorry about that, took the text basically from the doc. What would be the correct text to qualify for ?

If helpful and you don't wanna correct directly, can make a simple PR.

stas00:
it should check that both, the user and the repo are 'fastai' and only then suggest what your PR did.

PR #1802, opened 2019-03-10
Fix filter_by_folder

odysseus0:
Please refer to the notebook below for more details:

https://nbviewer.jupyter.org/gist/odysseus0/d7b10d1f0e226d74ff05f2cd6fdb5e00

### DIFF
--- a/fastai/data_block.py
+++ b/fastai/data_block.py
@@ -151,8 +151,11 @@ class ItemList():
     def filter_by_folder(self, include=None, exclude=None):
         "Only keep filenames in `include` folder or reject the ones in `exclude`."
         include,exclude = listify(include),listify(exclude)
+        include = [Path(o) for o in include]
+        exclude = [Path(o) for o in exclude]
+        
         def _inner(o):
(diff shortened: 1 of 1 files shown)


sgugger:
Oh yes, this doesn't work properly in this instance, thanks for flagging.
The problem with converting everything to a Path object is that it can be very slow (for datasets like imagenet for instance), so we would need a speed check, and if this doesn't pass it, we will need another way of fixing the problem.

odysseus0:
Sign. I guess that most convenience comes with a performance overhead. `pathlib` is so much nicer to use compared to `os`.

odysseus0:
I guess this is a place where there needs to be a consensus, about where we should continue to use `pathlib` for the path handling, or even whether we should continue at all if the performance overhead is indeed so bad. Enforcing an agreed upon way of doing things, at least internally inside the library, will help us avoid bugs like above.

sgugger:
I've just pushed another way to fix. Note that it won't work with your example of '3', it filters the base folders only.

PR #1803, opened 2019-03-10
Make heatmap work with rectangular images

velaia:
Training a model on rectangular images and afterwards using the ClassificationInterpretation and plot_top_losses resulted in heatmaps that were quadratic on top of rectangular images. With this little change it now works for quadratic and rectangular images.

**Before** (heatmap not well aligned and reaches down below the image boundaries):
Screenshot 2019-03-10 at 16 11 30

**After**:
Screenshot 2019-03-10 at 16 11 43

PS: I had first used a conditional (if) statement but then found that the _im.shape[-2:]_ in combination with reversing using _*sz[::-1]_ works both for quadratic and rectangular images.


velaia:
BTW: Nice to have the heatmap feature so easily available now 👍 Great job!

sgugger:
Oh good point, thanks!

PR #1805, opened 2019-03-10
Test Registry merge new with existing

Benudek:
wrt https://forums.fast.ai/t/test-registry-project/38344/216

@stas00 If there is value & a way to make merge_with_old_registry even more compact (e.g. the loops, if then) happy to look at it. 

For the rest, functionally: passed the local tests I was able to come up with ;-)

### DIFF
--- a/fastai/gen_doc/doctest.py
+++ b/fastai/gen_doc/doctest.py
@@ -70,7 +70,7 @@ class TestAPIRegistry:
         if path.exists() == False: return 
         with open(path, 'r') as f: old_api_tests_map = json.load(f)   
         if old_api_tests_map == False: return
-        print(f"\n*** Saving test registry @ {path}")
+        print(f"\n*** Merging test registry @ {path}")
         for func_fq_key, new_entries in TestAPIRegistry.api_tests_map.items():
             if func_fq_key in old_api_tests_map: 
                 for new_entry in new_entries:
--- a/fastai/gen_doc/doctest.py
+++ b/fastai/gen_doc/doctest.py
@@ -1,6 +1,7 @@
 import sys, re, json, pprint
 from pathlib import Path
 from collections import defaultdict
+from itertools import chain
 from inspect import currentframe, getframeinfo, ismodule
 
 __all__ = ['this_tests']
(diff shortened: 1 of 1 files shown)


stas00:
It looks good, but I don't have a way to test whether it works correctly or not.

Can you please add a test that validates its functionality?

i.e. 
1. extract the merging of 2 dicts into its own function
2. write a new test that merges 2 dicts using that new function (simple case and edge cases) and validates that it does the right thing
3. use that function

Thank you.

Benudek:
Yes, makes sense will add a test soon

Benudek:
Stas, makes sense :-) 

However - on 2nd look and 3rd look: it looks a little hairy to write tests for this. e.g. setting up data, but also I might need to rewrite the code a little (are you suggesting this with (1)?) to be able calling functions independently.

I wonder if we can be a little 'pragmatic' here  it's a utility function and we have some more-core function without tests ;-) If it fails in 'production' as far as I can see it wouldn't be a 'biggie' and we could fix then based on the edge case detected, that right now I couldn't predict.

So, what I mainly did is below - and yes it's time-consuming ( I always re-ran all tests to try breaking things) and likely not covering all edge cases: 

**1. delete the file test_api_db.json
2. run pytest -sv --testapireg** 

NOW: after that the merge function will call

**3a. open file in editor and take a random entry and overwrite the entry**

_FROM   "fastai.basic_data.DataBunch.create": [
        {
            "file": "tests/test_basic_data.py",
            "line": 19,
            "test": "test_DataBunch_Create"
        },
        {
            "file": "tests/test_basic_data.py",
            "line": 33,
            "test": "test_DataBunch_no_valid_dl"
        }
    ],

TO   "fastai.basic_data.DataBunch.create": [
        {
            "file": "DUMMY",
            "line": 19,
            "test": "test_DataBunch_Create"
        },
        {
            "file": "tests/test_basic_data.py",
            "line": 33,
            "test": "test_DataBunch_no_valid_dl"
        }
    ],_

EXPECTATION: when rerunning 

**4a. run pytest -sv --testapireg** 

I want to see both: the DUMMY entry and tests/test_basic_data.py

I also changed other lines here with the same expectation and result.

This tests the if-part of the function

_if func_fq_key in old_api_tests_map: 
                for new_entry in new_entries:
                    if new_entry not in old_api_tests_map[func_fq_key]:
                        old_api_tests_map[func_fq_key].append(new_entry)
            else:
                old_api_tests_map[func_fq_key] =  new_entries_            

The else part is tested here

**3b. open file in editor and take out all of "fastai.basic_data.DataBunch.create": ** 
**4b. run pytest -sv --testapireg** 

I expect to see the entire entry from above see FROM.

If you give me some ideas what else you would test, I could run some more tests manually - would that help?

stas00:
That's why I said, extra the functionality to a function that is totally unrelated to test registry:

```
def merge_dicts(old, new):
...
return merged
```

test:
```
old = { a: [1, 2, 3], b: [4, 5, 6])
new = { a: [11, 12, 13], c: [9])
expected = { a: [11, 12, 13], b: [4, 5, 6], c: [9])
merged = merge_dicts(old, new)
assert expected == merged
```
done. 
this is untested. just showing what I meant.

of course adjust old/new to match our needs - i.e. mimic the same structure.

Benudek:
understand, makes sense. 

Unfortunately I wont get to this any time soon :-( 

So need to close this for now.

stas00:
No problem, I will sort it out.

I also wanted to communicate to you that I find it very difficult to work with you, @Benudek.

I trust you will find someone else with a better character compatibility to continue forward.

stas00:
update: I implemented all this functionality including tests, so nothing else needs to be done.

Benudek:
No issues for me Stas at all. For sure I have a challenge to commit a significant timebudget, so I do not on plan

PR #1807, opened 2019-03-11
Test for cont_cat_split

kdorichev:
This patch:

1. Adds a test for cont_cat_split() in tabular/transforms.py

1. Adds return value to cont_cat_split()

1. Corrects a minor typo.


sgugger:
Thanks a lot!

PR #1808, opened 2019-03-11
bug fix to WideResNet. Change to input for flex channel input

RikTikTav:
In WideRestNet the input values for the last BatchNorm and Linear Layers has the same value as the 4th value in n_channels. This will lead to failure _unless_ the `num_groups` = 3. 
Also added an extra input of the number of channels at input `n_in_channels`. Set the default to 3 channels.
Brought this chnge up in forum...  https://forums.fast.ai/t/developer-chat/22363/868?u=aarons


sgugger:
Thanks for fixing!

PR #1815, opened 2019-03-13
Fix the get_preds(DatasetType.Train, ordered=True)

PiotrCzapla:
The code currently tries to resort the predictions after it  receives it from the super().get_preds unfortunately it fails to do so as it twice takes the order from the sampler which in case of randomized samplers is different each time.

A bit of dirty fix is to use the same random seed during get_pred and when getting the order from  random sampler.
But it would be much better to make it possible to introduce an intereface to the sampler that let us disable the randomization.


sgugger:
This is a bit hacky but I can't think of another way, so merging. I'll change if I get a better idea. Thanks!

PR #1816, opened 2019-03-13
Adding include_bos and include_eos to tokenization options

ohmeow:
Following changes include:

1. Added `EOS` token (set = 'xxeos')
2. Added `include_bos` and `include_eos` arguments to `_join_texts` (defaults set to work so it runs the same as it does now).
3. Updated `_get_processor` and various `_fromX` methods to include these new arguments


sgugger:
Sounds good. Thanks a lot!

PR #1817, opened 2019-03-14
Test/unet block

yyoon62:
I wrote this test to check whether U-Net blocks output the correction activation shapes.


sgugger:
Awesome, thanks a lot!

PR #1820, opened 2019-03-14
fbeta metric: last + current with "add_metrics" fun

tolitius:
in FBeta metrics in case of a set `average` (i.e. to `'macro'`) final `metrics` will be a tensor while `last_metrics` will be a list, hence can't be just `+`ed.
https://github.com/fastai/fastai/blob/99044b6fbf2774e83de6cf8a910b30c375738db8/fastai/metrics.py#L214-L216
this change would call `add_metrics` instead, also makes it consistent with all other metrics classes


sgugger:
Oh yes, that's better, thanks!

PR #1827, opened 2019-03-16
added some clarity to what label_cls is and multi-regression usage

jmhsi:
Added some documentation to the datablock api explaining more about `label_cls` usage and how to get correct labeling for multi-regression. See [here](https://forums.fast.ai/t/nlp-multi-regression-with-datablock-api/40950)


sgugger:
Perfect, thanks!

PR #1828, opened 2019-03-16
Data block minor fixes

jonppe:
-fix test_wrong_order by adding missing import and making test more strict
-fix typo in filter_by_folder and add unit test for the metohd. The method is still broken for string items


sgugger:
Oh yes, this is an annoying typo. Your test doesn't pass however so that needs fixing. Be careful that `filter_by_folder` filters... by folder, so you need to have items that look like {folder}/{fname}. Then you use names that aren't the same.

jonppe:
Thanks for pointing out using proper folder names in test. Now I actually read what the code is doing.
I added the fix to the method and the unit tests.

sgugger:
Thanks a lot!

PR #1829, opened 2019-03-16
Csv logger append

KeremTurgutlu:
Adding `append` option to `CsvLogger` callback. I've realized that only the most recent logs were kept as we train `learner`. For example this arises a problem especially when we are using the same callback but do training in multiple stages, e.g. fine-tuning `stage-1`, `stage-2` and `stage-3`. This option as a flag would allow user to log all three stages as they do training in different code cells.

One drawback is that user needs to parse the csv after training is done if append option is used since everything is logged including lr_finder, stage-1, lr_finder, stage-2, lr_finder, stage-3, etc..:

image


sgugger:
Sounds useful, thanks!

stas00:
Please don't forget to update the docs, otherwise new features remain undocumented.
https://docs.fast.ai/callbacks.csv_logger.html

The separation of docs and code makes it very easy to neglect docs.

Thanks.

KeremTurgutlu:
That's super important, sorry about it! Will fix it now. Thanks.

Issue #1830, opened 2019-03-16
Error when loading an LM encoder; 'map' is an unexpected argument


clinchergt:
**Describe the bug**
I'm trying to load an encoder from a previously fine-tuned LM, however I'm getting the following error:

```
TypeError                                 Traceback (most recent call last)
<ipython-input-9-ee84f2ddf586> in <module>
----> 1 learn.load_encoder('ft_enc-en')

/opt/anaconda3/lib/python3.7/site-packages/fastai/text/learner.py in load_encoder(self, name, device)
     67         if hasattr(encoder, 'module'): encoder = encoder.module
     68         encoder.load_state_dict(torch.load(self.path/self.model_dir/f'{name}.pth'))
---> 69         encoder.load_state_dict(torch.load(self.path/self.model_dir/f'{name}.pth', map=device))
     70         self.freeze()
     71 

TypeError: load() got an unexpected keyword argument 'map'
```

**Provide your installation details**

I installed fastai to get the latest version from github: `pip install git+https://github.com/fastai/fastai.git`.

**To Reproduce**
1. Train a language model and save the encoder using `save_encoder`.
1. Load the encoder to a classifier using `load_encoder`.

**Expected behavior**

Have the encoder load properly.

**Additional context**

Commit that introduced this bug was this: https://github.com/fastai/fastai/commit/ca5aad377b975f4fc86aa42d8d964571070e050d

PR #1831, opened 2019-03-17
Fix encoder loading by fixing argument name

clinchergt:
Argument is called `map_location` and not just `map`.

This also fixes [issues 1830](https://github.com/fastai/fastai/issues/1830)


sgugger:
My mistake, thanks for fixing!

PR #1832, opened 2019-03-17
Fix typo in data_block.py

cberner:
This fixes a crash due to `isinstance` being misspelled


sgugger:
Good catch! Thanks a lot but @jonppe beat you to this in a previous PR (#1828) ;)

PR #1834, opened 2019-03-17
Add FASTAI_HOME as optional location of config file

tg00:
[I didn't create a thread on the fastai forums as this code change seems tangential to the actual functionality of fastai]

This pr adds FASTAI_HOME environment var which works similar to TORCH_HOME (in [pytorch]/torch/utils/model_zoo.py).

I run fastai in a docker container and ~ points to /root inside the container, which means that any downloads go inside the container and consequently are lost when the container is stopped unless I copy them from the container (and copy them back in when the container is started).

Setting FASTAI_HOME to an external VOLUME (e.g. 'VOLUME /home' and 'ENV FASTAI_HOME=/home/.fastai') in the dockerfile saves all downloads outside the container (via the config.yml which also has /home/.. as for data_path and model_path).

### DIFF
--- a/docs_src/datasets.ipynb
+++ b/docs_src/datasets.ipynb
@@ -100,8 +100,11 @@
    "source": [
     "For the rest of the datasets you will need to download them with [`untar_data`](/datasets.html#untar_data) or [`download_data`](/datasets.html#download_data). [`untar_data`](/datasets.html#untar_data) will decompress the data file and download it while [`download_data`](/datasets.html#download_data) will just download and save the compressed file in `.tgz` format. \n",
     "\n",
-    "By default, data will be downloaded to `~/.fastai/data` folder.  \n",
-    "Configure the default `data_path` by editing `~/.fastai/config.yml`.  "
+    "The locations where the data and models are downloaded are set in `config.yml`, which by default is located in `~/.fastai`. This directory can be changed via the optional environment variable `FASTAI_HOME` (e.g FASTAI_HOME=/home/.fastai).\n",
+    "\n",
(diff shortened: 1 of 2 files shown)


sgugger:
I don't see why not, but if we do this, then the default config should also use that env variable for the defaults to the  data path and the models path.

stas00:
plus it needs to be documented. Probably at: https://docs.fast.ai/datasets.html#Config
https://docs.fast.ai/gen_doc_main.html#process-for-contributing-to-the-docs
Thank you.

tg00:
Okay, good points.
I'll update the pull request with the changes for the data path and models path and I'll also update the pull request with an updated datasets.ipynb with a reference to the optional FASTAI_HOME.

tg00:
Ive amended the commit.

sgugger:
Thanks a lot, it's looking good!

tg00:
I just found out that I didn't update the .html :-(
Looking at bearpelican's commit a11aa0f, I noticed that both .ipynb and the .html were committed.
I think I misread the documentation and thought it would be done after the commit on the github side.

Should I make a new PR for the .html update ?

sgugger:
No you don't need to do it directly (and we prefer you don't as it makes the diff unreadable). I haven't re-built the docs yet, will do this morning (in general it's done once a day).

stas00:
> Should I make a new PR for the .html update ?

To further clarify:

The maintainers usually commit both and bearpelican is one of the maintainers, so he commits both. Sylvain usually rebuilds all online docs once a day.

For contributors, we try to make things as simple as possible, so asking to just follow the minimal steps required as described here: https://docs.fast.ai/gen_doc_main.html#process-for-contributing-to-the-docs And also as Sylvain said not having any autogenerated files in the PR makes it easier to review your work.

PR #1839, opened 2019-03-18
Fix links to correct model directory

iamshwin:
These links in the docs currently go to the wrong place.

This is where I think they should go.

### DIFF
--- a/docs/text.learner.html
+++ b/docs/text.learner.html
@@ -56,9 +56,9 @@ summary: "Easy access of language models and ULMFiT"
 <div class="text_cell_render border-box-sizing rendered_html">
 <p>The model used is given by <code>arch</code> and <code>config</code>. It can be:</p>
 <ul>
-<li>an <a href="/text.models.awd_lstm.html#AWD_LSTM"><code>AWD_LSTM</code></a>(<a href="https://arxiv.org/abs/1708.02182">Merity et al.</a>)</li>
-<li>a <a href="/text.models.transformer.html#Transformer"><code>Transformer</code></a> decoder (<a href="https://arxiv.org/abs/1706.03762">Vaswani et al.</a>)</li>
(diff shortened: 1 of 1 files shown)


sgugger:
Thanks for catching this, but it's useless to change the html directly since they are auto-generated (so your fix will be erased the next we build the docs). @bearpelican can we fix the reason behind those broken links?

stas00:
@iamshwin, you're invited to re-apply this same fix to the source, which is  `docs_src/text.learner.ipynb`

The docs explain not to edit html, but ipynb instead. Please see:
https://docs.fast.ai/gen_doc_main.html#process-for-contributing-to-the-docs

Thank you!

stas00:
> @bearpelican can we fix the reason behind those broken links?

Yeah, something is messed up in the doc builder:
```
grep TransformerXL docs_src/text.learner.ipynb
    "- a [`TransformerXL`](/text.models.transformer.html#TransformerXL) ([Dai et al.](https://arxiv.org/abs/1901.02860))\n",
```
It's written correctly \`TransformerXL\` so it should be updated to the new location? Or perhaps the build script only expands \'Name\' once and then never tries to update it if it's expanded? Perhaps that's the culprit to our many broken anchors.

bearpelican:
Yup  going to check why those links aren't being updated

bearpelican:
Links are being updated. 

Problem is that we assume the doc path should be the same relative location as the module path. 
Docs builder currently doesn't have any idea that - 'text.models.transformer.py' module should be mapped to 'text.models.html'

I'll update/hardcode the links manually for now and think about how to handle 'module -> docs' mappings

stas00:
Ah, I see.

Perhaps it could check the parent module if it has the NAME defined in there and use it instead? 

One thing for sure, it shouldn't link to non-existing docs, which it does now. i.e. there is no text.models.transformer.ipynb - does it make sense?

PR #1840, opened 2019-03-19
Update freeze docsrc

EmbraceLife:
I have added a little explanation to the term 'layer group' of [`freeze`](https://docs.fast.ai/basic_train.html#Learner.freeze)


stas00:
You probably did 'git commit -A' don't do that, use -a (lower case `a` instead) it will only add files which are already under git. So yes, the extra file shouldn't be there.

Also the beginning of the notebook has some new stuff which shouldn't be there. Probably from your experiments with things that weren't in the 4-steps :)

EmbraceLife:
image

@stas00 Thanks for your helpful guides! I think I have done all the corrections. 
However, as you can seen there is the commit 845ed12 which is crossed, I think it was my attempt to merge my new feature branch with master. Is it not necessary?

stas00:
no, the red [x] means that some CI test failed - you can click on it to see what, but often it's not a real problem, since some of our tests occasionally fail because they are non-deterministic.

You can see the sum-total of your changes by clicking on the [ Files changed ] tab above, all is good and I merged it.

Thank you, @EmbraceLife

PR #1841, opened 2019-03-19
Setting the correct nf when concat_pool=False

renato145:
Fixed `nf` on `create_cnn_model`  being multiplied by 2 when `concat_pool=False`.


sgugger:
Thanks for flagging! I've just put everything on one line.

PR #1845, opened 2019-03-20
Fixed few typos in the vision.image.html doc page.

agiUnderground:
Documentation typos fix.


sgugger:
Thanks for flagging this, but you need to change the notebook that is behind this page (so `vision.image.ipynb`) and no the html page itself: docs are auto-generated so your change will be overwritten the next time we build the docs if we don't fix the source of the problem.
More information [here](https://docs.fast.ai/gen_doc_main.html).

PR #1847, opened 2019-03-20
Scripted forget_mult

t-vi:
This replaces forget_mult with a scripted version. My preliminary measurements show that it is within 5-10%  of the custom kernel.
I have not deleted the kernels.


t-vi:
So from what I can tell, the stable PyTorch version doesn't like the script, so this is not for now.

PR #1853, opened 2019-03-21
Update metrics.html

iyaja:
Corrected a minor error where <code>metrics</code> did not appear in the right format since [code]metrics[/code] was used instead (line 20).


stas00:
> Do not edit the docs/*html files - those are autogenerated and if you change those, your changes will get overwritten.

From: https://docs.fast.ai/gen_doc_main.html#process-for-contributing-to-the-docs

Kindly please re-submit with edits to ipynb instead. Thanks.

PR #1854, opened 2019-03-21
Update metrics.ipynb

iyaja:
Corrected a minor error where `metrics` did not appear in the right format since [code]metrics[/code] was used instead (line 14).

### DIFF
--- a/docs_src/metrics.ipynb
+++ b/docs_src/metrics.ipynb
@@ -11,7 +11,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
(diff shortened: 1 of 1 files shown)


stas00:
Thank you, @iyaja

iyaja:
@stas00, why hasn't it updated on [docs.fast.ai](https://docs.fast.ai/metrics.html#Training-metrics)?

stas00:
We autogenerate the docs about once a day, so your change will become visible on docs.fast.ai then.

PR #1855, opened 2019-03-21
Update index.ipynb

iyaja:
Correct discrepancy between installation instructions on [docs.fast.ai](https://docs.fast.ai/index.html#Installation-and-updating) and the [README file on GitHub](https://docs.fast.ai/index.html#Installation-and-updating). Not sure which version is correct, but the README was updated more recently. Please ignore if this is how the instructions were designed. I might have missed something.


stas00:
Correct. Thank you, @iyaja

PR #1856, opened 2019-03-21
fix typo in basic_data

k0ala:
change tfms to dl_tfms


stas00:
> Do not edit the docs/*html files - those are autogenerated and if you change those, your changes will get overwritten.

From: https://docs.fast.ai/gen_doc_main.html#process-for-contributing-to-the-docs

Please re-submit with edits to ipynb instead. Thanks.

PR #1860, opened 2019-03-22
Adding A New Kind of Data Augmentation for Images

rsomani95:
By randomising one of the three RGB channels, the `learner` essentially sees the same image, but with different colors. I've implemented this as a data augmentation called `rgb_randomise()` in `vision/transform.py` and explained the output and parameters in `docs_src/vision.transform.ipynb`.

[forum link](https://forums.fast.ai/t/rgb-transformations-for-data-augmentation/36876/6)


sgugger:
Thanks! fastai speaks American English though, so I changed the name to randomize ;)
Will update the docs after the merge.

PR #1865, opened 2019-03-22
Update docstring with learning rate suggestions

sshleifer:
With --backwards True, IMDB classifier achieves 93.9% accuracy with lr=.005, vs. 92.6% with lr=.01,the default. 
This tripped me up so I figured it should be documented. I have only run half of same experiments for the forward model, scoring 94.5% with the default lr=.01.


sgugger:
Glad to know, thanks!

PR #1868, opened 2019-03-23
Make classes to labels consistent

abnera:
Right now when you get labels from_csv or df it infers the type meaning if I get the labels from a CSV file and the labels are all numeric values it will assume they are an integer if not it will assume they are strings. However, label_from_df always assume the labels are string even if all the folder names are numeric. In isolation, this those not cause any problem but if I try to create a DataBunch where I read the train is from a csv file and the validation is from folder there will be an inconsistency between the mapping of classes names and labels. That is because sorting numeric string is different than sorting integers. 

I know my case is very unique but I experienced this and others might 2. The idea is that both labels_from_df and label_from_folder provide the same mapping which they currently don't mainly because string sorted not the same as int sorting when creating the label name to label dictionary.

### DIFF
--- a/fastai/data_block.py
+++ b/fastai/data_block.py
@@ -287,7 +287,8 @@ class ItemList():
 
     def label_from_folder(self, label_cls:Callable=None, **kwargs)->'LabelList':
         "Give a label to each filename depending on its folder."
-        return self.label_from_func(func=lambda o: o.parts[-2], label_cls=label_cls, **kwargs)
+        labels_func = lambda o: int(o.parts[-2]) if o.parts[-2].isdigit() else o.parts[-2]
+        return self.label_from_func(func=labels_func, label_cls=label_cls, **kwargs)
 
     def label_from_re(self, pat:str, full_path:bool=False, label_cls:Callable=None, **kwargs)->'LabelList':
         "Apply the re in `pat` to determine the label of every filename.  If `full_path`, search in the full name."


sgugger:
This is making the tests fail so it's breaking something. I'm not sure I understand when you have ints as label? The API is designed to have them be strings all the time, so I'd be curious where you get them, and this is probably the part that needs fixing.

abnera:
A bit too early to be closing the PR...

Both `label_from_folder` and `label_from_df` work as expected independently. However, for my specific case, I am using the data block API from different datasets to aggregate them as I explained in the forum to you. An example I create a training dataset using csv file and validation dataset from folder. This is where the methods are inconsistent which each other. Internally `labels_from_folder` convers the folders names o.parts[-2] as strings and then sorts this strings to create a idx label to class name mapping.`labels_from_df` on the other hand does not enforce the column of labels to be string. In my case the column is full of integer values. Hence internally they are sorted as int to then create the i2c mapping. Sadly in python sorting string of numbers and sorting ints of numbers don't produce the same order. Example strings are sorted as 010 going before 002. This inconsistency prohibits me and possible others to use the data block API to interchangeably mix from csv with from folder, cuz they both have different behavior. For my case, the mapping of the training is different than the one for the validation even if they all have the same class names. This is because df treats them as ints and  from_folder treats them as strings. To solve this I created this PR where from_folder is converted into ints when it can.  Another solution is too force the `labels_from_df` to be converted into strings, cuz right now the type is infer.

Any of this two solutions should fix it. I can assure you mine has it's intended behaviour as I am using it in my code. Without it it will train but the mapping of train and val will be wrong and it will not learn.

sgugger:
I'm sorry if I was unclear: as you say there are two ways to fix this, having the ints be ints whenever possible or casting them to strings. The first one breaks the tests (think of a situation when one label is 0, the other one is seven) so I was indicating I'd prefer the second way, which is why I closed the PR. It doesn't prevent us discussing it further like this.

You are more than welcome to suggest a new PR that ensures everything is always cast to strings, which would solve your problems without breaking the library (make sure to run the tests if possible, before pushing your PR).

abnera:
You are right my solution only works for my issue, not a good solution. I will fix the `labels_from_df` to force into string always.

PR #1876, opened 2019-03-24
Update imdb.ipynb, Redundant Parameters.

PranshuVinayak:
bs and bptt are not used in LanguageModelData.


sgugger:
That is correct. Thanks for fixing!

PR #1877, opened 2019-03-25
plot_metrics: add skip_start, skip_end params

tchklovski:
`Recorder`'s `plot_metrics`, like `plot_losses`, now supports  `skip_start`, `skip_end` params


sgugger:
Looking good, thanks!

tchklovski:
🙏

PR #1884, opened 2019-03-26
test learner create body and head

yyoon62:
I wrote a test for creating body and head from the pretrained models. For the body, I'm checking if the body outputs the correct activation values using hooks. It's a bit lengthy, so let me know if there is a more elegant way to write this!

### DIFF
--- a/tests/test_vision_learner.py
+++ b/tests/test_vision_learner.py
@@ -1,7 +1,9 @@
 import pytest
 import torch
+import torch.nn as nn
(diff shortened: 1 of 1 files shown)


sgugger:
With have an `HookOutput` class (in callbacks.hook) that could be useful here and make it a bit shorter.

yyoon62:
Thank you for the suggestion, I rewrote the hooks with `Hooks` class.

sgugger:
Thanks!

PR #1885, opened 2019-03-26
add PooledSelfAttention

Separius:
The previous implementation was based on [Self-Attention Generative Adversarial Networks](https://arxiv.org/abs/1805.08318) and the authors didn't mention max_pooling and the output_convolution in the paper but [their code](https://github.com/brain-research/self-attention-gan/blob/master/non_local.py) has them.

note that [BigGAN's unofficial implementation from the official authors](https://github.com/ajbrock/BigGAN-PyTorch/blob/master/layers.py#L144) and [compare_gan](https://github.com/google/compare_gan/blob/master/compare_gan/architectures/arch_ops.py)(from google) both have the same architecture

### DIFF
--- a/fastai/layers.py
+++ b/fastai/layers.py
@@ -76,19 +76,28 @@ class SelfAttention(nn.Module):
     "Self attention layer for 2d."
     def __init__(self, n_channels:int):
         super().__init__()
-        self.query = conv1d(n_channels, n_channels//8)
-        self.key   = conv1d(n_channels, n_channels//8)
-        self.value = conv1d(n_channels, n_channels)
+        self.theta = spectral_norm(conv2d(n_channels, n_channels//8, 1))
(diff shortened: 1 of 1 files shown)


sgugger:
Those are quite big changes, so I think we should have a new layer for it. I know people were using the current SelfAttention layer and I don't want to break anything for them. Since you have the spectral norm inside we can maybe call it SpectralNormedSelfAttention? Or something like that?

Separius:
```conv1d``` (the old code) also applies spectral norm

But I totally agree that SelfAttention is not a great name for this layer and maybe ```NonLocal``` or ```PooledSelfAttention``` are better

sgugger:
Oh yeah, I had forgotten about that. `PooledSelfAttention` seems like a great name! Thanks for this contribution, will merge once the tests are clear.

PR #1890, opened 2019-03-27
Add test for mae, mse, rmse and msle

florisdf:
Tests for mean absolute error, mean square error, root mean square error and mean square log error


stas00:
Thank you, @florisdf!

PR #1893, opened 2019-03-27
Serialization / Deserialization of Fastai objects to byte streams

bachsh:
Added options to save/export/load using `BytesIO` streams to the following functions: `Learn.save`, `Learn.export`, `load_learner`, `DataBunch.save`, `load_data`.

Following a discussion with @sgugger [here](https://forums.fast.ai/t/serialize-learner-to-string/40848).

### DIFF
--- a/fastai/basic_data.py
+++ b/fastai/basic_data.py
@@ -146,12 +146,12 @@ class DataBunch():
     def remove_tfm(self,tfm:Callable)->None:
         for dl in self.dls: dl.remove_tfm(tfm)
 
-    def save(self, fname:PathOrStr='data_save.pkl')->None:
-        "Save the `DataBunch` in `self.path/fname`."
+    def save(self, fname:PathOrStr='data_save.pkl', buffer:io.BytesIO=None)->None:
+        "Save the `DataBunch` in `self.path/fname`. If buffer is specified, save to buffer"
         if not getattr(self, 'label_list', False):
             warn("Serializing the `DataBunch` only works when you created it using the data block API.")
             return
-        try_save(self.label_list, self.path, fname)
+        try_save(self.label_list, self.path, fname, buffer)
 
     def add_test(self, items:Iterator, label:Any=None)->None:
         "Add the `items` as a test set. Pass along `label` otherwise label them with `EmptyLabel`."
@@ -271,8 +271,10 @@ class DataBunch():
 
 def load_data(path:PathOrStr, fname:str='data_save.pkl', bs:int=64, val_bs:int=None, num_workers:int=defaults.cpus,
                   dl_tfms:Optional[Collection[Callable]]=None, device:torch.device=None, collate_fn:Callable=data_collate,
-                  no_check:bool=False, **kwargs)->DataBunch:
-    "Load from `path/fname` a saved `DataBunch`."
-    ll = torch.load(Path(path)/fname, map_location='cpu') if defaults.device == torch.device('cpu') else torch.load(Path(path)/fname)
+                  no_check:bool=False, buffer:io.BytesIO=None, **kwargs)->DataBunch:
+    """Load from `path/fname` a saved `DataBunch`.
+    If `buffer` is specified, load from it (`path` is still required to set the working directory)"""
+    target = buffer if buffer else Path(path)/fname
+    ll = torch.load(target, map_location='cpu') if defaults.device == torch.device('cpu') else torch.load(target)
     return ll.databunch(path=path, bs=bs, val_bs=val_bs, num_workers=num_workers, dl_tfms=dl_tfms, device=device,
                         collate_fn=collate_fn, no_check=no_check, **kwargs)
--- a/fastai/basic_data.py
+++ b/fastai/basic_data.py
@@ -146,12 +146,12 @@ class DataBunch():
     def remove_tfm(self,tfm:Callable)->None:
         for dl in self.dls: dl.remove_tfm(tfm)
 
-    def save(self, fname:PathOrStr='data_save.pkl', buffer:io.BytesIO=None)->None:
-        "Save the `DataBunch` in `self.path/fname`. If buffer is specified, save to buffer"
+    def save(self, file:PathLikeOrBinaryStream= 'data_save.pkl')->None:
+        "Save the `DataBunch` in `self.path/file`. If `file` is a binary stream, save to it."
         if not getattr(self, 'label_list', False):
             warn("Serializing the `DataBunch` only works when you created it using the data block API.")
             return
-        try_save(self.label_list, self.path, fname, buffer)
+        try_save(self.label_list, self.path, file)
 
     def add_test(self, items:Iterator, label:Any=None)->None:
         "Add the `items` as a test set. Pass along `label` otherwise label them with `EmptyLabel`."
@@ -269,12 +269,12 @@ class DataBunch():
             warn(message)
             print(final_message)
 
-def load_data(path:PathOrStr, fname:str='data_save.pkl', bs:int=64, val_bs:int=None, num_workers:int=defaults.cpus,
-                  dl_tfms:Optional[Collection[Callable]]=None, device:torch.device=None, collate_fn:Callable=data_collate,
-                  no_check:bool=False, buffer:io.BytesIO=None, **kwargs)->DataBunch:
-    """Load from `path/fname` a saved `DataBunch`.
-    If `buffer` is specified, load from it (`path` is still required to set the working directory)"""
-    target = buffer if buffer else Path(path)/fname
-    ll = torch.load(target, map_location='cpu') if defaults.device == torch.device('cpu') else torch.load(target)
+def load_data(path:PathOrStr, file:PathLikeOrBinaryStream= 'data_save.pkl', bs:int=64, val_bs:int=None, num_workers:int=defaults.cpus,
+              dl_tfms:Optional[Collection[Callable]]=None, device:torch.device=None, collate_fn:Callable=data_collate,
+              no_check:bool=False, **kwargs)->DataBunch:
+    """Load from `path/file` a saved `DataBunch`.
+    If `file` is a binary stream (file or buffer), read from it (`path` is still required to set the working directory)."""
+    source = Path(path)/file if is_pathlike(file) else file
+    ll = torch.load(source, map_location='cpu') if defaults.device == torch.device('cpu') else torch.load(source)
     return ll.databunch(path=path, bs=bs, val_bs=val_bs, num_workers=num_workers, dl_tfms=dl_tfms, device=device,
                         collate_fn=collate_fn, no_check=no_check, **kwargs)
--- a/fastai/basic_data.py
+++ b/fastai/basic_data.py
@@ -147,7 +147,7 @@ class DataBunch():
         for dl in self.dls: dl.remove_tfm(tfm)
 
     def save(self, file:PathLikeOrBinaryStream= 'data_save.pkl')->None:
-        "Save the `DataBunch` in `self.path/file`. If `file` is a binary stream, save to it."
+        "Save the `DataBunch` in `self.path/file`. `file` can be file-like (file or buffer)"
         if not getattr(self, 'label_list', False):
             warn("Serializing the `DataBunch` only works when you created it using the data block API.")
             return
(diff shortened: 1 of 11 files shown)


stas00:
Any reason why we need a new arg and not just check whether `dest` is an instance of str or pathlib or io.BytesIO and act accordingly?

Except, I guess the current one is called `fname`, so it doesn't lend itself for various types of destinations. Given that most people don't set the arg name and just use the ordered arg, perhaps rename it to `target` or `dest` or `where`?

sgugger:
We had that discussion on the forum, I was the one who asked for a different name since some users use fname (me for instance ;) )

stas00:
Yes, but we can rename it to something more generic.

The proposed change for example makes:

...`name:PathOrStr=None, return_path:bool=False, with_opt:bool=True, buffer:io.BytesIO=None)`

with 2 similar arguments being far apart, at the very least put them next to each other?

I get lost when sometimes you guys are totally OK with changing and renaming APIs and at other times bolting things on top, without improving the API.

bachsh:
Honestly, I think using the same argument is preferable. It's consistent with how `torch.save` works. I implemented it the way you asked for, Sylvain, but if you change your mind please lmk and I'll refactor.

sgugger:
Trying to avoid unnecessary changes but if you want to break everything, be my guest ;) Let's rename it source then?

stas00:
Perhaps for now overload `fname` and then before the major API change after part2 rename it to something more generic when a lot of things will be renamed anyway.

bachsh:
+1 for `source`.
I don't like overloading `fname`, even in the interim. We can bump the minor version to 1.1.0 to make this breaking change noticeable, WDYT?

stas00:
it's source for load, but dest for save - `save(source)` reads weird, no?

bachsh:
Sure, `source` for loading functions, and `target` or `dest` for `save`/`export`.

stas00:
> Sure, source for loading functions, and target or dest for save/export.

That'd be confusing if you think of it from the perspective of the user code.

let's mimic torch.save/torch.load and call it `f`. That's the easiest IMHO.

bachsh:
> let's mimic torch.save/torch.load and call it `f`. That's the easiest IMHO.

I don't like this name which implies file, but I'm ok with it.

Naming suggestions for the new Type: `Union[Path, str, BytesIO]`?

stas00:
> +1 for `source`.
> I don't like overloading `fname`, even in the interim. We can bump the minor version to 1.1.0 to make this breaking change noticeable, WDYT?

If we were to bump the minor version for all the API changes we would have been at 1.99.0 now. ;)

note, I've been updating one earlier comment of mine with different conventions in major python libs.

stas00:
My vote is to keep fname for now, adding the io support to it. and schedule a rename to a more suitable name once part2 is over - 5 weeks from now, when there will be a huge number of breaking changes.

Pros:

    very minor change to the codebase and docs
    not breaking user's code
    support new functionality

Cons:

    unintuitive arg name for the next 5 weeks.

sgugger:
If we break something, I'd rather do it now than later.

stas00:
I don't follow your logic. But I'm OK with it, everybody is used to breaking changes by now.

So we just need to decide on the new arg name then. I listed all the existing options I found here https://github.com/fastai/fastai/pull/1893#issuecomment-477379507, I'm not attached to any of them as long as they are symmetrical (i.e. same arg name for save/load).

sgugger:
I'm voting for `file`.

stas00:
So it looks like we have come an agreement to not add a new arg but to overload `fname` arg and rename it to `file`. I hope that's OK with you, @bachsh. Apologies for conflicting messages - I personally wasn't aware of that prior discussion at the forums.

So we will now have like numpy and pickle:

`file`: file-like object, string, or pathlib.Path

bachsh:
No need to apologize, it's best to have this decision after a thorough discussion.

I hope to get to this later today, I'm still contemplating on the name for the Type `Union[PathOrStr, BytesIO]`. WDYT?

stas00:
> No need to apologize, it's best to have this decision after a thorough discussion.
> 
> I hope to get to this later today, I'm still contemplating on the name for the Type `Union[PathOrStr, BytesIO]`. WDYT?

That works. 

Or alternatively, define:

`PathOrStrOrBytesIO = Union[Path,str,BytesIO]` in `fastai/core.py`.

bachsh:
Done. Updated the tests accordingly. Ended up calling the new type:
`PathLikeOrBinaryStream = Union[PathOrStr, BufferedWriter, BytesIO]`

fastai/basic_data.py
```diff
@@ -269,10 +269,12 @@ def sanity_check(self):
...
 
-def load_data(path:PathOrStr, fname:str='data_save.pkl', bs:int=64, val_bs:int=None, num_workers:int=defaults.cpus,
-                  dl_tfms:Optional[Collection[Callable]]=None, device:torch.device=None, collate_fn:Callable=data_collate,
-                  no_check:bool=False, **kwargs)->DataBunch:
-    "Load from `path/fname` a saved `DataBunch`."
-    ll = torch.load(Path(path)/fname, map_location='cpu') if defaults.device == torch.device('cpu') else torch.load(Path(path)/fname)
+def load_data(path:PathOrStr, file:PathLikeOrBinaryStream= 'data_save.pkl', bs:int=64, val_bs:int=None, num_workers:int=defaults.cpus,
+              dl_tfms:Optional[Collection[Callable]]=None, device:torch.device=None, collate_fn:Callable=data_collate,
+              no_check:bool=False, **kwargs)->DataBunch:
+    """Load from `path/file` a saved `DataBunch`.
```
stas00: fastai style is only one shortish line "about" docstring with the rest of the details inside the doc notebooks.

fastai/basic_train.py
```diff
@@ -237,31 +238,34 @@ def export(self, fname:PathOrStr='export.pkl', destroy=False):
...
-    def save(self, name:PathOrStr, return_path:bool=False, with_opt:bool=True):
-        "Save model and optimizer state (if `with_opt`) with `name` to `self.model_dir`."
-        self._test_writeable_path()
+    def save(self, file:PathLikeOrBinaryStream=None, return_path:bool=False, with_opt:bool=True):
+        """Save model and optimizer state (if `with_opt`) with `file` to `self.model_dir`.
+        If `file` is a binary stream (buffer or file), write to it."""
+        if is_pathlike(file): self._test_writeable_path()
         if rank_distrib(): return # don't save if slave proc
-        path = self.path/self.model_dir/f'{name}.pth'
+        target = self.path/self.model_dir/f'{file}.pth' if is_pathlike(file) else file
```
stas00: do we want to validate that file is actually a buffer if it's not str or pathlib?

fastai/basic_train.py
```diff
@@ -237,31 +238,34 @@ def export(self, fname:PathOrStr='export.pkl', destroy=False):
...
         "Return DataLoader for DatasetType `ds_type`."
         return self.data.dl(ds_type)
 
-    def load(self, name:PathOrStr, device:torch.device=None, strict:bool=True, with_opt:bool=None, purge:bool=True,
-            remove_module:bool=False):
-        "Load model and optimizer state (if `with_opt`) `name` from `self.model_dir` using `device`."
+    def load(self, file:PathLikeOrBinaryStream=None, device:torch.device=None, strict:bool=True,
+             with_opt:bool=None, purge:bool=True, remove_module:bool=False):
+        """Load model and optimizer state (if `with_opt`) `file` from `self.model_dir` using `device`.
+        If `file` is a binary stream (buffer or file), read from it."""
```
stas00: as the comment above. Probably just make it brief `file can be a buffer` or something like that.

fastai/torch_core.py
```diff
@@ -403,8 +403,9 @@ def add_metrics(last_metrics:Collection[Rank0Tensor], mets:Union[Rank0Tensor, Co
     last_metrics,mets = listify(last_metrics),listify(mets)
     return {'last_metrics': last_metrics + mets}
 
-def try_save(state:Dict, path:Path, fname:PathOrStr):
-    try: torch.save(state, open(path/fname, 'wb'))
+def try_save(state:Dict, path:Path=None, file:PathLikeOrBinaryStream=None):
+    target = open(path / file, 'wb') if is_pathlike(file) else file
```
stas00: fastai uses `foo/bar` style and not `foo / bar` - so it mimics the actual file path. here and below.

fastai/basic_train.py
bachsh: Well, Python is not strongly typed so `typing` doesn't really take care of it (IDEs will issue a warning).
I followed the `torch` convention here where they check if it's path-like and assume file-like otherwise.

fastai/torch_core.py
bachsh: TBH it was like that before I got here. Will fix :)

fastai/basic_train.py
stas00: we could be more defensive by perhaps validating that `file` arg is one of the 4 types as the first thing in the function. But, perhaps, it's highly unlikely this will happen.

And I'm not sure @sgugger will agree as it'll add an extra line in several functions. So probably it's just fine for now.

stas00:
Just a few small suggestions, but otherwise looks great, @bachsh

bachsh:
> Just a few small suggestions, but otherwise looks great, @bachsh

Done.

fastai/basic_train.py
bachsh: Honestly, I don't it's worth the trouble. We could be defensive all over the place, not just here. Recall the following line from the "how to contribute" guide:
> Contributing to Parkinson’s law of triviality has negative consequences for a project. Let’s focus on deep learning!

bachsh:
@stas00  and @sgugger  thanks for being super responsive!

stas00:
The code + tests looks good, @bachsh. Thank you.

Perhaps add the doc change into this same PR and we will merge it as one change? If you're comfortable updating the docs that is.

bachsh:
> Perhaps add the doc change into this same PR and we will merge it as one change? If you're comfortable updating the docs that is.

I updated the docstrings. What other docs need changing?

> 
> might need to do some grep'ing for the modified functions to catch any examples in docs where `fname` is used (but shouldn't be many instances if any). Also under `examples/`.

Found a few instances and eliminated them.

> 
> Oh and please add a new entry into CHANGES.md

Thanks, done.

stas00:
the doc nbs need to be updated. I pasted the link with instructions earlier:  https://docs.fast.ai/gen_doc_main.html#process-for-contributing-to-the-docs

stas00:
> > might need to do some grep'ing for the modified functions to catch any examples in docs where `fname` is used (but shouldn't be many instances if any). Also under `examples/`.
> 
> Found a few instances and eliminated them.

I didn't see the commit, unless it was earlier already, right?

bachsh:
> the doc nbs need to be updated. I pasted the link with instructions earlier: https://docs.fast.ai/gen_doc_main.html#process-for-contributing-to-the-docs

Not sure what to update. The docs should update automatically when you generate them, shouldn't they? 

> > > might need to do some grep'ing for the modified functions to catch any examples in docs where `fname` is used (but shouldn't be many instances if any). Also under `examples/`.
> > 
> > 
> > Found a few instances and eliminated them.
> 
> I didn't see the commit, unless it was earlier already, right?

These examples where in the docs (`basic_train.ipynb` and `text.ipynb`)

fastai/basic_train.py
sgugger: I think the error message PyTorch will throw if `file` is of the wrong type so it's no necessary to worry here.

sgugger:
I think the rest of the places where it's in the docs will be updated during a rebuild yes. Merging. Will follow up with two more details: the import in core should be in imports.core and then the new type alias needs to be in the dictionary for pretty printing in the docs.

sgugger:
All done, thanks a lot for your work @bachsh !
Here is the [follow-up commit](https://github.com/fastai/fastai/commit/7d0ca5152c0af43dd195988cdc7efa0c772bd1db)

bachsh:
> I think the rest of the places where it's in the docs will be updated during a rebuild yes. Merging. Will follow up with two more details: the import in core should be in imports.core and then the new type alias needs to be in the dictionary for pretty printing in the docs.

Will keep that in mind for next time, thanks!

> All done, thanks a lot for your work @bachsh !

It's been a real pleasure!

PR #1901, opened 2019-03-30
improve optimizer excel with name variable

noklam:
So I just learned about you can actually name an excel cell directly, which I think it improves readability and reduces the eyeball lookup movement.

I have only edit the "data" and "basic SGD" page as demo, if this is useful, I can continue with changing the whole excel file. The VBA functions works fine as well.

image
image


jph00:
Sorry we sat on this so long! I agree named ranges are the right way to do this, but for teaching I think many people will not understand what they're seeing. So I think it's best we stick with range addresses for our teaching workbooks.

Issue #1898, opened 2019-03-29
[Feature Request] Datasets config


tydlwav:
**Feature Request**

Separate configs in config.yml for downloaded data and untar destination. 

**Justification**

I am running on my own desktop machine and I have an SSD for fast file storage, but I use a harddrive for archiving purposes and files that I don't need all the time. To not use too much space on my SSD, I usually save my data on the HDD as a compressed file and then decompress them onto my SSD when I need them. This way, I can easily delete the whole data directory when I need to free up space. 

To achieve this purpose, I need to set the config.yml data directory to my HDD and then specifically declare dest="SOMEWHERE/ON/MY/SSD" when running notebooks. This is not so convenient and the trivial issue should be solved with less than 10 lines of code to get a quality of life improvement. 

**I'm willing to help**

Hopefully you guys can consider this. I'm willing to fork, change the code and docs and submit a pull request if you guys approve of this.

stas00:
Looks like a useful feature to me. I think your situation would be quite common for large datasets.

Needless to say, but the default should remain as it is now (shared folder).

But let's wait for @sgugger to double check before you proceed. Thank you, @tydlwav

sgugger:
Hi there. This could be a useful feature for some users, yes. It's not going to be on our priority list for now but if you're willing to tackle it, we'll gladly look at a PR. I'm closing this here as issues are for standing bugs only, but I've created a topic on the forum to continue the discussion if needed.

tydlwav:
Sure. I'll look into writing an implementation and adding it to the docs. I'll shoot up a PR when I'm done.

PR #1904, opened 2019-03-31
Adds archive config to datasets

tydlwav:
New config file structure with `data_archive_path` for tgz compressed archives, `data_path` for uncompressed dataset folders and `model_path` untouched. Could add a `model_archive_path` for the future, but since I've never used `model_path`, I'm leaving it out of this PR. 

New defaults:
```
FASTAI_HOME
|- config.yml
|- data
|    |- archive
|    |      \ COMPRESSED
|     \ DECOMPRESSED
\ models
```
Solves #1898

### DIFF
--- a/tests/test_vision_data.py
+++ b/tests/test_vision_data.py
@@ -165,7 +165,7 @@ responses = try_import('responses')
 def test_trunc_download():
     this_tests(untar_data)
     url = URLs.COCO_TINY
-    fname = datapath4file(f'{url2name(url)}.tgz')
+    fname = datapath4file(url2name(url)).with_suffix(".tgz")
     # backup user's current state
     fname_bak = fname.parent/f"{fname.name}-bak"
     if fname.exists(): os.rename(fname, fname_bak)


tydlwav:
@sgugger 
:laughing: :laughing: :laughing:  
:dancer: :dancer: :dancer:

tydlwav:
Hmm, wondering why the checks are not passing.

stas00:
> Hmm, wondering why the checks are not passing.

Because you didn't run the test suite and one of the tests now fails. Click on `Details` and then go to `View more details on Azure Pipelines` at the end of the page which takes you to 
https://dev.azure.com/fastdotai/fastai/_build/results?buildId=4284
then click on the failing job and you will see:
```
=================================== FAILURES ===================================
_____________________________ test_trunc_download ______________________________

    @pytest.mark.skipif(not responses, reason="requires the `responses` module")
    def test_trunc_download():
        this_tests(untar_data)
        url = URLs.COCO_TINY
>       fname = datapath4file(f'{url2name(url)}.tgz')
E       TypeError: datapath4file() missing 1 required positional argument: 'archive'

tests/test_vision_data.py:168: TypeError
-------------- generated xml file: /home/vsts/work/1/s/result.xml --------------
```

fastai/datasets.py
```diff
@@ -180,20 +186,22 @@ def url2name(url): return url.split('/')[-1]
...
+    return datapath4file(f'{url2name(url)}{ext}', ext=ext, archive=True) if data else modelpath4file(f'{url2name(url)}{ext}', ext=ext)
 
 def modelpath4file(filename, ext:str='.tgz'):
     "Return model path to `filename`, checking locally first then in the config file."
     local_path = URLs.LOCAL_PATH/'models'/filename
     if local_path.exists() or local_path.with_suffix(ext).exists(): return local_path
     else: return Config.model_path()/filename
 
-def datapath4file(filename, ext:str='.tgz'):
+def datapath4file(filename, archive:bool, ext:str='.tgz'):
```
sgugger: This is the line causing problems in the tests (so potentially for users that use this function directly). Please add a default to that argument consistent with the current behavior (`archive=False`).

tydlwav:
Hmm. This should fix it. Had to modify the test as somebody wrote a test using `datapath4file` intended as the archived version.

tests/test_vision_data.py
```diff
@@ -165,7 +165,7 @@ def test_download_images():
 def test_trunc_download():
     this_tests(untar_data)
     url = URLs.COCO_TINY
-    fname = datapath4file(f'{url2name(url)}.tgz')
+    fname = datapath4file(f'{url2name(url)}.tgz', archive=True)
```
stas00: If this feature didn't exist, how can it be that the test that didn't not require it, now requires it?

fastai/datasets.py
```diff
@@ -132,6 +132,7 @@ class Config():
     DEFAULT_CONFIG_PATH = DEFAULT_CONFIG_LOCATION + '/config.yml'
     DEFAULT_CONFIG = {
         'data_path': DEFAULT_CONFIG_LOCATION + '/data',
+        'data_archive_path': DEFAULT_CONFIG_LOCATION + '/data/archive',
```
stas00: I thought that we agreed that the default should remain as it was. And letting the user to override it. This is a different implementation.

fastai/datasets.py
```diff
@@ -180,20 +186,22 @@ def url2name(url): return url.split('/')[-1]
 def url2path(url, data=True, ext:str='.tgz'):
     "Change `url` to a path."
     name = url2name(url)
-    return datapath4file(name, ext=ext) if data else modelpath4file(name, ext=ext)
-def _url2tgz(url, data=True, ext:str='.tgz'):
-    return datapath4file(f'{url2name(url)}{ext}', ext=ext) if data else modelpath4file(f'{url2name(url)}{ext}', ext=ext)
+    return datapath4file(name, ext=ext, archive=False) if data else modelpath4file(name, ext=ext)
+def _url2tgz(url, data=True, ext: str = '.tgz'):
```
stas00: please restore the original code here - vertical and horizontal whitespace.

fastai/datasets.py
```diff
@@ -180,20 +186,22 @@ def url2name(url): return url.split('/')[-1]
...
+    return datapath4file(f'{url2name(url)}{ext}', ext=ext, archive=True) if data else modelpath4file(f'{url2name(url)}{ext}', ext=ext)
 
 def modelpath4file(filename, ext:str='.tgz'):
     "Return model path to `filename`, checking locally first then in the config file."
     local_path = URLs.LOCAL_PATH/'models'/filename
     if local_path.exists() or local_path.with_suffix(ext).exists(): return local_path
     else: return Config.model_path()/filename
 
-def datapath4file(filename, ext:str='.tgz'):
+def datapath4file(filename, archive:bool=False, ext:str='.tgz'):
```
stas00: why the new arg is added in the middle - this will break the API for those who use positional args.

stas00:
Thinking more about it, your proposition is trying to solve only a partial problem - you want the "./archive" sub-folder to be a symlink to a bigger drive, so that all downloaded `tgz` get dumped there. But this is not flexible enough, e.g. I don't store large datasets in one flat folder, they are organized in a hierarchical structure and multiple sub-folders.

So ideally depending on the dataset I will want to set the archive folder to a unique path in each specific notebook. 

And then like in your case, some users will want to set the default for archive folder to be global, so it goes into another drive. And they should be able to do so by tweaking the config object once and forget about it.

Hardcoding `./archive` as a default is not super convenient, since you can't just untar files from the console if you need to - you now need to either remember to move them or pass the flag to untar at a different destination.

That's why I suggest that there is no need to change the API, other than making the code use the archive setting from the Config object and let the user set it to whatever she wants it to be.

Does it make sense?

fastai/datasets.py
tydlwav: OK. I'll change this to fit the specs.

fastai/datasets.py
tydlwav: You're right.

fastai/datasets.py
tydlwav: Sorry, think I accidentally did this.

tests/test_vision_data.py
tydlwav: So in the old implementation, `datapath4file` is used to both get the path for decompressed datasets and the compressed dataset file, since they are strictly in the same path if the user does not specify `fname` in `download_data` or `dest` in `untar_data`. This will cause an issue if we separate archived files and decompressed files in two different paths, so we need an API change. 

In this particular case, the test is looking for the archive path to compare checksum, so I need to add archive=True to the end.

tydlwav:
@stas00 Coming to think of it, I agree. It's a better idea to set the defaults of data_path and data_archive_path to the same place. As to hierarchical organization, I don't see any way to easily add per dataset specifications in the config. It could potentially require scrapping a lot of functions in the module, but I haven't thoroughly thought over it. We would probably need to finish the TODOs in the module ("coded more shortly and nicely" and "simplify this mess") before we attempt that.

tydlwav:
> @stas00 Coming to think of it, I agree. It's a better idea to set the defaults of data_path and data_archive_path to the same place. As to hierarchical organization, I don't see any way to easily add per dataset specifications in the config. It could potentially require scrapping a lot of functions in the module, but I haven't thoroughly thought over it. We would probably need to finish the TODOs in the module ("coded more shortly and nicely" and "simplify this mess") before we attempt that.

If you guys are willing to write me some specifications and guidelines, I can attempt to clean it up and add per dataset settings to the config file.

stas00:
> @stas00 Coming to think of it, I agree. It's a better idea to set the defaults of data_path and data_archive_path to the same place. 

Great!

> As to hierarchical organization, I don't see any way to easily add per dataset specifications in the config. It could potentially require scrapping a lot of functions in the module, but I haven't thoroughly thought over it. We would probably need to finish the TODOs in the module ("coded more shortly and nicely" and "simplify this mess") before we attempt that.

I think I didn't communicate clearly, I was just talking about various ways this can be used and not saying we need to accomplish this in this PR. All I was trying to say is that once we have the archive setting, then a user will be able to override it globally (because their fast drive is small - this PR), and then a user should be able to override it just for a specific notebook - need to create API (another PR). The hierarchy of the storage does't matter, I was just trying to make a point that not everybody will have all .tgz in one flat folder.

So let's complete this PR with just that - giving a user a way to override the defaults.

stas00:
continuing from the code comment above. Unrelated to this PR:
```
def datapath4file(filename, ext:str='.tgz'):
    "Return data path to `filename`, checking locally first then in the config file."
    local_path = URLs.LOCAL_PATH/'data'/filename
    if local_path.exists() or local_path.with_suffix(ext).exists(): return local_path
    else: return Config.data_path()/filename
```
currently this needs some work, as 

1. it's quite ambiguous since the addition of `ext` arg.

So the test calls `fname = datapath4file(f'{url2name(url)}.tgz')` and gets the filename and not the path, because double exists() above doesn't check that the `filename` already includes `tgz` and returns a path that has `tgz` in it, which is wrong. 

So first the test should go:

```    
fname = f"{ datapath4file(url2name(url)}) }.tgz"
```

2. It doesn't test that `Config.data_path()/filename` actually exists

Now you're saying that it needs to look for it under `Config.data_archive_path()/filename` - why do we need an extra argument for that?

tydlwav:
Sorry I think I used the master branch of the fork. I'll create a new branch and a new PR that refers back here. :cry:

PR #1905, opened 2019-03-31
Continuing #1904

tydlwav:
Continuing #1904

### DIFF
--- a/fastai/datasets.py
+++ b/fastai/datasets.py
@@ -132,6 +132,7 @@ class Config():
     DEFAULT_CONFIG_PATH = DEFAULT_CONFIG_LOCATION + '/config.yml'
     DEFAULT_CONFIG = {
         'data_path': DEFAULT_CONFIG_LOCATION + '/data',
+        'data_archive_path': DEFAULT_CONFIG_LOCATION + '/data',
         'model_path': DEFAULT_CONFIG_LOCATION + '/models'
     }
 
@@ -149,6 +150,11 @@ class Config():
     def data_path(cls):
         "Get the path to data in the config file."
         return cls.get_path('data_path')
+
+    @classmethod
+    def data_archive_path(cls):
+        "Get the path to data archives in the config file."
+        return cls.get_path('data_archive_path')
 
     @classmethod
     def model_path(cls):
@@ -221,8 +227,8 @@ def untar_data(url:str, fname:PathOrStr=None, dest:PathOrStr=None, data=True, fo
         if dest.exists(): shutil.rmtree(dest)
     if not dest.exists():
         fname = download_data(url, fname=fname, data=data)
-        data_dir = Config().data_path()
+        data_archive_dir = Config().data_archive_path()
         if url in _checks:
-            assert _check_file(fname) == _checks[url], f"Downloaded file {fname} does not match checksum expected! Remove that file from {data_dir} and try your code again."
+            assert _check_file(fname) == _checks[url], f"Downloaded file {fname} does not match checksum expected! Remove that file from {data_archive_dir} and try your code again."
         tarfile.open(fname, 'r:gz').extractall(dest.parent)
     return dest
--- a/fastai/datasets.py
+++ b/fastai/datasets.py
@@ -200,8 +200,8 @@ def datapath4file(filename, ext:str='.tgz', archive=True):
     "Return data path to `filename`, checking locally first then in the config file."
     local_path = URLs.LOCAL_PATH/'data'/filename
     if local_path.exists() or local_path.with_suffix(ext).exists(): return local_path
-    elif archive: return Config.data_path() / filename
-    else: return Config.data_archive_path() / filename
+    elif archive: return Config.data_archive_path() / filename
+    else: return Config.data_path() / filename
 
 def download_data(url:str, fname:PathOrStr=None, data:bool=True, ext:str='.tgz') -> Path:
     "Download `url` to destination `fname`."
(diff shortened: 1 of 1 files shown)


sgugger:
Not sure I follow, this archive path isn't used anywhere so why is the data downloaded in a separate folder?

stas00:
Well, it's a mess because we have started talking about what the API should really be if we separate `data_path` and `data_archive_path`. So let's hold off on experimental changes and first define how we want it to work. Then implementing it is the easy part.

tydlwav:
I think @stas00 is trying to come up with a new function to replace `modelpath4file`, so I didn't modify that function. I have a modified version that passes the tests locally, so I can commit and push any time we need.

stas00:
So we have 2 ways the files are saved in on download:
`untar_data(url:str, fname:PathOrStr=None, dest:PathOrStr=None...`

* local path (user defined path)
* global path (via config object)

it should work just fine for local path, since a user can say:
`untar_data(url, "/download/here/file.tgz", "/untar/here/file")`

So we could have just stopped here and say, well, why can't you just use this existing feature, right?

So if you don't override the `fname` and `dest` args then you'd like to have these set by default to 2 different locations, correct? because you want this to be the default behavior and not needing to set it manually in all notebooks.

If so `untar_data` should set its fallback `dest` to that new `data_archive_path` and the code modified accordingly.

Are we on the same page?

stas00:
Looking more these features need a serious redesign (well it says TODO in the file)

```
def _url2tgz(url, data=True, ext:str='.tgz'):
    return datapath4file(f'{url2name(url)}{ext}', ext=ext) if data else modelpath4file(f'{url2name(url)}{ext}', ext=ext)
```
why are passing `ext` twice? I think it needs to be removed from the first arg.
```
def datapath4file(filename, ext:str='.tgz'):
    "Return data path to `filename`, checking locally first then in the config file."
    local_path = URLs.LOCAL_PATH/'data'/filename
    if local_path.exists() or local_path.with_suffix(ext).exists(): return local_path
    else: return Config.data_path()/filename
```
this is certainly is incorrect. You can check either or but return only one of them. It won't work if someone pass `filename` that includes tgz, like the test I have just fixed.

same applies to `modelpath4file`

sgugger:
The TODO is for after the course, we don't have time to redesign the whole datasets module and thoroughly test right now. i'd suggest focusing on the feature to add for now.

tydlwav:
> So we have 2 ways the files are saved in on download:
> `untar_data(url:str, fname:PathOrStr=None, dest:PathOrStr=None...`
> 
> * local path (user defined path)
> * global path (via config object)
> 
> it should work just fine for local path, since a user can say:
> `untar_data(url, "/download/here/file.tgz", "/untar/here/file")`
> 
> So we could have just stopped here and say, well, why can't you just use this existing feature, right?
> 
> So if you don't override the `fname` and `dest` args then you'd like to have these set by default to 2 different locations, correct? because you want this to be the default behavior and not needing to set it manually in all notebooks.
> 
> If so `untar_data` should set its `dest` when it's not set in the arguments to that new `data_archive_path` and the code modified accordingly.
> 
> Are we on the same page?

Yeah. It is a hassle to specify the exact same location for `dest` in every notebook.

tydlwav:
> The TODO is for after the course, we don't have time to redesign the whole datasets module and thoroughly test right now. i'd suggest focusing on the feature to add for now.

I have code that kinda adds to this mess, but fulfils the feature request. If you guys think it's fine, we can use that code first and then redesign the module altogether. I'd love to help with the redesign and implementation.

stas00:
If it's OK with @sgugger, it's fine with me if as long as it gets cleaned up/untangled after part2. i understand you're referring to the original PR https://github.com/fastai/fastai/pull/1904

tydlwav:
Cool, I'll push up my modified `modelpath4file` then.

stas00:
And @sgugger gives green light, please add a new test, so that when things get redesigned this feature won't break. Thanks.

tydlwav:
Yeah, those changes are in this new PR. I'll add a few tests when @sgugger green lights my code.

fastai/datasets.py
```diff
@@ -190,11 +196,12 @@ def modelpath4file(filename, ext:str='.tgz'):
     if local_path.exists() or local_path.with_suffix(ext).exists(): return local_path
     else: return Config.model_path()/filename
 
-def datapath4file(filename, ext:str='.tgz'):
+def datapath4file(filename, ext:str='.tgz', archive=True):
     "Return data path to `filename`, checking locally first then in the config file."
     local_path = URLs.LOCAL_PATH/'data'/filename
     if local_path.exists() or local_path.with_suffix(ext).exists(): return local_path
-    else: return Config.data_path()/filename
+    elif archive: return Config.data_path() / filename
```
stas00: should this be the other way around?

stas00:
Odd, github is not updating things - I have to force page reload to see changes.

tydlwav:
omg, my brain isn't working right, how did I miss that. Should've read through before committing.

stas00:
> omg, my brain isn't working right, how did I miss that. Should've read through before committing.

That's why we write tests first. Make it a habit and then you can give your brain a break :)

tydlwav:
Agreed. Gotta develop better software development habits. XD

sgugger:
Ok, looking good. Merging this then you can add a test in a separate PR.

stas00:
and documentation please. Without it only you know about this feature.

tydlwav:
I will do so during the week. Thanks for being patient and helping me through the process. :D

PR #1907, opened 2019-03-31
Improved performance of cont_cat_split using Pandas

antoinebon:


sgugger:
Thanks!

csteel45:
dtype == int does not work with Int64. Can you please add a '... or dtype == "Int64"'?
Thank you.

PR #1910, opened 2019-04-01
Disable progress context manager

bfarzin:
Adding a context manager in `fastai.utils` to disable display of the context bar in jupyter notebooks.  

When you are searching for a batchsize or trying to fit across several parameter sets this can suppress the updates so that you have a clean notebook.  In particular when you use `StopAfterNBatches` you can get red bars that can be alarming but are expected.

I also tried to add a notebook that shows usage.  

I was unsure where to put the code for this.  Happy to move to another location or combine in another file if that is a better fit.

Feedback always appreciated!

### DIFF
--- a/fastai/utils/mod_display.py
+++ b/fastai/utils/mod_display.py
@@ -1,10 +1,7 @@
-" Utils for modifying what is displayed in notbooks and command line"
+" Utils for modifying what is displayed in notebooks and command line"
 import fastai
 import fastprogress
 
-#from fastai.basic_train import Learner,Recorder
-#from fastai.core import copy,partial
-from fastai import *
 from ..basic_train import *
 from ..core import *
 
@@ -12,14 +9,17 @@ __all__ = ['progress_disabled']
 
 class progress_disabled():
     ''' Context manager to disable the progress update bar and Recorder print'''
-    def __init__(self,learner:Learner):
-        self.learn = learner
-        self.orig_callback_fns = copy(learner.callback_fns)
+    def __init__(self,learn:Learner):
+        self.learn = learn
+
     def __enter__(self):
         #silence progress bar
         fastprogress.fastprogress.NO_BAR = True
         fastai.basic_train.master_bar, fastai.basic_train.progress_bar = fastprogress.force_console_behavior()
-        self.learn.callback_fns[0] = partial(Recorder,add_time=True,silent=True) #silence recorder
+        self.orig_callback_fns = copy(self.learn.callback_fns)
+        rec_name = [x for x in self.learn.callback_fns if x.func == Recorder][0]
+        rec_idx = self.learn.callback_fns.index(rec_name)
+        self.learn.callback_fns[rec_idx] = partial(Recorder,add_time=True,silent=True) #silence recorder
 
         return self.learn
 
--- a/fastai/utils/mod_display.py
+++ b/fastai/utils/mod_display.py
@@ -1,0 +1,19 @@
+" Utils for modifying what is displayed in notbooks and command line"
+import fastai
+import fastprogress
+
+class progress_diabled():
+    ''' Context manager to disable the progress update bar and Recorder print'''
+    def __init__(self,learner:Learner):
+        self.learn = learner
+        self.orig_callback_fns = copy(learner.callback_fns)
+        def __enter__(self):
+            #silence progress bar
+            fastprogress.fastprogress.NO_BAR = True
+            fastai.basic_train.master_bar, fastai.basic_train.progress_bar = fastprogress.force_console_behavior()
+            self.learn.callback_fns[0] = partial(Recorder,add_time=True,silent=True) #silence recorder
+            return self.learn
+
+        def __exit__(self,type,value,traceback):
+            fastai.basic_train.master_bar, fastai.basic_train.progress_bar = master_bar,progress_bar
+            self.learn.callback_fns = self.orig_callback_fns
--- a/fastai/utils/mod_display.py
+++ b/fastai/utils/mod_display.py
@@ -5,9 +5,9 @@ import fastprogress
 from ..basic_train import *
 from ..core import *
 
-__all__ = ['progress_disabled']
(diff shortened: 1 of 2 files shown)


sgugger:
Note that on the last of fastprogress, it won't even show the total time (that part isn't printed anymore).

fastai/utils/mod_display.py
```diff
@@ -0,0 +1,28 @@
+" Utils for modifying what is displayed in notbooks and command line"
+import fastai
+import fastprogress
+
+#from fastai.basic_train import Learner,Recorder
```
sgugger: No need to leave commented imports.

fastai/utils/mod_display.py
```diff
@@ -0,0 +1,28 @@
+" Utils for modifying what is displayed in notbooks and command line"
```
sgugger: Little typo at notebooks.

fastai/utils/mod_display.py
```diff
@@ -0,0 +1,28 @@
...
+import fastai
+import fastprogress
+
+#from fastai.basic_train import Learner,Recorder
+#from fastai.core import copy,partial
+from fastai import *
```
sgugger: Why do you need this one?

fastai/utils/mod_display.py
```diff
@@ -0,0 +1,28 @@
...
+from ..basic_train import *
+from ..core import *
+
+__all__ = ['progress_disabled']
+
+class progress_disabled():
+    ''' Context manager to disable the progress update bar and Recorder print'''
+    def __init__(self,learner:Learner):
+        self.learn = learner
+        self.orig_callback_fns = copy(learner.callback_fns)
```
sgugger: I would move this line in __enter__ just to be sure.

fastai/utils/mod_display.py
```diff
@@ -0,0 +1,28 @@
...
+from fastai import *
+from ..basic_train import *
+from ..core import *
+
+__all__ = ['progress_disabled']
+
+class progress_disabled():
+    ''' Context manager to disable the progress update bar and Recorder print'''
+    def __init__(self,learner:Learner):
+        self.learn = learner
```
sgugger: Let's call everything `learn` to be consistent.

fastai/utils/mod_display.py
```diff
@@ -0,0 +1,28 @@
...
+        self.orig_callback_fns = copy(learner.callback_fns)
+    def __enter__(self):
+        #silence progress bar
+        fastprogress.fastprogress.NO_BAR = True
+        fastai.basic_train.master_bar, fastai.basic_train.progress_bar = fastprogress.force_console_behavior()
+        self.learn.callback_fns[0] = partial(Recorder,add_time=True,silent=True) #silence recorder
```
sgugger: Maybe a little loop over the callback_fns to find which one is the `Recorder`? It's supposed to be the first one, but who knows what the users will do.

sgugger:
It's looking good and useful! I have a few comments to clean this is a little bit before merging.

stas00:
And when you guys complete please also add an example to deploy that at
https://docs.fast.ai/callbacks.misc.html#StopAfterNBatches
editing `callbacks.misc.ipynb` https://docs.fast.ai/gen_doc_main.html#step-3-edit-the-documents

Thank you!

fastai/utils/mod_display.py
bfarzin: All great ideas.  Let me make these small changes and be back soon.

stas00:
Another thought for post-PR-merge is that there must be a better way to do it. A context manager requires you to go and change all the fit calls which defeats the purpose of `StopAfterNBatches` global callback, which you can turn on/off in one line of code.

So my preference would be a global setting for fastprogress that will do this for me from one place in the notebook.
so perhaps renaming your proposal to `progress_disabled_ctx` and adding `progress_enable`/`progress_disable` that will allow us to:
```
from fastai.callbacks.misc import StopAfterNBatches
from fastai.utils.mod_display import progress_enable, progress_disable
# True turns the speedup on, False return to normal behavior
tune = True # False
if tune:
    defaults.extra_callbacks = [StopAfterNBatches(n_batches=2)]
    progress_disable() # tell fastprogress not to print anything here
else:
    defaults.extra_callbacks = None
    progress_enable() 
```

Thank you!

bfarzin:
@stas00 I don't have a strong point of view. It could be easy to add into `StopAfterNBatches` so you have it in just one line.  Or, there could be a separate call that would flow as you indicate.

What I like about the context manager, is that after my "test" it goes back to normal.  And I won't forget to put it back if I am doing this as a smaller test in a bigger notebook or project.  For example, we might wan this to `do_the_magic()` and find the batch size or other params, then proceed as needed.  I see how that works in your use case as well with the `tune` condition.

If you and @sgugger think one way works better than another, I am happy to implement it!

stas00:
I'm not asking for any change, other than putting the enter/exit functionality into functions that could be used elsewhere.

In the case of `StopAfterNBatches` which you can enable globally or locally, the callback already acts as a context manager - we just need to add `on_train_begin`/`on_train_end` which would call your enter/exit funcs - so as soon as it's no longer configured for a given `learn` object or `fit` call, the behavior goes back to normal - i.e. with that change in `StopAfterNBatches` you won't need the context manager (if you're just using it with  `StopAfterNBatches` )

Renaming it to add `_ctx` is just an idea to tell those functions apart.

bfarzin:
Oh I see now.  I think adding _ctx is a good idea. I’ll do that.

stas00:
looking good - next, it'd be good to have a test to ensure this functionality doesn't break when things change.

bfarzin:
I agree.  I was just going to ask how to check the displayed output.  I’ll write that up tomorrow.  Thanks so much for all the feedback.

sgugger:
I'll merge this for now, waiting for your additional test. Thanks!

bfarzin:
Thanks!  I got a simple test out.  I will put out a new PR for that

PR #1911, opened 2019-04-01
Fix typo in distributed.md

yeldarby:
The docs were setting `args = parser.parse_args()` (plural) but trying to read it back as `arg.local_rank` (not plural).


sgugger:
Good catch, thanks for fixing!

PR #1912, opened 2019-04-02
Fix off-by-one error in LanguageLearner show_results

yeldarby:
As pointed out [on the forums here](https://forums.fast.ai/t/1-word-skipped-language-model-learner-show-results/42589), `LanguageLearner.show_results` is off by one word when displaying the `target` and `pred`. This happens because x and y are one word offset from each other coming from `one_batch` (which is ultimately pulling from `LanguageModelPreLoader.__getitem__`).

**Before (missing "1")**
image

**After (fixed)**
image


sgugger:
Thanks for fixing!

PR #1915, opened 2019-04-02
core function parallel returns results of processes if any

H4dr1en:
Hi,

In this PR I propose a minor change to the function parallel. Currently the function does not return the results of the parallel processes. This is a problem for the cases where such return values are required. 

Currently:
image

With the change:
image

I added a check so that the function parallel returns the results only if at least one process has returned a value.


H4dr1en:
I just found a bug with the current implementation. Please don't merge for now.

sgugger:
Just restyled for vertical space, let me know when this is ready.

H4dr1en:
The bug was not related to this part of code, so you can merge.

PR #1916, opened 2019-04-02
fixed return value for parallel for the case max_workers = 1

H4dr1en:
Nearly forgot it!

For the case max_workers = 1, the result is thrown away. This PR fixes that.


sgugger:
Indeed, good catch!

PR #1920, opened 2019-04-03
Update `fbeta` and `roc_auc_score` for binary

cstorm125:
Check if it's a case for one-label binary classification and if so modify `fbeta` and `roc_auc_score` to work in that case


sgugger:
There is a version of `fbeta` for single-label problems that is the class `FBeta`. 

As for the ROC curve, why pick the label index number 1 (instead of 0, 2, 3...)? It doesn't seem to make sense. We can also have a variant of this function suitable for single-label classification but I'm not sure it's the one you are suggesting.

Closing this. Maybe you can suggest a PR that checks if it's a single-labeled task and issue a warning instead?

PR #1927, opened 2019-04-05
FillMissing was super slow when a dataset had a lot of columns

EtienneT:
FillMissing was really slow on a dataset with a lot of columns with nan (1200 columns).  Simply removing the use of loc seems to speed it up tremendously (from minutes to 2-3s).


sgugger:
That's weird bu why not.

PR #1931, opened 2019-04-05
Bugfix/default drops text learner

bachsh:
When setting more than one hidden layer using `lin_ftrs` in `get_text_classifier` without setting `ps`, a wrong architecture is created. This PR fixes is and also issues an error when trying to init the internal `PoolingLinearClassifier` improperly.

### DIFF
--- a/fastai/text/learner.py
+++ b/fastai/text/learner.py
@@ -273,7 +273,7 @@ def get_text_classifier(arch:Callable, vocab_sz:int, n_class:int, bptt:int=70, m
     for k in config.keys(): 
         if k.endswith('_p'): config[k] *= drop_mult
     if lin_ftrs is None: lin_ftrs = [50]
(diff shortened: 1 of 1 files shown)


sgugger:
It looks like it's breaking existing behavior since the tests aren't passing. Can you check and debug? Thanks!

bachsh:
Sure, I was in a hurry and didn't wait for the tests to finish. The error
says something is wrong with bash, so I can't tell which test failed
exactly. I'll look for them when I get to my computer.

sgugger:
If you click on the line Run tests 1 error, you will see the details.

bachsh:
I made the change and tests in one repo and then copied to the other without testing. What a noob :(

Fixed it now.

sgugger:
Thanks!

PR #1936, opened 2019-04-06
New parameter to TabularModel

EtienneT:
Added a new parameter to TabularModel to be able to add a RELU at the end if you ever want to use TabularModel inside another module where the result of TabularModel is passed to something else.


sgugger:
I think if you pass `y_range = (0,1)` you will have a final ReLU (not visible in the summary but that will be applied during the forward pass). Please tell me if it doesn't work and reopen in this case.

sgugger:
Eh, how could have I confused a sigmoid with a ReLU... Guess I was really tired!

PR #1937, opened 2019-04-07
Datasets config

tydlwav:
Added 2 tests for `datasets.py`, one testing default config and another testing custom config. 

Also updated the documentation to include `data_archive_path`.


tydlwav:
Tests and documentation for #1904 #1905

sgugger:
Thanks! Two comments: I get weird HTML codes instead of seeing the markdown on ReviweNB for the datasets notebook. Is it because their representation is buggy or does it appear like this?

For the actual test, we try to avoid downloads during testing since it takes time. Can you use MNIST_TINY and ML_SAMPLE (instead of MNIST_SAMPLE) as your second dataset (it's the smallest one)?

tydlwav:
Realized the issue: Shouldn't run the code in the notebook of the ipynb. Probably should put that tip in [CONTRIBUTING.md](https://github.com/fastai/fastai/blob/master/CONTRIBUTING.md).

sgugger:
Thanks for the changes, it's looking much better. What do you mean by not run the code in the notebook? It's supposed to be run and be interactive... it's a notebook.

tydlwav:
Ah. It changes some of the contents in the notebook, so the diff looks horrible. Better not touch it than git reset. XD

PR #1951, opened 2019-04-09
pil2tensor - Only one tranpose required

mogwai:
The 2 transposes can be combined into one

```python
n1 = np.random.rand(2, 5, 3)

a = np.transpose(n1, (1, 0, 2))
a = np.transpose(a, (2, 1, 0))

print((n1.transpose(2,0,1) == a).all())
```


sgugger:
Indeed, but unless PyTorch has changed its code, it's much slower...
Closing this since we had done the two different transposes for speed reasons. If you can show this approach is now faster, please reopen with evidence!

PR #1952, opened 2019-04-09
Adding TextClassificationInterpretation docs

bfarzin:
Adding a doc entry for the TextClassificationInterpretation class to show usage and example.


sgugger:
This is a really long example, which I don't really mind for the text, but for the interesic attention, it makes for a very long tensor. Can you adjust for something a little bit shorter? Thanks a lot!

bfarzin:
I agree, example was too long.  I updated with the classic "Zombiegeddon" :)

sgugger:
Looking better, thanks!

PR #1954, opened 2019-04-09
Speed up Normalize

tchklovski:
very similar to #1927, but for Normalize
I was having this issue with both preprocessors


sgugger:
I don't see why not, thanks!

PR #1957, opened 2019-04-10
Optimise transpose flow

lvaleriu:
Use one single transpose operation in pil2tensor function


sgugger:
Like I said on #1951, this is for speed reasons. We won't change it without evidence that the new way is faster.

PR #1959, opened 2019-04-10
fix Image.resize() not returning self on same size

egonuel:
Image.resize() was not returning `self` if the requested size is the same as the current size.
Also adds test for this.

I did not add a test for for the general resize behavior, because there appears to be a bug with a target size of 1, which I can currently not look into and I don't want to add a test which does not cover this.


sgugger:
Thanks! Will look at this size 1 bug when I have time.

PR #1960, opened 2019-04-10
CLA Test - ignore this please

cluePrints:


cluePrints:
Was looking for this https://www.clahub.com/agreements/fastai/fastai, found elsewhere

PR #1964, opened 2019-04-11
Add show_top_losses() to TextClassificationInterpretation

thousfeet:
Enhance TextClassificationInterpretation by adding `show_top_losses()`, inspired by `plot_top_losses` in vision.learn.
This method creates a tabulation that shows texts in `top_losses` along with their prediction, actual, loss, and probability of actual class.
Also, I made `TextClassificationInterpretation` inherit `ClassificationInterpretation` so that all methods in `ClassificationInterpretation` can be called by `TextClassificationInterpretation` instance.

### DIFF
--- a/conda/meta.yaml
+++ b/conda/meta.yaml
@@ -53,6 +53,7 @@ requirements:
     - beautifulsoup4
     - fastprogress >=0.1.19
     - matplotlib
+    - tabulate
     - numpy >=1.15
     - pandas
     - bottleneck
--- a/conda/meta.yaml
+++ b/conda/meta.yaml
@@ -53,7 +53,6 @@ requirements:
     - beautifulsoup4
     - fastprogress >=0.1.19
     - matplotlib
-    - tabulate
     - numpy >=1.15
     - pandas
     - bottleneck
--- a/docs_src/text.learner.ipynb
+++ b/docs_src/text.learner.ipynb
@@ -332,6 +332,82 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
+    "Create a tabulation showing the first `k` texts in top_losses along with their prediction, actual,loss, and probability of actual class. `max_len` is the maximum number of tokens displayed. If `max_len=None`, it will display all tokens."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 14,
+   "metadata": {},
+   "outputs": [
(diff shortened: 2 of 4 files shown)


sgugger:
Thanks a lot. One thigns annoy me, and it's the added dependency. Can you put your table in a dataframe instead? If you look at `show_xy` in `TextList`, you'll see we show the data by putting it in a dataframe `df` then going
```
with pd.option_context('display.max_colwidth', -1):
    display(HTML(df.to_html(index=False)))
```
It would be great if you could use the same here as it would make it consistent with the rest of the library.

bfarzin:
Will you also please add to the docs?

thousfeet:
> 
> 
> Will you also please add to the docs?

Yes, with pleasure. I'll write them after this PR be merged.

thousfeet:
> 
> 
> Thanks a lot. One thigns annoy me, and it's the added dependency. Can you put your table in a dataframe instead? If you look at `show_xy` in `TextList`, you'll see we show the data by putting it in a dataframe `df` then going
> 
> ```
> with pd.option_context('display.max_colwidth', -1):
>     display(HTML(df.to_html(index=False)))
> ```
> 
> It would be great if you could use the same here as it would make it consistent with the rest of the library.

I've rewrote it and discarded dependency. Now the table looks like this:
图片

sgugger:
Thanks a lot. For future reference, don't forget to run
```
tools/run-after-git-clone
```
before making PRs with the docs notebooks, as it will automatically strip them of unnecessary metadata when you commit. That's why you had a failing test.

thousfeet:
> 
> 
> Thanks a lot. For future reference, don't forget to run
> 
> ```
> tools/run-after-git-clone
> ```
> 
> before making PRs with the docs notebooks, as it will automatically strip them of unnecessary metadata when you commit. That's why you had a failing test.

Sure, thanks!

mrdbourke:
Massive update @thousfeet! Thank you for this.

However, I'm running into an issue when I try to use it?

```
<ipython-input-144-1aad6567562c> in <module>
----> 1 txt_ci.show_top_losses(5)

AttributeError: 'TextClassificationInterpretation' object has no attribute 'show_top_losses'
```

I've uninstalled fastai via pip uninstall, and reinstalled it.

Potentially I have to build from source?

PR #1973, opened 2019-04-13
Reduce mem usage function. 

jamesdietle:
Including reduce memory make dataset smaller by changing types to the smallest needed.

The discussion is here:
https://forums.fast.ai/t/reduce-mem-usage-to-help-with-large-datasets/42707?u=mindtrinket

I included it in the tabular example to show it working.


sgugger:
I think all floats should be converted to float64: the weights of the models are going to be tensors of float32 at the end anyway, so I don't think there is any benefit keeping them when we want to reduce memory usage.

jamesdietle:
I might be misreading, do you mean all float64 should be converted to float32? If so, I would agree. 

The only reason that a float64 wouldn't become a float32 is when the value exceeds what can be handled by a float32.

sgugger:
Yes there might be nans, that would be processed by the tabular processors, but eventually, the tensors will be in FP32, so I don't think it's a good idea to keep the float64.

jamesdietle:
Made the change so float64 is never selected and to print a warning for the user if it would not fit in FP32.

sgugger:
Thanks a lot!

PR #1974, opened 2019-04-14
init parameter not being used in cnn_learner

hofesh:
fix init parameter not being used in cnn_learner


sgugger:
Good catch, thanks!

PR #1975, opened 2019-04-14
Feature/better learner predict fn

antoinebon:
* added support for models with multiple heads (list or tuple of tensors)
* reconstruct out with respect to item.data. For example needed for PointsLabelList where FlowField must be with respect to the size of the input image and not with respect to the size of the potentially resized version of the image in the batch


sgugger:
Sounds good, thanks a lot!

PR #1978, opened 2019-04-15
Add test for SortSampler implemented in fastai.text.data.py

numb3r33:
Adding a new test for SortSampler implemented in the fastai.text.data.py module.


sgugger:
Thanks!

PR #1980, opened 2019-04-15
Added to_cpu test on test_torch_core.py. On branch M0O-branch.

SOVIETIC-BOSS88:
Your branch is up-to-date with 'origin/M0O-branch'.

 Changes to be committed:
	modified:   fastai/test_registry.json
	modified:   tests/test_torch_core.py


sgugger:
Thanks!
It would be nice to have a on the GPU before applying the function (mark your test with the decorator `@pytest.mark.cuda` to note that it would use the GPU). And b should be a list of tensors to be fully testing the function.

SOVIETIC-BOSS88:
Thank you. I will add them now.

PR #1982, opened 2019-04-16
Added progress bar for parallel with 1 worker

H4dr1en:
I suggest the following change to show the nice progress bar when calling the `parallel` function with `max_workers = 1`

Example:
image


sgugger:
Good idea, but that would need to be an option with a boolean flag, as people might not want to use one.

H4dr1en:
Well, why would we let the choice for displaying the progress bar for the case `max_workers = 1` and enforce showing the progress bar otherwise ?

sgugger:
Oops! Had forgotten is was always there for multiproc!

PR #1986, opened 2019-04-17
Tests for vision.gan

ericricky:
This is my first attempt at a test script for vision.gan. Before I go any further, I was hoping to get some feedback. Can the tests that I've written so far be improved? My test for GANTrainer takes much longer than the other tests at around 6 seconds. Is there a guideline for how long the entire script should take? I've been taking inspiration from a few different test modules. Is there any one in particular that I should look at to use as a model? Also, I've noticed that there are some implicit requirements on the shapes of images and possibly other parameters. Image sizes seem to need to be powers of 2. Is that expected?

### DIFF
--- a/fastai/test_registry.json
+++ b/fastai/test_registry.json
@@ -1494,7 +1494,7 @@
     "fastai.vision.gan.GANTrainer": [
         {
             "file": "tests/test_vision_gan.py",
-            "line": 79,
+            "line": 80,
             "test": "test_gan_trainer"
         }
     ],
--- a/fastai/test_registry.json
+++ b/fastai/test_registry.json
@@ -1418,6 +1418,11 @@
             "file": "tests/test_vision_data.py",
             "line": 217,
             "test": "test_vision_datasets"
+        },
+        {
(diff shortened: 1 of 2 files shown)


sgugger:
This is looking good, thanks a lot! One thing is that we can't have a training without marking it with `@pytest.mark.slow` as we want the suite to run fast most of the times. Can you just add that decorator to the slow test?

ericricky:
@sgugger Yeah I'll do that. Also, should I include the update to the test registry in my pull?

sgugger:
Yes please, it's better this way.

ericricky:
Actually, how do I run a test marked with slow? I tried `pytest -m slow tests/test_vision_gan.py` but it still skipped.

sgugger:
pytest --runslow

ericricky:
@sgugger Awesome, thanks. I'll try to add a few more tests in the next few days, but feel free to merge if it looks good to you.

sgugger:
Let's do small increments! Thanks for this!

ericricky:
No problem, and sounds good!

PR #1988, opened 2019-04-18
adding time column to CSVLogger

sutt:
**Discussion + Context**
Request originated by @stas [on forum](https://forums.fast.ai/t/expand-recorder-to-deal-with-non-int-float-data/41534d)

>  ...so for example CSVLogger can’t save the time column of per-epoch training....

**Feature Description:**
The added functionality simply ports the same logic that is used in `Recorder` to record epoch time in `CSVLogger`: the time between `on_epoch_begin` and 'on_epoch_end' is returned as a formatted string of format 'MM:SS'.  

The only issue I see is that dataframe interprets the time value as strings / object-type. (There does appear to be pandas type to handle “timedeltas” but they seem more obscure and troublesome than helpful). 

Overall, I think this is helpful for people who want to review past training protocols and notice discrepancies in training time, especially when the recorder's output in a notebook get's erased with an overwrite or kernel interrupt.

**Testing Impact**
Both [panda] asserts in the tests had to be changed slightly to accommodate the extra column in this table. When comparing `CSVLogger` dataframe against `Recorder` dataframe, we can not include the time field as Recorder does not store (only prints) this information data.

### DIFF
--- a/fastai/callbacks/csv_logger.py
+++ b/fastai/callbacks/csv_logger.py
@@ -4,6 +4,8 @@ from ..torch_core import *
 from ..basic_data import DataBunch
 from ..callback import *
 from ..basic_train import Learner, LearnerCallback
+from time import time
+from fastprogress.fastprogress import format_time
 
 __all__ = ['CSVLogger']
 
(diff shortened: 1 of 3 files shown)


sgugger:
Looking good! I'll just put `float_or_x` in core since it belongs there more than in `csv_logger` as a follow-up. Thanks a lot

PR #1989, opened 2019-04-18
Add documentation for fastai.imports

ericricky:
I wrote a quick reference for the convenience imports that are pulled in whenever using a star import on a fastai module. This kind of documentation can help Python beginners get exposure to functions and libraries that experienced programmers find useful. It also should be helpful to have a compact list of these imports with links to documentation. I'm not sure where this would go, maybe under the "core" docs.

### DIFF
--- a/docs/imports.md
+++ b/docs/imports.md
@@ -1,0 +1,70 @@
+---
+title: Imports
+---
+
+## Introduction
+
+To support interactive computing, fastai provides easy access to commonly-used external modules. A star import such as
+```
+from fastai.basics *
+```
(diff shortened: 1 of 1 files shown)


sgugger:
This is looking good and is a good idea. @stas00 can you link to it under core (last item)? I think that would be a good place for it indeed.

stas00:
Done: https://docs.fast.ai/imports.html

The only issue is how to keep that page up-to-date - since it was created manually very quickly it'll become outdated...

ericricky:
Looks good, thanks!

stas00:
Committed.

PR #1992, opened 2019-04-19
make .to_fp16().to_fp16() work

tchklovski:
Currently, applying `.to_fp16()` to a `Learner` which is already in half-precision mode will make it have multiple `batch_to_half` tfm and `MixedPrecsion` callbacks. Always resetting with `.to_fp32()` should fix that and not cause performance issues

check with eg 
`[cb.__class__ for cb in learn.callbacks]`


sgugger:
Wacky ;)
Thanks for the fix!

PR #1994, opened 2019-04-20
Modify init to pull the collect_env libraries

kevinbird15:
This will allow somebody to the show_install function with `from fastai.utils import *` instead of needing to remember that it is in collect_env

Here is the forum thread discussing this change: https://forums.fast.ai/t/show-install-modification-suggestion/44214

### DIFF
--- a/docs/performance.md
+++ b/docs/performance.md
@@ -8,7 +8,7 @@ This document will show you how to speed things up and get more out of your GPU/
 
 To check your setup for recommended performance improvements, run:
 ```
-python -c "import fastai.utils.collect_env; fastai.utils.collect_env.check_perf()"
+python -c "import fastai.utils; fastai.utils.check_perf()"
 ```
 
 ## Mixed Precision Training
--- a/docs/support.md
+++ b/docs/support.md
@@ -52,7 +52,7 @@ When you make a post, make sure to include in your post:
    ```
    git clone https://github.com/fastai/fastai
    cd fastai
-   python -c 'import fastai.utils.collect_env; fastai.utils.collect_env.show_install(1)'
+   python -c 'import fastai.utils; fastai.utils.show_install(1)'
    ```
 
    The reporting script won't work if `pytorch` wasn't installed, so if that's the case, then send in the following details:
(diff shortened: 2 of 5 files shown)


sgugger:
Seems consistent with the rest of what we have. @stas00 any reason not to do this?

stas00:
I have no objection, but if you do that, please adjust the docs, otherwise only you will know that there is a different way.

sgugger:
Would you mind adding the proper documentation in this PR @kevinbird15 ?

kevinbird15:
Sure, I will get that added today sometime.

stas00:
Just to clarify, there is only a need to edit the existing docs:
```
docs/performance.md:python -c "import fastai.utils.collect_env; fastai.utils.collect_env.check_perf()"
docs/support.md:   python -c 'import fastai.utils.collect_env; fastai.utils.collect_env.show_install(1)'
docs/troubleshoot.md:python -c 'import fastai.utils.collect_env; fastai.utils.collect_env.show_install(1)'
docs/troubleshoot.md:python -c 'import fastai.utils.collect_env; fastai.utils.collect_env.show_install(1)'
docs_src/utils.collect_env.ipynb:    "from fastai.utils.collect_env import *"
```
Thank you.

kevinbird15:
Ok, so would you change it to just be fastai.utils and remove the collect_env part then or just note that there is another way to do it?  So for each of these: 
```
docs/performance.md:python -c "import fastai.utils; fastai.utils.check_perf()"
docs/support.md:   python -c 'import fastai.utils; fastai.utils.show_install(1)'
docs/troubleshoot.md:python -c 'import fastai.utils; fastai.utils.show_install(1)'
docs/troubleshoot.md:python -c 'import fastai.utils; fastai.utils.show_install(1)'
docs_src/utils.collect_env.ipynb:    "from fastai.utils import *"
```

kevinbird15:
Ok, probably terrible practice, but I added commits to change these docs.  I would have rather put all of the commits into one file, but I did it with the online edit file functionality.  It seems like it worked ok, but I would figure out how to do it on my local version next time and push the changes up instead. Quick and dirty won me over this time though

sgugger:
We don't mind quick and dirty, I know how painful it can be to checkout your PR branch and add commits when not used to git :)

stas00:
> I know how painful it can be to checkout your PR branch and add commits when not used to git :)

FYI, it's trivial to do if you use [fastai-make-pr-branch](https://docs.fast.ai/dev/git.html#helper-program). It does all the work for you.

PR #1999, opened 2019-04-23
Add LMTextList to __all__

kevinbird15:
Add LMTextList so it pulls in with from fastai.text import * 

This will allow LMLabelList to be called similar to how LMLabelList can be at the moment.  I ran into an issue where I was trying to debug label_for_lm and that caused an issue for me. Pretty sure this won't break anything.


sgugger:
This is an oversight, this function should be public. Thanks for fixing!

kevinbird15:
I was hoping that was the case.  I was having a tough time getting my LM flowing through the data_block API.  Once I swapped to LMTextList instead of just TextList, it all seems to be working much better.

PR #2002, opened 2019-04-24
Optionally disable transforms on y

xeTaiz:
Hi there!
As discussed in [the forums](https://forums.fast.ai/t/tfm-y-only-on-some-transforms/43813/9), here is functionality to disable specific transforms for targets (assuming `tfm_y=True` as well)

With those changes, you can "skip" transforms as follows:
```
tfm_train, tfm_val = get_transforms()
tfm_train += [
    cutout(n_holes=(1,4), length=(1, hp['im_sz']//8), skip_y=True),
    crappify_randomly(max_noise_std=.025, skip_y=True)
]
# ....
data = Data.blocks().fun().transform((tfm_train, tfm_val), tfm_y=True)
```
In this case all transformations from `get_transforms()` are applied to both `x` and `y`, however the `cutout` and `crappify_randomly` are only applied to `x`, but not to `y`.
If nothing is specified the behavior is unchanged.

Maybe there's a better parameter name than `skip_y`? Suggestions are welcome on this one. Will add documentation when the name is decided. @sgugger

### DIFF
--- a/fastai/data_block.py
+++ b/fastai/data_block.py
@@ -719,8 +719,8 @@ class LabelList(Dataset):
         _check_kwargs(self.y, tfms, **kwargs)
         self.tfm_y=True
         if tfms is None:
-            self.tfms = list(filter(lambda t: t.use_on_y, self.tfms))
-            self.tfms_y,self.tfmargs_y = self.tfms,{**self.tfmargs, **kwargs}
+            self.tfms_y = list(filter(lambda t: t.use_on_y, self.tfms))
+            self.tfmargs_y = {**self.tfmargs, **kwargs}
         else:
             tfms = list(filter(lambda t: t.use_on_y, tfms))
             self.tfms_y,self.tfmargs_y = tfms,kwargs
--- a/fastai/data_block.py
+++ b/fastai/data_block.py
@@ -25,7 +25,7 @@ def get_files(path:PathOrStr, extensions:Collection[str]=None, recurse:bool=Fals
     "Return list of files in `path` that have a suffix in `extensions`; optionally `recurse`."
     if recurse:
         res = []
-        for i,(p,d,f) in enumerate(os.walk(path)): 
+        for i,(p,d,f) in enumerate(os.walk(path)):
             # skip hidden dirs
             if include is not None and i==0:  d[:] = [o for o in d if o in include]
             else:                             d[:] = [o for o in d if not o.startswith('.')]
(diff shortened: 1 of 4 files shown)


fastai/vision/image.py
```diff
@@ -459,10 +459,10 @@ def __init__(self, func:Callable, order:Optional[int]=None):
         setattr(Image, func.__name__,
                 lambda x, *args, **kwargs: self.calc(x, *args, **kwargs))
 
-    def __call__(self, *args:Any, p:float=1., is_random:bool=True, **kwargs:Any)->Image:
+    def __call__(self, *args:Any, p:float=1., is_random:bool=True, use_on_y=True, **kwargs:Any)->Image:
```
sgugger:
```suggestion
    def __call__(self, *args:Any, p:float=1., is_random:bool=True, use_on_y:bool=True, **kwargs:Any)->Image:
```

fastai/data_block.py
```diff
@@ -709,16 +709,21 @@ def transform(self, tfms:TfmList, tfm_y:bool=None, **kwargs):
...
         return self
 
     def transform_y(self, tfms:TfmList=None, **kwargs):
         "Set `tfms` to be applied to the targets only."
         _check_kwargs(self.y, tfms, **kwargs)
         self.tfm_y=True
-        if tfms is None: self.tfms_y,self.tfmargs_y = self.tfms,{**self.tfmargs, **kwargs}
-        else:            self.tfms_y,self.tfmargs_y = tfms,kwargs
+        if tfms is None:
+            self.tfms = list(filter(lambda t: t.use_on_y, self.tfms))
```
sgugger: Why change `self.tfms`? It's only `self.tfms_y` that need to be the filtered list, no?

xeTaiz: yep, you're right

xeTaiz:
I wonder where I should add the documentation of this? Where ever `tfm_y` is described? 

I have put a note on `Transform` and also in the data_blocks docu wher `tfm_y` is described with the segmentation mask example.
Can you think of more sensible places? Otherwise I think this is good to go

sgugger:
That's seems good to me, and I can't think of any other places to add documentation. Thanks!

kuds:
I believe this change has caused an issue with MixedItemList and applying transformations.

This code below works in version 1.0.51, but 1.0.52

`mixed = (MixedItemList([imgList, tabList], path='', inner_df=tabList.inner_df)
            .split_by_idx(valid_idx=testSplit.index)
            .label_from_df(cols=5, label_cls=FloatList)          
            .transform(tfms=[[get_transforms()[0], []], [get_transforms()[1], []]], size=100))

AttributeError: 'list' object has no attribute 'use_on_y'`

Is there work around that I can implement? Maybe adding the 'use_on_y' attribute to the tabular transformation object? Or should I open up an issue?

sgugger:
Just pushed a fix to make it optional (it grabs the use_on_y attribute if it's there but doesn't complain if it's not).

PR #2003, opened 2019-04-24
Removing duplicate code in transformer.py

KushajveerSingh:
The changes are for `fastai/text/models/transformer.py`.

The changes are for `MultiHeadAttention` and `MultiHeadRelativeAttention` classes, more specifically in the `_apply_attention` and `_attention_einsem` functions of these classes.

The changes include changing `x.size(1)` to `x_len` for all the above functions in the following line
`wq,wk,wv = map(lambda x:x.view(bs, x.size(1), self.n_heads, self.d_head), (wq,wk,wv))`

And the second change, is in the if statement
```
if self.scale: attn_score = attn_score.div_(self.d_head ** 0.5)

if self.scale: attn_score.div_(self.d_head ** 0.5)
```

### DIFF
--- a/fastai/text/models/transformer.py
+++ b/fastai/text/models/transformer.py
@@ -105,7 +105,7 @@ class MultiHeadRelativeAttention(MultiHeadAttention):
         context = x if mem is None else torch.cat([mem, x], dim=1)
         wq,wk,wv = torch.chunk(self.attention(context), 3, dim=-1)
         wq = wq[:,-x_len:]
-        wq,wk,wv = map(lambda x:x.view(bs, x.size(1), self.n_heads, self.d_head), (wq,wk,wv))
+        wq,wk,wv = map(lambda x:x.view(bs, x_len, self.n_heads, self.d_head), (wq,wk,wv))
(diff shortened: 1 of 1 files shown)


sgugger:
Thanks!

jolackner:
As of fastai 1.0.52 , I get a runtime error after loading LM data into a TransformerXL and then running lr_find(). The error seems connected to these changes described above (until and including 1.0.51 it worked fine):  

image
image
image

I would be grateful for any suggestions on how to fix this problem. 
Best regards
Johannes Lackner

sgugger:
Oh, that's what happens when you have x designing two different things (input and variable in the lambda function). I see the problem, will push a fix shortly, thanks for raising the issue!

sgugger:
Fixed in [this commit](https://github.com/fastai/fastai/commit/097af1dbdd9c571a7dab89241f7c55ba8c84deac). It will be in the next release of fastai, in the meantime you can either stay with 1.0.51 or do an editable install of the library.

jolackner:
Wow, that was lightning fast. Thank you very much!

PR #2004, opened 2019-04-24
Low Memory Tokenizer

bfarzin:
This is my attempt at a lower memory tokenizer.  I ran a few tests with a large corpus (multiple GB) and confirmed that this version peaks at 17 GB RAM vs. 46 GB on the old tokenizer.

More/bigger changes than I have made before to the code base, so looking for feedback on any style changes or ways to layout the changes in the functions.  It passes tests on my machine but I realize there is not 100% coverage with the test suite.

Thanks for considering.


sgugger:
Closing this, not because I don't like the idea, but because I don't want tokenization and numericalization to be merged together. You should keep the two separate processors. Also, it might be easier to start by defining two new processors, then we can test and gradually merge them with the existing ones.

bfarzin:
Got it. Let me try that path and see what I can come up with.

PR #2006, opened 2019-04-27
train/valid split not-reproducible b/w machines - sort file objs on load

sutt:
[Forum Post](https://forums.fast.ai/t/train-valid-split-not-reproducible-b-w-machines-sort-file-objs-on-load) (contains additional details if you want them)
tldr - random operation don't reproduce across machines b/c the same data-dir on different machines loads in a different original order, before random operations even take place.

**Problem:** I'm failing to reproduce the train/valid split generated by split_rand_pct between my local env and server env. It's not just one-off but completely different: with an 80/20 split scheme, my server-generated data only matches 79.9% of my locally-generated data. (Yes, I've checked np.random seeding thoroughly)

The basic reason appears to be os.scandir() returns the files in a different order on different machines. According to its doc, scandir is an "iterator of DirEntry objects for given path", and so we can not expect return order to be preserved across all implementations.

**Impact:** when a datablock is built on machine1, then trained and the model exported to machine2, machine2 must be able to exactly reproduce the train/valid split created on machine1. Otherwise, when evaluating model error against the validation split on machine2, it will contain items originally used in training the model on machine1, and therefore likely understate the model's error (since train_err < valid_err on average).

Differences in loading the unzip-directory b/w `7z` and `unzip` likely caused my issue. But thinking about it, I also was able to create the same error in a simple [toy example notebook](https://github.com/sutt/fastai-tmp/blob/master/inode-sort-bug.ipynb), where someone inserts a new data file into an existing data-dir (if data-dir has data_1 and data_3 and then data_2 is subsequently inserted, then tarred/un-tarred with only linux tools then file order can become unreproducible). This issue will also impact other sampling operations like filter_rand, etc. 

**Solution:** Since adding in data in an ad-hoc manner is a common workflow in research, and the ecosystem of zip/tar tools can't be expected to always match in behavior, I recommend a fix to completely eliminate the possibility of the bug at its root. Simply sort the data loaded into an ItemList from a directory  by alphabetical ordering in get_files(), which will ensure the same order across platforms/os's python versions etc, and ensure the same operations on all appropriately seeded subsequent random operations. This will allow us to reproduce all random operations across machines.

**Backwards compatibility:** this could re-shuffle the data-blocks for some users of previous version of fastai if their data-directory were janky like mine. I think this very limited issue for mooc-students as course code, dataset and tools are predefined. Testing impact does not show any problems. (Also, one other bug may be fixed with this, that running course-v3 cells that get a particular image by idx number do not match between teacher and student). Nonetheless, the proposed PR, which solves the problem has an argument, `presort` in `from_folder`, which can be manually turned to False to allow previous behavior and thus reproduce the ordering from previous versions. This should be added to the docs.

**Sorting Strategy:** we'll sort on the full relative path because filenames may be duplicated in recursive/multi-directory-loads. We'll sort on a string representation of paths to avoid any peculiarties of sort applied to a path object. And because the same WindowsPath and PosixPath are converted to different strings (my\\path vs my/path) we'll sort with all forwardslashes and backslashes replaced to '.'. This all occurs in data_block::_path_to_same_str().


sgugger:
I would default this new behavior to `False` for
1. backward compatibility
2. this adds computation in a function that has been designed to be fast. I don't know what is the impact of the new sort but for an imagenet loading, it could be substantial.

sgugger:
Hi @sutt . Do you have time to follow up on this? Happy to apply the fix myself if you don't.

sutt:
Yes, it will be done by tomorrow afternoon.
(Got a little carried away on the coursework this week)

sgugger:
Thanks a lot!

PR #2008, opened 2019-04-27
reconstruct x with respect to batch item x[0], return reconstructed b…

antoinebon:
I would like to revert back to using `x[0]` to reconstruct `x` (instead of item.data)
This is because the prediction is based on `x` and not on the `item.data` and these might be different
However it is interesting to return the reconstructed `x` to be able to perform visualization such as:
```
x, y, pred, raw_pred = learn.predict(img)
x.show(y=y)
```
if `img` and `x` are different (for example different sizes), then `img.show(y=y)` will not make sense but `x.show(y=y)` will.

However it still requires some logic to convert a prediction `y` for `x` to `img`, when resizing transformations have been applied

Also that this PR changes the api: `predict` here returns a 4-tuple

The documentation would also need to be updated with this api change.
I could also only return `y, pred, raw_pred` so that it does not change the api

### DIFF
--- a/fastai/basic_train.py
+++ b/fastai/basic_train.py
@@ -359,7 +359,7 @@ class Learner():
                           cb_handler=CallbackHandler(self.callbacks))
         return loss
 
-    def predict(self, item:ItemBase, batch_first:bool=True, **kwargs):
+    def predict(self, item:ItemBase, return_x:bool:False, batch_first:bool=True, **kwargs):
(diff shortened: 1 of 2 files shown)


sgugger:
Please add a boolean flag that defaults to False to provoke this behavior: `learn.predict` is used in all the deployed apps in production as taught in the MOOC so we can't break the existing behavior.

sgugger:
Thanks!

PR #2010, opened 2019-04-28
Added usage examples for functions in data_block docs

KushajveerSingh:
The discussion for this was done on the fastai forums. ([link](https://forums.fast.ai/t/adding-detailed-explanation-of-the-use-cases-of-the-various-data-block-api-functions/44506)). 

The changes included are
1. form_folder:- Added one line explaining path and added an image showing an example of usage
2. from_df:- Added text. It explains how from_df can be used in any general situation and give an example of explaining how to set the paths.
3. split_by_files:- Added a cifar-10 dummy example, on how to use the function
4. split_by_fnames_files:- Added, Internally makes a call to split_by_files . fname contains your image file names like 0001.png.
5. split_by_idx:- Added an example showing usage.
6. split_by_list:- Referring to this issue. I think it was confusing to use this function, as it is used mostly as an internal function. So added a line telling this.
7. split_from_df:- Added example showing usage


sgugger:
Thanks. Please note you don't have to update the HTML yourself, we will do it on our side.

KushajveerSingh:
I did not update the HTML. When I committed the code, I also saw that some HTML changes were pushed. And I don't know how they occurred in the first place, as I did not make any edits in the `docs` folder.

PR #2011, opened 2019-04-28
fix core.recurse_eq to check for length of compared arrays

MLSlayer:
Currently, core.py's recurse_eq method might not enumerate all elements of the array, but still consider two arrays equal

for example,

```
>>> def recurse_eq(arr1, arr2):
...     if is_listy(arr1): return np.all([recurse_eq(x,y) for x,y in zip(arr1,arr2)])
...     else:              return np.all(np.atleast_1d(arr1 == arr2))
... 
>>> 
>>> arr1 = [1,2,3]
>>> arr2 = [1,2]
>>> recurse_eq(arr1, arr2)
True
```

In order to fix this, I've added a check for length for each step of the recursion.

```
def recurse_eq(arr1, arr2):
    if len(arr1) != len(arr2): return False
    if is_listy(arr1): return np.all([recurse_eq(x,y) for x,y in zip(arr1,arr2)])
    else:              return np.all(np.atleast_1d(arr1 == arr2))

```

### DIFF
--- a/fastai/core.py
+++ b/fastai/core.py
@@ -164,7 +164,7 @@ class ItemBase():
     def __eq__(self, other): return recurse_eq(self.data, other.data)
 
 def recurse_eq(arr1, arr2):
-    if is_listy(arr1): return np.all([recurse_eq(x,y) for x,y in zip(arr1,arr2)])
+    if is_listy(arr1): return len(arr1) == len(arr2) and np.all([recurse_eq(x,y) for x,y in zip(arr1,arr2)])
     else:              return np.all(np.atleast_1d(arr1 == arr2))
         
 def download_url(url:str, dest:str, overwrite:bool=False, pbar:ProgressBar=None,


sgugger:
This is making the tests fails, because it also happens that arr1 and arr2 are ints or floats. You should amend your PR to add the test inside the `is_listy` condition.

MLSlayer:
I've made the changes and force pushed with the` git commit --amend --no-edit` to the forked repo(wasn't sure if this is the correct practice). 
Is there any way to run the tests above even though I took this approach, or should I make a whole new PR and close this one? 

The new code is now

```
def recurse_eq(arr1, arr2):
    if is_listy(arr1): return len(arr1) == len(arr2) and np.all([recurse_eq(x,y) for x,y in zip(arr1,arr2)])
    else:              return np.all(np.atleast_1d(arr1 == arr2))

```

MLSlayer:
Oh it seems github was smart enough to kick off the tests and good to see they all passed!

sgugger:
Perfect, thanks a lot!

MLSlayer:
Great, glad it was merged!

Infact, due to not being familiar with the fastai codebase, now I'm not sure if the fix would suffix due to the case when two lists might have different sublists, for example

```
>>> arr1 = [[1]]
>>> arr2 = [1]
>>> def is_listy(x)->bool: return isinstance(x, (tuple,list))
... 
>>> def recurse_eq(arr1, arr2):
...     if is_listy(arr1): return len(arr1) == len(arr2) and np.all([recurse_eq(x,y) for x,y in zip(arr1,arr2)])
...     else:              return np.all(np.atleast_1d(arr1 == arr2))
... 
>>> recurse_eq(arr1,arr2)
  File "<stdin>", line 2, in recurse_eq
  File "<stdin>", line 2, in <listcomp>
  File "<stdin>", line 2, in recurse_eq
TypeError: object of type 'int' has no len()
```

First, is this a legitimate concern, due to how fastai is using ItemBase(i.e preconditions or data passed in to the self.data and other.data of the `ItemBase.__eq__`)? 

Second, should the above actually be equal?(i.e. [[1]] is equal to [1])

Third, if in the future I want to contribute code with modifying tests you guys have hooked to github, how would I do that and which folder can I find that in? 

The fix for the above depending on [[1]] not being equal to [1] would be 

```
def recurse_eq(arr1, arr2):
    if is_listy(arr1): return is_listy(arr2) and len(arr1) == len(arr2) and np.all([recurse_eq(x,y) for x,y in zip(arr1,arr2)])
    else:              return np.all(np.atleast_1d(arr1 == arr2))
```

sgugger:
This is a legitimate concern as we never know how this core function may be used in the future and anyone who sees it is going to assume it works, so you should go ahead and suggest your second fix.

If you want to add more tests, it's all in the tests folder of the repo.

MLSlayer:
Will do, I plan to submit a pull request with branch named `rebugfix-recurse_eq `sometime soon then

PR #2014, opened 2019-04-29
update docs for get_files and _get_files

EmbraceLife:
I have experimented these two functions and provided basic docs to explain their supposed behaviors.


sgugger:
This is too long for docstrings that we want short and to the point. You can add more prose in the doc notebook instead!

PR #2015, opened 2019-04-29
fix core.recurse_eq to check that arr2 is listy

MLSlayer:
-- added test cases for both commit ac1c92d and this commit


sgugger:
Thanks!

Issue #1998, opened 2019-04-23
Tokenization is time and space inefficient


kliron:
Going through the code in [transform.py](https://github.com/fastai/fastai/blob/master/fastai/text/transform.py) I cannot help but notice several opportunities for optimization for parallel execution. In its current form it would take >4 days to tokenize a 12Gb corpus on a 16-core/32 thread CPU (if it wouldn't run out of memory first as 36G RAM weren't enough). Writing a custom implementation reduced both the time to a little more than 4 hours and memory use by 2-3x. The code is mission-specific and write-once dirt so I'm feeling reluctant to share, but I'd gladly share the gist of it below.

1. In its current implementation the tokenization process is parallelized using the very inefficient `concurrent.futures.ProcessPoolExecutor`'s `map` function which creates Future objects where there is no good reason to. These are good for fine-grained control like progress reporting, cancelling etc, but are fairly heavy. In this case we are actually only interested in the returned tokens. `multiprocessing.Pool`'s `map` should perform considerably better. See [this SO post](https://stackoverflow.com/questions/18671528/processpoolexecutor-from-concurrent-futures-way-slower-than-multiprocessing-pool
) for more details.

2. A number of new processes are created to tokenize each bach of text. This means a new batch of processes need to be forked, initialize a fresh instance of Spacy Tokenizer and receive a fairly large chunk of text via IPC  every few seconds. This seems very inefficient. For small enough batches more time will be spent in fork-IPC-joining than on the actual work being done. Alternatively, there should be a number of long-lived tokenizer worker processes initialized at the beginning with a workload to process and each should process a stream of text with the more efficient `Tokenizer.pipe` function from [Spacy](https://spacy.io/api/tokenizer#pipe). 

Also, for large batches it is by leaps and bounds more efficient to have each process read its own batch from disk than to have some producer process provide it by IPC. Python's performance for reading large objects through IPC is atrocious (I don't know if this is Python-specific). See [this SO post](https://stackoverflow.com/questions/47085458/why-is-multiprocessing-queue-get-so-slow) for more context.

3. The current implementation requires enormous amounts of RAM for relatively small corpora (something like ~1Gb requires >24G RAM). Serializing the tokens and word counts from each tokenizer worker to disk, merging them after tokenization and (if needed) truncate the vocabulary and replace deleted instances with UNK in the tokenized text files is vastly more memory-efficient and can scale to much larger texts. 

[...]

PR #2016, opened 2019-04-29
File tokenizer

bfarzin:
OK, Another attempt at this with feedback form @kliron to help with managing the process across multiple cores.  Writes to temp files and then pulls form those to `Numericalize`

Some issues/questions:
1. Where is the best place to write these temp files?
2. Should we use a memory map to prevent copying the file across processes (I still see a big memory footprint when I run this code.)  My input file is 1.75 GB and I run on 16 cores and my peak memory is about 16 * 1.75 = 28 GB.  That seems high. Any pointers on how to lower that would be very helpful.
3. How/what should be done with the current `Tokenizer`?  Can we replace or should we keep two copies and if two copies any suggestions about naming?

Also, as usual, I appreicate any feedback on coding style or format that makes things easier to understand.  All is a bit messy right now, but if the overall approach looks good, I can clean up and streamline moving some functions into `core`, etc. 

Thank you for reviewing and considering!


sgugger:
Thank you, Bobak, but as discussed in the [issue](https://github.com/fastai/fastai/issues/1998) I was already planning on resolving it but it can't happen until the course material is merged into master (new fastai code base), so now is a really bad timing to sort the nuances out.

I think you work would be very helpful to me to complete this task, I hope you're OK if we put this on hold and then revisit it once the new code base is created.

Until then @kliron or any person interested could use your branch if need to (it's breaking the test suite though, so at their own risks).

bfarzin:
Totally ok with putting on hold.  Glad that it can be of some help to you when you come back to it (if it is.) 
Great exercise for me to work through the details.  I understand that part of the code much better than I did before.  I believe I can create my own files that reference this and use till you get to it.  

Thanks for having a look.  Let me know if I can help in any way with it when you get around to the details.  Happy to assist!

sgugger:
I will definitely look at your code as I create the new tokenizer and will appropriately cite on the bits I take. I'm leaving this PR open until v1.1.0 is out (at least in a branch state) so anyone can fork your code in the meantime.

kliron:
Regarding your memory concerns, take into consideration that tokenization is a *very* memory intensive operation (as just about anything in NLP is). Spacy is clever in trying to optimize as much as it can by caching all tokens, hashing them and not keeping multiple copies of strings in memory but, again, the limitations of Python's multi-process concurrency model takes its toll as the cache cannot be shared across all tokenizers. 
Even worse, memory that is garbage collected by Python [is not guaranteed to be released back](http://effbot.org/pyfaq/why-doesnt-python-release-the-memory-when-i-delete-a-large-object.htm) to the OS - the only way to have this guarantee is to have the memory intensive operation run in a subprocess and have that terminate immediately after work, which is kinda what we try _not_ to do here.

sgugger:
Closing for the same reasons as the issue. Thanks a lot for the PR as it gave us an example on which we have built. The fork can still be accessible for anyone who wants to use this even if the PR is closed (and you can link on the forum to help people find it @bfarzin).

bfarzin:
Glad that this helped you in some small way!!

PR #2030, opened 2019-05-03
Passing model.data instead the full de model

trackq:
LabelList.load_state() expects model.data not the full model


sgugger:
Yes that is why this is done [here](https://github.com/fastai/fastai/blob/fa1b0b451eb284bd782b218a0b1f29a9ecc8c46c/fastai/basic_train.py#L601). If you're using `LabelList.load_state` directly, you should feed it something you go with `LabelList.export`.

PR #2033, opened 2019-05-03
Favor the metric function's direct __name__

### LINKED ISSUE #2032: Custom names for partially applied metrics
I'm using `top_k_accuracy` with different `k`s in my projects using partial application like this in example:

```python
top10_acc = partial(top_k_accuracy, k=10)
top20_acc = partial(top_k_accuracy, k=20)
```

Unfortunately, these are printed/logged with the same base metric name during training:

```python
>>> learn.fit(1)
epoch     train_loss  valid_loss  top_k_accuracy top_k_accuracy  time
```

I would like to able to give them a custom name, so they could be displayed like this:
```python
>>> learn.fit(1)
epoch     train_loss  valid_loss  top10_accuracy top20_accuracy  time
```

I've already digged into the code and found out that callbacks like the `CSVLogger` use the `Recorder`'s name property as source for the metric names. The `Recorder` in turn reads the `name` property of the (wrapped) metric (`AverageMetric` in this case). So the relevant code line should be this one: https://github.com/fastai/fastai/blob/master/fastai/callback.py#L332 => the current implementation use the `__name__` property of the original function `func` in the first place if it is a partially applied function.

I suggest replacing the line with the following logic:
```python
# If func has a __name__ use this one else it should be a partial
name = func.__name__ if hasattr(func, '__name__') else func.func.__name__
```
Now I could assign a `__name__` property to my partials and they would be used as custom metric names:
```python
top10_acc = partial(top_k_accuracy, k=10)
top20_acc = partial(top_k_accuracy, k=20)

top10_acc.__name__ = 'top10_accuracy'
top20_acc.__name__ = 'top20_accuracy'
```

Do you think this would be an appropriate solution to my problem? If this is the case, I would create a pull request including tests for it.

juriwiens:
Enables custom names for partially applied metrics => closes #2032

I was not really sure where to locate the tests, as `AverageMetric` is a `Callback` but test_metrics seemed to me intuitively to be the better place.

### DIFF
--- a/fastai/callback.py
+++ b/fastai/callback.py
@@ -328,8 +328,8 @@ class CallbackHandler():
 class AverageMetric(Callback):
     "Wrap a `func` in a callback for metrics computation."
     def __init__(self, func):
-        # If it's a partial, use func.func
-        name = getattr(func,'func', func).__name__
+        # If func has a __name__ use this one else it should be a partial
+        name = func.__name__ if hasattr(func, '__name__') else func.func.__name__
         self.func, self.name = func, name
         self.world = num_distrib()
 
--- a/fastai/test_registry.json
+++ b/fastai/test_registry.json
@@ -230,6 +230,13 @@
             "test": "test_accuracy"
         }
     ],
+    "fastai.callback.AverageMetric": [
+        {
+            "file": "tests/test_metrics.py",
+            "line": 190,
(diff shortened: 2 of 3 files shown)


sgugger:
test_metrics is the good place for that. Thanks a lot for your PR!

PR #2034, opened 2019-05-03
Move attempt at writing to write check

mitar:
So it you do not have write permissions then currently creating directory for models fails. This makes it impossible to use this library even if you will never save anything. It is better to do all potentially failing operations in one place. So I moved it to where a check is done.


sgugger:
Makes sense, thanks!

sgugger:
I reverted this because you made your PR on the release branch, not the master branch. Can you do it again on the master branch?

mitar:
Oh, how that happened. Sorry about that.

sgugger:
Np, we just need to put it in the right place ;)

mitar:
Done: #2034.

PR #2039, opened 2019-05-04
Move attempt at writing to write check

mitar:
Same as #2034, but to the correct branch.

So it you do not have write permissions then currently creating directory for models fails. This makes it impossible to use this library even if you will never save anything. It is better to do all potentially failing operations in one place. So I moved it to where a check is done.


sgugger:
Ah, it makes the test fails because it's creating a models directory when we didn't have one and were testing for failure. I'll add something to make it read-only.

sgugger:
Fixed the test, thanks for your PR!

mitar:
Thanks for merging!

PR #2040, opened 2019-05-04
provide examples on how `open_image` working with `convert_mode`

EmbraceLife:

PR #2041, opened 2019-05-05
example of using `cmap` for `image.show`

EmbraceLife:


sgugger:
I like this one, but this including the whole of #2040 which is too long as I commented there. Can you fix it?

EmbraceLife:
> I like this one, but this including the whole of #2040 which is too long as I commented there. Can you fix it?

Thanks! it is done now.

sgugger:
Thanks!

PR #2044, opened 2019-05-06
fix: fix typo

Ir1d:
fixing a tiny typo here.
Btw, I think using notebooks to maintain docs will make it hard for others to contribute (at least a bit harder for me, I suggest maintaining some plain text version :smile: )


sgugger:
Thanks!
Having docs in notebooks makes it easier for anyone to interact with the docs, which trumps the fact it's a little bit harder to maintain in our opinion ;)

PR #2051, opened 2019-05-08
add examples and docs on `ItemList.label_from_folder`

EmbraceLife:


sgugger:
I don't want to show the use of labelling `ItemList` directly in the docs as it's not standard good practice (you had to hack your way around and use the filter `from_item_lists=True`).
Can you do an example when you label a split ItemLists?

PR #2052, opened 2019-05-09
add docs and an example for `ItemList.label_from_func`

EmbraceLife:


sgugger:
Same as #2052 . This is interesting but you should label a split `ItemLists` for your example.

PR #2054, opened 2019-05-09
MultiLabelFbeta Metric

KeremTurgutlu:
This is an implementation of fbeta score metric using `LearnerCallback` for multilabel targets. There is already [FBeta(CMScores)](https://github.com/fastai/fastai/blob/master/fastai/metrics.py#L202) but it uses confusion matrix, hence `argmax` which doesn't allow multilabel output.

As a solution there is also [fbeta()](https://github.com/fastai/fastai/blob/master/fastai/metrics.py#L12) which is intended for mutlilabel use case but lacks the average metric properties, only calculated per batch and doesn't give the average options: [micro, macro, weighted, none]. 

Implementation is tested and compared with [sckit-learn fbeta](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.f1_score.html). Here is a [notebook](https://github.com/KeremTurgutlu/experimental/blob/master/Multiclass%20Fbeta%20.ipynb) using planet dataset.

### DIFF
--- a/fastai/metrics.py
+++ b/fastai/metrics.py
@@ -2,6 +2,7 @@
 from .torch_core import *
 from .callback import *
 from .layers import *
+from .basic_train import LearnerCallback
 
 __all__ = ['error_rate', 'accuracy', 'accuracy_thresh', 'dice', 'exp_rmspe', 'fbeta','FBeta', 'mse', 'mean_squared_error',
             'mae', 'mean_absolute_error', 'rmse', 'root_mean_squared_error', 'msle', 'mean_squared_logarithmic_error',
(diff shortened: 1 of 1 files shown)


sgugger:
Don't forget to run the test suite locally before pushing your PR, you were missing an import ;)
Thanks a lot!

KeremTurgutlu:
Sorry for that thanks :)

driesbuyck:
When using this metric and save and load the learn object, I get a pickle error.

learn.fit_one_cycle(1,0.1,moms=(0.8,0.7))
learn.save('first')
learn.load('first)

--> 292     pickler.dump(obj)
    293 
    294     serialized_storage_keys = sorted(serialized_storages.keys())

TypeError: can't pickle weakref objects

KeremTurgutlu:
Can you provide a minimal code example so that we can reproduce the issue. I've tried saving and loading locally with CPU, it worked. So there might be another issue. Thanks

PR #2055, opened 2019-05-09
add docs and an example for `ItemList._label_from_list`

EmbraceLife:


sgugger:
Methods that being with an _ are private and aren't supposed to be used directly, so we shouldn't document them.

PR #2070, opened 2019-05-11
add docs to `index_row`

EmbraceLife:

### DIFF
--- a/docs_src/core.ipynb
+++ b/docs_src/core.ipynb
@@ -1164,6 +1164,13 @@
    ]
   },
   {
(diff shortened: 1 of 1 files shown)


sgugger:
You should just say `a` is something you can index into like a dataframe, an array or a list.

sgugger:
Sorry clicked the wrong button and merged instead of closing.

EmbraceLife:
> You should just say `a` is something you can index into like a dataframe, an array or a list.

Thanks, I got it

PR #2076, opened 2019-05-12
docs and example added to `LabelLists.process`

EmbraceLife:


sgugger:
I wouldn't document process more than to say it's internally called after the labeling to apply the processors. 
The end user will never have to call it himself.

EmbraceLife:
> I wouldn't document process more than to say it's internally called after the labeling to apply the processors. What you did in the other PR was enough, no need for more!
> The end user will never have to call it himself.

Thanks, it is helpful.

PR #2083, opened 2019-05-13
add docs to ItemList.process and CategoryProcessor.process

EmbraceLife:
I noticed these two methods are intentionally hidden, I just wonder whether I can still add docs to them in case someone may find them useful, even though they are not displayed in the online docs.

### DIFF
--- a/docs_src/data_block.ipynb
+++ b/docs_src/data_block.ipynb
@@ -6730,6 +6730,16 @@
    ]
   },
   {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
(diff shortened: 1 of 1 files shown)


sgugger:
There is no `self` in the signature of the function in the docs so it's weird to make a reference to it. You can add a short documentation of those functions, but you would need to move them from the section they are in: everything in there is ignored.

EmbraceLife:
> There is no `self` in the signature of the function in the docs so it's weird to make a reference to it. You can add a short documentation of those functions, but you would need to move them from the section they are in: everything in there is ignored.

Thanks! it is very helpful.

sgugger:
Thanks!

PR #2084, opened 2019-05-13
add docs to `CategoryProcessor.create_classes`

EmbraceLife:


sgugger:
Only refers to the argument of the function you have: here it should be "create the inner mapping category name to index (stored in `c2i`) from the `classes`."
But don't reference to `self` when tehre is no `self`.

EmbraceLife:
<p>Thanks a lot!</p>

sgugger:
Thanks!

PR #2085, opened 2019-05-13
Add date feature columns to the original inner_df

herrmann:
The `add_cyclic_datepart` function was adding columns to a new `DataFrame`, instead of the instance at `inner_df`, and this prevented `TabularProc` from introducing new columns (and adding them to `cont_names` and `cat_names`).


sgugger:
Thanks

PR #2086, opened 2019-05-13
add docs to CategoryProcessor.process_one

EmbraceLife:


sgugger:
Same here: there is no `c2i` in the definition of this function. Just say it converts `item` which is a category name to an index.

EmbraceLife:
> it converts `item` which is a category name to an index.

Thanks, yes, this is much simpler and clear.

sgugger:
Thanks!

PR #2092, opened 2019-05-15
Fix nan caused by missing classes in validation

KeremTurgutlu:


sgugger:
See, it passed this time ;). Thanks!

PR #2101, opened 2019-05-19
Add segmentation acc

KeremTurgutlu:
Added non foreground accuracy metric from camvid notebook and 3 random (p,t) paired test for it., which passed.

I wrote random tests using the expect == expected values (which is 1/n_classes) with large batch size for three different number of classes (2, 8, 16) since it would be very space consuming to hard code segmentation predictions.


sgugger:
Thanks!

PR #2103, opened 2019-05-20
Missing .items in __repr__

KaiLicht:
.items is missing in __repr__ of ItemList


sgugger:
No, the idea is to have the get method called for all of these items, which is why there is the index into self instead of self.item: it goes through __getitem__ which in turns calls get.

PR #2107, opened 2019-05-23
Special handling of MixedItemList on add_test

herrmann:
When a `MixedItemList` is `add_test`ed, the current code creates the new instance using its `items`, but this list contains only the indexes. The right attribute in this case would be `item_lists`, which contains the list of `ItemList`s. The patch adds another type check to the method.


sgugger:
Should work, thanks!

PR #2111, opened 2019-05-24
Tensorboard x logdir fix

jantic:
I'm changing the log_dir parameter to be read in as a positional parameter rather than a keyword parameter for SummaryWriter, as the code in TensorboardX was recently changed to use logdir instead of log_dir.  Making it positional will ensure there's no problems with both old and new spellings.


sgugger:
Thanks!

PR #2115, opened 2019-05-26
Interpret

KeremTurgutlu:
## Changes

- Added `interpret.py` module
- Created base class `Intepretation`
- Copied already implemented `ClassificationInterpretation` and `TextClassificationInterpretation`. `ClassificationInterpretation` inherits from `Interpretation` now. Left the original copies in their own modules so that nothing breaks.
- Implemented  `SegmentationInterpretation`, tried to address issues from the forums:https://forums.fast.ai/t/image-segmentation-understanding-inference/45855
- Created skeletons for `MultiLabelClassificationInterpretation` and`ObjectDetectionInterpretation`.

### TODOs

- Implement `MultiLabelClassificationInterpretation` with regards to current `ClassificationInterpretation`, e.g. what would be the multilabel version of it?
- Do literature survey to see how people interpret different tasks which are missing right now
- Brainstorm on how to add test suites for this module
- Start creating documentation for the accepted parts of the PR
- Create a thread in the forums for interpret so that the community can contribute and discuss together on what's needed/desired


sgugger:
Thanks, this looks nicely refactored! The problem I have is that we tried to keep the applications separate but this puts them together in one module. I would suggest to have:
- the base interpretation classes (Intepretation, ClassificationInterpretation...) in either an interpretation module or the train module
- then have an interpretation module in each of the application, which would import the base interpretation module and add the new stuff (segmentation, object detection for image for isntance) or monkey-patch for new methods (show_top_losses for instance). I would remove TextClassificationInterpretation and just monkey-patch ClassificationInterpretation in the text.interpretation module, so that it's all in one object which gets new functionality depending on the applications loaded.

KeremTurgutlu:
Thanks for the quick feedback :)

Let me do these changes and also need to fix one thing with `SegmentationIntepretation`, then I can update the PR.

KeremTurgutlu:
Example [notebook](https://github.com/KeremTurgutlu/experimental/blob/master/SegmentationInterpretation.ipynb) for segmentation interpret.

KeremTurgutlu:
I am adding this change to `def dice()` in `metrics.py`, iou should be 1 if union is 0. Am I missing a point, or was it left without 1 on purpose?

sgugger:
Ok, this is looking good, thanks a lot!
I don't know where that 1. came from for IoU but it seems like a bug, thanks for fixing and adding tests. Merging the progress up to know, don't forget to add documentation of your new feature :)

KeremTurgutlu:
I will add documentation as a next step. Thanks a lot!

pyjaime:
Hey Kerem, 

I saw you changed the ClassificationInterpretation's **probs** parameter to **preds** (not judging that), but didn't change the /fastai/vision/learner.py reference. And I think it isn't the only one problem.

I'll write a quick fix for that (don't know if it will be accepted), and then you could do a proper refactoring with no hurry (maybe with tests). I hope that is right for you.

Thank you!

KeremTurgutlu:
Thanks for fixing that I must have forgot to change references in learner methods. 

- The reason why I changed the name from probs to preds in base `Interpretation` is that predictions might not be always probabilities for some other task. 

- Can you mention the other problems that you are referring to? So that we can quickly fix them. Apart from **probs -> preds** everything should be same.

- I strongly agree with adding tests for interpretation methods.

Thanks a lot!

sgugger:
I have no objection to make probs -> preds as long as nothing is forgotten ;) Make sure notebook 01 runs with the change!

KeremTurgutlu:
I will run the notebooks with all the already existing interpretation classes and also add tests for each, based on those same examples. Sorry for the trouble 😄

pyjaime:
Hello! 
 
>     * Can you mention the other problems that you are referring to? So that we can quickly fix them. Apart from **probs -> preds** everything should be same.

Well, it isn't a different problem... it's just that I found more references. You'll easily see those when refactoring! :)

sgugger:
It takes a ds_type, not a dataset, which is an enum with only 5 cases, so you can't have an error. If you want to use a different dataset, you can easily change your DataBunch attributes with `data.test_dl = bla`

muellerzr:
When I do that for another databunch object (like you suggest making if we want it labeled) the error still pops up. This was working before I moved to the newest dev build, but perhaps I am confused on what you mean. If I generate a new databunch object like such:

```
data_test = (TabularList.from_df(test, path=nbPath, cat_names=cat_var, cont_names=cont_var,
                           procs=procs)
       .split_none()
       .label_from_df(dep_var)
       .databunch())
```

I should be able to run this, correct?

`ClassificationInterpretation.from_learner(learn, data_test.train_dl)`

muellerzr:
I can't replicate this issue on the ADULTS dataset as an example, only my own private dataset I am working with. But both were made the exact same way. Should I move this to a forum post to help debug where I can show the notebook?

sgugger:
No you can't run 
```
ClassificationInterpretation.from_learner(learn, data_test.train_dl)
```
as this method takes a `ds_type` for its second argument, so DatasetType.something
If you want to use a new set, do `data.test_dl = data_test.whatever_dl` then use this method with `ds_type=DatasetType.Test`

In general, please use the forum to debug code as it will benefit more people if it's there.

muellerzr:
Got it! Thank you!

PR #2122, opened 2019-05-28
Fix beam search

hnisonoff:
As discussed on the [forums](https://forums.fast.ai/t/beam-search-prediction-bug-in-text-learner/46485?u=hunter.nisonoff), I've modified the predict and beam_search methods to put the model in eval mode. Additionally, I removed the change in #1598 that makes `top_k` copies of the root node. This line produces incorrect behavior in beam search. I tested this code with the transformerxl and it works as expected.


fastai/text/learner.py
```diff
@@ -118,6 +118,7 @@ def predict(self, text:str, n_words:int=1, no_unk:bool=True, temperature:float=1
         "Return the `n_words` that come after `text`."
         ds = self.data.single_dl.dataset
         self.model.reset()
+        self.model.eval()
```
sgugger: No need for this one, it will be done by `pred_batch` a bit below.

sgugger:
Seems good, just one unnecessary model.eval() to remove and we can merge, thanks!

hnisonoff:
Great. I removed the `model.eval()` in the `predict` method.

hnisonoff:
I updated my branch to be consistent with fastai/master. As a result it looks like this pull request modifies a yml and jupyter notebook file. Am I doing something incorrect, or does github just not update the pull request to keep in sync with the target branch?

sgugger:
All good, thanks a lot!

PR #2125, opened 2019-05-29
change ref probs to preds

KeremTurgutlu:
Thanks to @pyjaime, missing references in monkey patched `learner.py` fixed.


sgugger:
Thanks!

PR #2126, opened 2019-05-29
Update distributed.py

sebastienwood:
Adding a warning message on _learner_parallel if CUDA is not available (in V1.0.52 it continues training on CPU without warning the user). Just a bit of sugar for airheads :smile:


sgugger:
Thanks!

PR #2127, opened 2019-05-30
Added plot_top_losses functionality

muellerzr:
In this PR, I added functionality in the tabular module's ClassificationInterpretation to include showing the top_losses. This is done through a pandas dataframe, with the first four columns becoming Prediction, Actual, Loss, and Probability. There is an option for return_table that allows the user to return the table, and I inherit from the defaults for return_fig for simplicity sake on the return. I am unsure how to go about a test to ensure we have a tabular model, as I know with the image-related models we test for two layer groups. Perhaps I could get advice on this?


muellerzr:
Really I suppose this adds ClassificationInterpretation to tabular, as it was not there originally. Also, would it be worth having a show_all value where we can view the entirety of the losses into a dataframe? Or would that be better suited for a separate show_losses

sgugger:
Thanks!
`show_all` would be a bit too long to actually show anything. If it just returns a dataframe, it would be more of a `df_with_loss` method or something of the like. Not too sure it's as useful?

muellerzr:
Gotcha. Okay! I wasn't sure if a functionality like that would be important. Thanks!!!!!!

muellerzr:
I believe I have an issue. Was probs replaced with preds in the most recent library?

muellerzr:
Looks like it.

sgugger:
Yup, indeed it was (always best to pull master before adding a new feature ;) )

muellerzr:
Oops! My apologies! Do I need to submit another PR for that fix?

sgugger:
If possible, otherwise other people won't be able to use your feature :)

muellerzr:
Done :) Thanks Sylvain!

Issue #2133, opened 2019-06-01
Redundant code in Learner.freeze and Learner.unfreeze


KushajveerSingh:
The `freeze_to` function is defined as 
```python
for g in self.layer_groups[:n]:
   for l in g:
     if not self.train_bn or not isinstance(l, bn_types): requires_grad(l, False)
for g in self.layer_groups[n:]: requires_grad(g, True)
self.create_opt(defaults.lr)
```
Optimizer object is created in the `freeze_to` function itself, whereas `freeze` and `unfreeze` do it again as shown
```python
def freeze(self)->None:
   "Freeze up to last layer group."
   assert(len(self.layer_groups)>1)
   self.freeze_to(-1)
   self.create_opt(defaults.lr)

def unfreeze(self):
   "Unfreeze entire model."
   self.freeze_to(0)
   self.create_opt(defaults.lr)
```

sgugger:
Yes, the optimizer needs to be created again because at creation, it only takes trainable parameters. After a freeze/unfreeze those change so it needs to be adjusted.

KushajveerSingh:
But `freeze_to` does that in the last line `self.create_opt(defaults.lr)`. So how does this happen? In the `freeze` function after calling `self.freeze_to(-1)` no other code is being executed. So we are creating optimizer twice.

sgugger:
Oh I see what you mean. Yes it's unnecessary in freeze and unfreeze.

PR #2136, opened 2019-06-02
Remove redundant code from freeze() and unfreeze()

KushajveerSingh:
learn.freeze() and learn.unfreeze() called freeze_to() function where an instance of opt is created. When using freeze and unfreeze opt is created once more. So the second creation of opt is removed.

This was discussed in Issue #2133


sgugger:
Yup, thanks for fixing!

PR #2138, opened 2019-06-03
Proposal for model_summary update

sebastienwood:
I did a bit of work on model.summary() as discussed in https://forums.fast.ai/t/developer-chat/22363/942?u=sebastienwood
Now the model name, optimizer, loss function and callbacks functions are listed as well. Extra code for AdamW. Would however require to further discuss what and whatnot include, and how to present it to the users.

### DIFF
--- a/fastai/callbacks/hooks.py
+++ b/fastai/callbacks/hooks.py
@@ -186,13 +186,11 @@ def model_summary(m:Learner, n:int=70):
     res += f"Total non-trainable params: {total_params - total_trainable_params:,}\n"
            
     res += f"Optimized with {str(m.opt_func)[25:-1].replace('>', '')}\n"
-    if m.true_wd:
-        res += f"Using true weight decay as discussed in https://www.fast.ai/2018/07/02/adam-weight-decay/ \n"
+    if m.true_wd: res += f"Using true weight decay as discussed in https://www.fast.ai/2018/07/02/adam-weight-decay/ \n"
     res += f"Loss function : {m.loss_func.__class__.__name__}\n"
     res += "=" * n + "\n"
     res += "Callbacks functions applied \n"
-    for cbk in m.callbacks:
-        res += f"    {cbk.__class__.__name__}\n"
+    res += "\n".join([f"    {cbk.__class__.__name__}" for cbs in m.callbacks])
 
     return PrettyString(res)
 
--- a/fastai/callbacks/hooks.py
+++ b/fastai/callbacks/hooks.py
@@ -164,8 +164,6 @@ def layers_info(m:Collection[nn.Module]) -> Collection[namedtuple]:
 
 def model_summary(m:Learner, n:int=70):
     "Print a summary of `m` using a output text width of `n` chars"
-    #TODO docstrings of callbacks appear to be none (bug)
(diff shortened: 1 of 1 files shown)


fastai/callbacks/hooks.py
```diff
@@ -181,6 +184,18 @@ def model_summary(m:Learner, n:int=70):
...
     res += f"Total non-trainable params: {total_params - total_trainable_params:,}\n"
+           
+    res += f"Optimized with {str(m.opt_func)[25:-1].replace('>', '')}\n"
+    if m.true_wd and "adam" in str(m.opt_func).lower():
+        res += f"Using true weight decay as discussed in https://www.fast.ai/2018/07/02/adam-weight-decay/ \n"
+    res += f"Loss function : {m.loss_func.__class__.__name__}\n"
+    res += "=" * n + "\n"
+    res += "Callbacks functions applied \n"
+    for cbk in m.callbacks:
+        if cbk.__doc__ is None:
```
sgugger: I'd remove the calls to __doc__ and only print the names: there is a `doc` function that fastai provides that will show the doc if needed, with the links to the source.

fastai/callbacks/hooks.py
```diff
@@ -181,6 +184,18 @@ def model_summary(m:Learner, n:int=70):
     res += f"\nTotal params: {total_params:,}\n"
     res += f"Total trainable params: {total_trainable_params:,}\n"
     res += f"Total non-trainable params: {total_params - total_trainable_params:,}\n"
+           
+    res += f"Optimized with {str(m.opt_func)[25:-1].replace('>', '')}\n"
+    if m.true_wd and "adam" in str(m.opt_func).lower():
```
sgugger: Don't test for Adam here, the behavior is different for all optimizers, even SGD when there is momentum (as discussed in the same article).

sgugger:
Thanks for tackling this! I left a few comments for changes I think would make it a little bit better.

fastai/callbacks/hooks.py
sebastienwood: Ok ! (I didn't understood that when reading the doc :\)

sebastienwood:
I made a quick update following your comments, it may be worth to precise in the doc that true_wd has impacts outside of Adam ! :) Which makes me think : does the Learner wd parameter erase the opt_func wd parameter ? 
Eg in :
opt_func = partial(optim.SGD, lr=args.lr, momentum=args.mom, nesterov = args.nest, weight_decay=args.wd)
        learn = Learner(data, model, loss_func=nn.CrossEntropyLoss(), metrics=[accuracy], opt_func=opt_func, callback_fns=ShowGraph, true_wd=False)

I could precise it in summary() if it's the case :)

sgugger:
Oh, if the docs are wrong, feel free to suggest a PR to fix them! If you pass a wd to learner, it will be passed at the creation of the optimizer, yes.

sgugger:
Just restyled two lines. Are the TODO still relevant?

sebastienwood:
You are right they're not anymore. I fixed a small bug and added the weight decay flag :)

sgugger:
Looking great, thanks again!

Issue #2143, opened 2019-06-05
Text learner: error in save_encoder function


tchambon:
The RNNLearner (file fastai/text/learner.py) seems to have an issue:

The function save_encoder is using "file" which is an undefined var:
```
def save_encoder(self, name:str):
        "Save the encoder to `name` inside the model directory."
        if is_pathlike(file): self._test_writeable_path()
        encoder = get_model(self.model)[0]
        if hasattr(encoder, 'module'): encoder = encoder.module
        torch.save(encoder.state_dict(), self.path/self.model_dir/f'{name}.pth')
```
I guess the desired behavior is to check if the name var is "pathlike" (string or pathlib object).

When used as is, the error is the following:
```
NameError                          Traceback (most recent call last)
<ipython-input-37-9c134fe50d5b> in <module>()
----> 1 learn.save_encoder('ft_enc')

~/AI/fastaiv1/fastai/fastai/text/learner.py in save_encoder(self, name)
     57     def save_encoder(self, name:str):
     58         "Save the encoder to `name` inside the model directory."
---> 59         if is_pathlike(file): self._test_writeable_path()
     60         encoder = get_model(self.model)[0]
     61         if hasattr(encoder, 'module'): encoder = encoder.module

NameError: name 'file' is not defined
```

After correction it works (I am making a pull request)

sgugger:
Thanks for flagging and fixing!

PR #2144, opened 2019-06-05
fix issue #2143

tchambon:
Fix of issue #2143 Text learner: error in save_encoder function


sgugger:
Thanks, that will teach me to copy-paste without checking if the names are the same...

PR #2151, opened 2019-06-07
Fix #2150

### LINKED ISSUE #2150: About add_metrics
If we add a custom metric through overwriting the LearnerCallback, we should do this:
```python
        self.learn.recorder.add_metric_names(['my_metric'])
    def on_epoch_end(self, last_metrics, **kwargs):
        return add_metrics(last_metrics, [my_metric])
```
Just like:
https://github.com/fastai/fastai/blob/master/fastai/callbacks/mem.py#L49

Right?
now,the logic of add_metric in `basic_train.py > Recorder > on_train_begin`(https://github.com/fastai/fastai/blob/master/fastai/basic_train.py#L456)  is
```python
    def on_train_begin(self, pbar:PBar, metrics_names:Collection[str], **kwargs:Any)->None:
        "Initialize recording status at beginning of training."
        self.pbar = pbar
        self.names = ['epoch', 'train_loss'] if self.no_val else ['epoch', 'train_loss', 'valid_loss']
        self.metrics_names = metrics_names
        self.names += self.metrics_names
        if hasattr(self, '_added_met_names'): self.names += self._added_met_names
        if self.add_time: self.names.append('time')
```
There's a problem if we do `learn.recorder.plot_metrics` because we add a new metric to `Recorder.metrics`, but there is no corresponding metric names in `Recorder.metrics_names`
I think it should be this:
```python
    def on_train_begin(self, pbar:PBar, metrics_names:Collection[str], **kwargs:Any)->None:
        "Initialize recording status at beginning of training."
        #...
        self.metrics_names = metrics_names
        #- self.names += self.metrics_names
        #- if hasattr(self, '_added_met_names'): self.names += self._added_met_names
        if hasattr(self, '_added_met_names'): self.metrics_names += self._added_met_names
        self.names += self.metrics_names
        if self.add_time: self.names.append('time')
```
Am I right?

LLouice:
Fix #2150

### DIFF
--- a/fastai/basic_train.py
+++ b/fastai/basic_train.py
@@ -458,8 +458,8 @@ class Recorder(LearnerCallback):
         self.pbar = pbar
         self.names = ['epoch', 'train_loss'] if self.no_val else ['epoch', 'train_loss', 'valid_loss']
         self.metrics_names = metrics_names
+        if hasattr(self, '_added_met_names'): self.metrics_names += self._added_met_names
         self.names += self.metrics_names
-        if hasattr(self, '_added_met_names'): self.names += self._added_met_names
         if self.add_time: self.names.append('time')
         if not self.silent: self.pbar.write(self.names, table=True)
         self.losses,self.val_losses,self.lrs,self.moms,self.metrics,self.nb_batches = [],[],[],[],[],[]


sgugger:
Thanks for flagging and fixing!

Issue #2153, opened 2019-06-07
Possible optimization in from_df method of data_block.py


dejanbatanjac:
This is an optimization, so minor issue.  
This specific part of the method [`from_df`](https://github.com/fastai/fastai/blob/master/fastai/data_block.py#L135):

```
inputs.isna().sum().sum() == 0
```
May be optimized with `not inputs.isna().any().any() `, since pandas v.0.23.2 (very old version of pandas)

I tested like this:
```
from fastai.vision import *
path = untar_data(URLs.MNIST_TINY)
import pandas as pd
df = pd.read_csv("C:/Users/dj/.fastai/data/mnist_tiny/labels.csv")
df.iloc[0,0]=2 #np.nan
print(df.head())
print(df.isna().any().any())
assert df.isna().sum().sum() == 0, f"You have NaN values in column(s)"
assert not df.isna().any().any(), f"You have NaN values in column(s)"

%timeit df.isna().sum().sum() == 0
%timeit not df.isna().any().any()
```
And got the result:

```
601 µs ± 5.92 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
499 µs ± 4.91 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```
So minor speed up in the second case.

sgugger:
By all means, please go ahead and propose a PR with this change!

PR #2155, opened 2019-06-08
Optimization based on https://github.com/fastai/fastai/issues/2153

dejanbatanjac:
Based on the issue (in title) we are using any() instead of sum() to check if there are any "na" values in our dataframe. This includes checks on: np.nan, np.nat, and None.


sgugger:
Thanks!

PR #2161, opened 2019-06-11
Download load branch

Rimmavane:
Added functionalities to download from internet and load COCO annotations and images sets as FastAI DataBunch using data block API. 
Added class for loading a Video and iterate by its frames, as well as getting some basic info about the video itself.


sgugger:
Hi there, this is great work! I am closing this however, because this should be in a separate repo, not in the core fastai library. Once you have created it, please ping me with the link to it as we would love to link to if from the docs or on the forum and give it the best visibility we can since it's really great!

Rimmavane:
I am sorry, I pulled to the wrong fork. Thanks for the feedback tho!

PR #2162, opened 2019-06-12
Add random seeds option for data.vision.from_** apis

Joldnine:
Currently, when we create data bunch by data.vision api, we cannot specify random seed for the valid splitting dataset, which means every time we create data bunch (even from the same dataset), we always get different training/validation datasets.

 It is not so reasonable if we run our training in an jupyter notebook, where we need always re-initialise our dataset's training/validation split for the same training task due to OOM, disconnection, etc. 

So it is better to allow users to specify the training/validation data split seed by themselves.


sgugger:
Yes, the idea was to have people use the data block API for that, but exposing them in the higher factory methods is fine. Thanks!

PR #2166, opened 2019-06-12
add solution to bug caused by old version

Light--:
add solution to the problem: "ModuleNotFoundError: No module named ‘fastai.vision’"


sgugger:
Sounds good!

stas00:
But why did you remove the very useful diagnostics information? There are many ways to mess things up and `pip` is not always going to help. 

Seeing the exact path fastai is loaded from is going to be very helpful, so I recommend you restore the diagnostics part, with the extra information you added.

PR #2168, opened 2019-06-13
WIP: Add support for training sampler

lewfish:
This PR adds support for a sampler to use for the training set `DataLoader`. This was motivated by wanting to use a `WeightedRandomSampler` to support oversampling of rare classes. The sampler can't simply be passed in as a `kwarg` to `databunch()` because we only want to pass it to the `train_dl`. Also, when specifying a sampler, we need to turn off shuffle for the `train_dl` because that option is not compatible with setting a sampler -- PyTorch won't allow it.

### TODO
- [ ] Add unit tests


sgugger:
Don't forget to run the tests locally (with `pytest` in your cloned copy of the fastai repo) as you're breaking existing behavior. It looks like you need to also adapt the create methods for the various `TextDataBunch`

lewfish:
I won't be able to work on this for a while, so I'm going to close this and re-open it later.

sgugger:
Sure, take the time you need!

PR #2169, opened 2019-06-13
Remove redundant import

KushajveerSingh:
`from .image import _affine_mult` is a redundant import


sgugger:
Not it's not: import * only gives you what is insiade the __all__ varialbe of that modulem and _affine_mult isn't in it.

PR #2179, opened 2019-06-24
Update vision.learner.html

dblclik:
docs reference "arch" in cnn_learner description instead of "base_arch" which is the actual variable name in the cnn_learner source


sgugger:
Thanks for your PR, but please change the notebook `vision.learner.ipynb` and not the HTML file: the docs are auto-generated from the notebooks so if I merge your PR, the change will be erased at the next build.

PR #2180, opened 2019-06-24
Update vision.learner.ipynb

dblclik:
Update vision.learner.ipynb to change references from `arch` to `base_arch` in `cnn_learner()`.  Also updated one typo, changing `lin_frts` to `lin_ftrs`


sgugger:
Thanks a lot!

PR #2184, opened 2019-06-25
Update dependencies.svg swapping basic_data and data_block dependency

dataista0:
Swap the dependency between data_block and basic_data in the diagram at the bottom of the docs' main page (index.html). data_block depends on basic_data, not the other way. The change is transparent to index.ipynb.


sgugger:
True, thanks for flagging!

PR #2185, opened 2019-06-26
Unet fix

feras-oughali:
Fix an edge case for DynamicUnet to handle input with odd sizes. The error happens when the size of the input image is an odd number. 
I added the exact input size as a parameter when creating the model. This can be fed dynamically while creating a `unet_learner` by accessing the size of the first element of the training dataset.

### DIFF
--- a/fastai/vision/learner.py
+++ b/fastai/vision/learner.py
@@ -113,7 +113,11 @@ def unet_learner(data:DataBunch, arch:Callable, pretrained:bool=True, blur_final
     "Build Unet learner from `data` and `arch`."
     meta = cnn_config(arch)
     body = create_body(arch, pretrained, cut)
-    model = to_device(models.unet.DynamicUnet(body, n_classes=data.c, img_size=data.train_ds[0][0].size, blur=blur, blur_final=blur_final,
+    try:
+        size = data.train_ds[0][0].size
+    except:
+        size = next(iter(data.train_dl))[0].shape[-2:]
+    model = to_device(models.unet.DynamicUnet(body, n_classes=data.c, img_size=size, blur=blur, blur_final=blur_final,
           self_attention=self_attention, y_range=y_range, norm_type=norm_type, last_cross=last_cross,
           bottle=bottle), data.device)
     learn = Learner(data, model, **learn_kwargs)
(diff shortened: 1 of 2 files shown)


sgugger:
Thanks! I think `unet_leaner` should also take a size argument to overwrite the result of `data.train_ds[0][0]`. Also we should do this a try except block (that would end with size=None) because if someone created their DataBunch from PyTorch dataloaders, it won't have a `train_ds`.

feras-oughali:
Will something like this be okay if the DataBunch was created from PyTorch dataloaders?
```python
next(iter(data.train_dl.dl))[0].shape[-2:]
```

sgugger:
Yes, good idea! But test the dataset first as it's faster. Also, no need for the `.dl`, just `data.train_dl` should be enough I think.

feras-oughali:
Yes, `data.train_dl` is sufficient.

```python
def unet_learner(data:DataBunch, arch:Callable, pretrained:bool=True, blur_final:bool=True,
                 norm_type:Optional[NormType]=NormType, split_on:Optional[SplitFuncOrIdxList]=None, blur:bool=False,
                 self_attention:bool=False, y_range:Optional[Tuple[float,float]]=None, last_cross:bool=True,
                 bottle:bool=False, cut:Union[int,Callable]=None, **learn_kwargs:Any)->Learner:
    "Build Unet learner from `data` and `arch`."
    meta = cnn_config(arch)
    body = create_body(arch, pretrained, cut)
    try:
        size = data.train_ds[0][0].size
    except:
        size = next(iter(data.train_dl))[0].shape[-2:]
    model = to_device(DynamicUnet2(body, n_classes=data.c, img_size=size, blur=blur, blur_final=blur_final,
          self_attention=self_attention, y_range=y_range, norm_type=norm_type, last_cross=last_cross,
          bottle=bottle), data.device)
    learn = Learner(data, model, **learn_kwargs)
    learn.split(ifnone(split_on, meta['split']))
    if pretrained: learn.freeze()
    apply_init(model[2], nn.init.kaiming_normal_)
    return learn
````
is it ok now?

> I think `unet_leaner` should also take a size argument to overwrite the result of `data.train_ds[0][0]`

is this comment still valid? I don't see why you would want to overwrite the size.

sgugger:
If we always have a way to get it right, we don't need an argument, no.

sgugger:
Thanks a lot!

feras-oughali:
Appreciate your help :)

PR #2186, opened 2019-06-26
Epoch as MLflow step number in metrics

herrmann:
The `step` parameter was missing when `log_metric` is called.


sgugger:
Thanks!

PR #2188, opened 2019-06-27
Oversampling Callback

tmabraham:
Adding a callback for oversampling. Currently does this by changing the sampler to be a weighted random sampler with weights corresponding to 1/counts of the classes. Previously tested on MNIST dataset.

### DIFF
--- a/fastai/callbacks/__init__.py
+++ b/fastai/callbacks/__init__.py
@@ -8,3 +8,4 @@ from .rnn import *
 from .tracker import *
 from .csv_logger import *
 from .loss_metrics import *
+from .oversampling import *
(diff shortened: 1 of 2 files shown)


sgugger:
Thanks a lot! There is a little thing to change IMO: having the learner as an attribute creates a cyclical reference (the learner has a reference to the callback and the callback one to the learner) which causes memory leak issues.
You should use `LearnerCallback` as superclass then uncomment `super().__init__(learn)` in your init, the `LearnerCallback` class is specifically there to remove that problem by using a weak reference to the `Learner`.

tmabraham:
@sgugger I have fixed this. Hope this is fine.

sgugger:
Perfect, thanks!

PR #2189, opened 2019-06-28
Changed tvm to models in the example

swagato-c:
The example used tvm from the older versions. Changed it to models


sgugger:
Thanks!

PR #2191, opened 2019-06-30
Fix rand state in filter_by_rand()

cudawarped:
Currently filter_by_rand() sets the seed for the numpy random generator before calling rand_bool().  This has no effect because rand_bool() relies on random numbers generated by either tensorflow or random.
  
This fix is a little unsightly but I am guessing everything will change soon when a new version is out so it should do until then.

On a separate issue, is it a good idea to keep changing the global random number generator seeds, couldn't this lead to unintentionally removing the randomness from the learning process, which as Jeremy pointed out provides a good mechanism for testing for bugs.


sgugger:
Thanks for spotting that. One change I would make is to make `set_uniform_seed` set the seed received (leave the if None test in the function filter_by_rand) and set all the seeds (torch, torch.cuda, random and numpy) since it could then be used by someone wanted to launch a reproducible training.

Do you mind adding that yo your PR or do your prefer I merge and apply that change afterward?

cudawarped:
Hi, no that's not a problem, if I set the numpy seed as well should I rename the function to something more appropriate like set_random_seed()?

Is it intentional then to set global seeds every time instead of creating generators and passing them around?  That is if I pass a seed to filter_by_rand() I might just want the same data set but not want the same batch data loading order, model weight initialization etc. which are also randomly generated.

I think that now [torch.manual_seed()](https://pytorch.org/docs/stable/notes/randomness.html) sets the seed for all devices have you experienced anything different?

sgugger:
I think `set_all_seed` would be even more appropriate. I don't have use cases for `filter_by_rand` in mind, but in any case it needs setting the seed for everything apart from numpy as you pointed out, so if someone wants randomness, they can set the seed to a random number after collecting the data. 

Didn't torch.manual_seed was doing everything PyTorch related now, so if that's the case, no need to add something else.

cudawarped:
Added the changes.

My use case for `filter_by_rand ` was to get a sample of a larger data set to run some quick tests, is this the correct function to use?   The reason I asked about global seeds is because I wanted the same sample set for each run but I didn't really want to seed all the other prng's which are used throughout the fastai library.  It probably doesn't matter I just remember Jeremy mentioning that it is useful for finding bugs if you have random initialization.

cudawarped:
Sorry I forgot to rename the function as you suggested, apologies for killing the build bot.

sgugger:
Np, `filter_by_rand` is indeed the function you would want for your use case.

PR #2192, opened 2019-06-30
added epsilon to avoid numerical instability in precision and recall …

antoinebon:
…computations

### DIFF
--- a/fastai/metrics.py
+++ b/fastai/metrics.py
@@ -166,7 +166,7 @@ class CMScores(ConfusionMatrix):
     eps:float=1e-9
 
     def _recall(self):
-        rec = torch.diag(self.cm) / (self.cm.sum(dim=1) + self.eps)
+        rec = torch.diag(self.cm) / self.cm.sum(dim=1)
         if self.average is None: return rec
         else:
             if self.average == "micro": weights = self._weights(avg="weighted")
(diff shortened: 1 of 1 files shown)


sgugger:
Thanks for the PR. Since we're summing integers, it's not really numerical stability we're talking about. I'm open to have a test for the sum 0 and a sensible result (I do think NaN is a good value since precision/recall are just not defined in those situations though) but adding an eps to the denominator isn't a right solution IMO.

Also, not that if you get an error it means there is not a single item of the measured class in the whole validation set (metric is computed after going through the whole thing) so it's really weird to have a 0 sum at that point, which is why the user should also get a weird result.

antoinebon:
Thanks for the feedback. So indeed for recall, having a null sum probably means there is an issues with the validation set. However for precision, if one class end-up never been predicted, then the sum will be zero for that class. I ran into this issue with while training a classifier on a very skewed dataset, where some of the classes have very few samples. Because of that the averaged precision is NaN. Then maybe the eps only makes sense for the precision case, in order to at least have a feel of the precision for the rest of the classes.

sgugger:
I just checked and scikit learn returns 0 and issues a warning in those cases, so I think we should mimic that behavior.

antoinebon:
OK makes sense. Thanks

PR #2195, opened 2019-07-02
Added a `sep` attribute to `TextList` that gets passed to `textify` in `get` calls.

StatisticDean:
Based on this forum post : https://forums.fast.ai/t/separator-in-textify-should-a-sep-attribute-be-added-to-vocab-class/49454/2


sgugger:
Thanks a lot!

PR #2196, opened 2019-07-03
Update performance.md

michaeltlu:
Hi -- I've modified the conda and pip uninstall commands to also uninstall previous versions of libjpeg-turbo.


sgugger:
Thanks

PR #2197, opened 2019-07-04
PoolingLinearClassifier: factor out stateless concat pool logic

sshleifer:
This facilitates using fastai as a sentence encoder without a linear classifier.


sgugger:
Thanks!

PR #2198, opened 2019-07-04
fix grammatical error in vision.learner Transfer Learning docs

gndowns:
Simple grammar fix, the old wording did not makes sense.


sgugger:
Thanks!

PR #2201, opened 2019-07-05
Specify Which torchvision models are included in vision.models docs

gndowns:
The current vision.models [docs](https://docs.fast.ai/vision.models.html#Computer-Vision-models-zoo) do not specify which models are included from the torchvision model zoo. 

This is problematic, since some of torchvision's models are *not* imported by fastai e.g. inception_v3, vgg16 (no batch norm).

Currently the only reference for which models are imported is the direct source: 
https://github.com/fastai/fastai/blob/master/fastai/vision/models/__init__.py

This PR gives a comprehensive list of all models that can be imported via `models.<model_name>` directly in the docs.

### DIFF
--- a/docs_src/vision.models.ipynb
+++ b/docs_src/vision.models.ipynb
@@ -24,7 +24,16 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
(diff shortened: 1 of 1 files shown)


sgugger:
Thanks a lot!

PR #2205, opened 2019-07-08
Remove split_on from create_cnn_model

rohitgr7:
split_on parameter is not used while creating the cnn_model in create_cnn_model().


sgugger:
Ah true, it's only used in `cnn_learner`. Thanks for flagging!

PR #1881, opened 2019-03-25
AUC metric

SvenBecker:
Added the option to use AUC based on ROC for the binary classification task.
This implementation is very similar to the one provided by Sklearn but everything is done in PyTorch instead of Sklearn/NumPy.

Based on https://forums.fast.ai/t/use-auc-in-place-of-accuracy-as-loss/41168/3 or https://forums.fast.ai/t/using-auc-as-metric-in-fastai/38917/7 I thought it might be a nice feature to have.

PR #2211, opened 2019-07-09
Documentation for ROC-related metrics

Huizerd:
Added documentation for metric methods/classes introduced by #1881, referring to sklearn and Wikipedia. I was wondering whether it might be a better idea to have `roc_curve` and `auc_roc_score` as methods to the `AUROC` class, given that they are not meant to be passed directly to the `metrics` argument of your learner?

### DIFF
--- a/docs_src/metrics.ipynb
+++ b/docs_src/metrics.ipynb
@@ -1033,7 +1033,73 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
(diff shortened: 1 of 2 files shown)


sgugger:
Thanks, this is a step in the right direction. The functions `roc_curve` and `auroc_curve` can be useful on their own, it's just that they are only appropriate if you have all your predictions and targets available (for instance after a `get_preds`), they are not suitable to be passed to `Learner` directly as you pointed out however.

PR #2216, opened 2019-07-11
Added the sep attribute to the copy_new attribute of TextList

StatisticDean:
Following the PR https://github.com/fastai/fastai/pull/2195
The `sep` attribute was not passed to copy_new, as a result, when new was called (as for example in show_batch), the default separator would be used. Adding `'sep'` to the copy_new attributes fixes this.


sgugger:
Good catch, thanks!

PR #2220, opened 2019-07-13
Fix oversampling

tmabraham:
There was an error with the oversampling callback, and did not perform as I originally intended. 
It currently redistributed the dataset so that all classes have same frequencies but the size of the dataset is the same. So some samples were omitted from the final dataset.

This version now duplicates the samples so that all classes have the same number of samples as the class with the highest frequency.


sgugger:
Thanks!

PR #2223, opened 2019-07-15
Added se_resnext101_32x4d

swagato-c:
Added se_resnext101_32x4d which is now present in cadene's repository


sgugger:
Looking good, thanks!

PR #2224, opened 2019-07-15
Add activ to get_preds

rohitgr7:
PR for: https://forums.fast.ai/t/shouldnt-we-able-to-pass-an-activ-function-to-learner-get-preds/50492


sgugger:
Thanks a lot!

PR #2227, opened 2019-07-16
ClassConfusion

muellerzr:
ClassConfusion is a widget for further analyzing how both tabular and image models perform and operate, and to scientifically assess their behavior. For tabular models we analyze the trends in the outliers of the data, looking at the relative distributions between two confused classes and the original distribution. This let's us paint a picture to how our data is being used. If we are worried about how many unique names we have in a categorical dataset, `cut_off` is there to make sure we do not exceed our limit.

For images, we can pass in a set of classes and see what the worst-performing or most-confused images were, and we can also see their filenames if we want a quick deletion. I will be working on the documentation later tonight. But here are some images of the results. If we have an ordered set of pairs to look at, we adjust the `is_ordered` value to True on the constructor. 

Tabular
Pets

Let me know what needs to be adjusted or changed.

### DIFF
--- a/fastai/widgets/class_confusion.py
+++ b/fastai/widgets/class_confusion.py
@@ -27,6 +27,12 @@ class ClassLosses():
         self.classl = classlist
         self._show_losses(classlist)
 
+        
+    def _show_losses(self, classl:list, **kwargs):
+        "Checks if the model is for Tabular or Images and gathers top losses"
+        _, self.tl_idx = self.interp.top_losses(len(self.interp.losses))
+        self._tab_losses() if self._is_tab else self._create_tabs()
+        
     def _create_tabs(self):
         "Creates a tab for each variable"
 
@@ -54,6 +60,15 @@ class ClassLosses():
         for i in range(len(items)):
             self.tabs.set_title(i, self.tbnames[i])
         self._populate_tabs()
+        
+    def _populate_tabs(self):
+        "Adds relevant graphs to each tab"
+        with tqdm(total=len(self.tbnames)) as pbar:
+            for i, tab in enumerate(self.tbnames):
+                with self.tabs.children[i]:
+                    self._plot_tab(tab) if self._is_tab else self._plot_imgs(tab, i)
+                pbar.update(1)
+        display(self.tabs)
         
     def _plot_tab(self, tab:str):
         "Generates graphs"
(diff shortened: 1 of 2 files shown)


muellerzr:
Not 100% sure why those checks failed, doesn't seem related to my functionality though if I'm not mistaken?

fastai/widgets/class_confusion.py
```diff
@@ -0,0 +1,226 @@
...
+import pandas as pd
+import matplotlib.pyplot as plt
+
+from itertools import permutations, combinations
+from fastai.train import ClassificationInterpretation
+import ipywidgets as widgets
+
+class ClassLosses():
+    
+    """Plot the most confused datapoints and statistics for your misses.
```
sgugger: Just give a one-line docstring here, then put the rest in the documentation of that widget (notebook to come after). Also, try to give a one-line docstring for each of the methods of this class (or make them private if the user isn't supposed to use them).

fastai/widgets/class_confusion.py
```diff
@@ -0,0 +1,226 @@
...
+        display(self.tabs)
+        
+    def show_losses(self, classl:list, **kwargs):
+        if str(type(self.interp.learn.data)) == "<class 'fastai.tabular.data.TabularDataBunch'>":
+            self.tab_losses(classl)
+        else:
+            self.im_losses(classl)
+
+    def im_losses(self, classl:list, **kwargs):
+        if self.is_ordered == True:
```
sgugger: Could be shortened with `lis = bla if condition else bli`

fastai/widgets/class_confusion.py
```diff
@@ -0,0 +1,226 @@
...
+                plt.show(fig)
+                plt.tight_layout()
+        display(self.tabs)
+
+    def tab_losses(self, classl:list, **kwargs):
+        tl_val, tl_idx = self.interp.top_losses(len(self.interp.losses))
+        classes = self.interp.data.classes
+        cat_names = self.interp.data.x.cat_names
+        cont_names = self.interp.data.x.cont_names
+        if self.is_ordered == False:
```
sgugger: Could be shortened with `comb = bla if condition else bli`

muellerzr:
I have made the changes you suggested, along with discovered and fixed a bug with the image display, and some spelling errors

muellerzr:
Don't push just yet though please, I realized what you meant now by the one liners for those checks, fixing now :)

muellerzr:
Okay now it's all done :) Let me know any further changes you would like or need!

muellerzr:
One afterthought though is is there a better way to check for a tabular databunch type in python that you know of? This would help in the future if in V2 you guys change that.

sgugger:
YEah your way is a bit clunky. A better way would be to use isinstance:
```
isinstance(interp.learn.data, TabularDataBunch)
```

muellerzr:
Got it. Thank you! Did not know quite how to do the python check for that.

sgugger:
Thanks! Don't forget to write docs for it now :)

PR #2229, opened 2019-07-17
Failed to export classes for TextClasDataBunch.from_ids

troilusc:
TextClasDataBunch generated by `TextClasDataBunch.from_ids` has not set up the CategoryProcessor, so the info of classes get lost when export and load it again.

Let me know what needs to be adjusted or changed (My first pull request, so I assume there will a lot.).

### DIFF
--- a/tests/test_text_data.py
+++ b/tests/test_text_data.py
@@ -188,14 +188,13 @@ def test_from_ids_works_for_variable_length_sentences():
 
 def test_from_ids_exports_classes():
     this_tests(TextClasDataBunch.from_ids)
-    ids = [np.array([0]),np.array([0,1])]*5 # notice diffrent number of elements in arrays
+    ids = [np.array([0])]*10
     lbl = [0]*10
     data = TextClasDataBunch.from_ids('/tmp', vocab=Vocab({0: BOS, 1:PAD}),
                                       train_ids=ids, train_lbls=lbl,
                                       valid_ids=ids, valid_lbls=lbl, classes={0:0}, bs=8)
     data.export('/tmp')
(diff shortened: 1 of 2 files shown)


tests/test_text_data.py
```diff
@@ -186,6 +186,16 @@ def test_from_ids_works_for_variable_length_sentences():
                                       train_ids=ids, train_lbls=lbl,
                                       valid_ids=ids, valid_lbls=lbl, classes={0:0}, bs=8)
 
+def test_from_ids_exports_classes():
+    this_tests(TextClasDataBunch.from_ids)
+    ids = [np.array([0])]*10
+    lbl = [0]*10
+    data = TextClasDataBunch.from_ids('/tmp', vocab=Vocab({0: BOS, 1:PAD}),
+                                      train_ids=ids, train_lbls=lbl,
+                                      valid_ids=ids, valid_lbls=lbl, classes={0:0}, bs=8)
```
sgugger: Here classes should be a list, like ['a', 'b', 'c']

tests/test_text_data.py
```diff
@@ -186,6 +186,16 @@ def test_from_ids_works_for_variable_length_sentences():
...
+    data.export('/tmp/export.pkl')
+    TextClasDataBunch.load_empty('/tmp').classes
```
sgugger: Make an assert to test we get back the same thing as we passed before.

troilusc:
Thanks for your suggestion! I've updated the test accordingly.

sgugger:
Thanks a lot!

troilusc:
My pleasure!

PR #2231, opened 2019-07-18
Quick ClassConfusion fix and Documentation

muellerzr:
Tested it on images real quick and realized I had forgot something. My bad!

### DIFF
--- a/docs_src/widgets.class_confusion.ipynb
+++ b/docs_src/widgets.class_confusion.ipynb
@@ -22,7 +22,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 3,
+   "execution_count": null,
    "metadata": {},
    "outputs": [],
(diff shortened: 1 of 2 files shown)


muellerzr:
Finished the documentation too :) Let me know if there are issues. I wasn't 100% sure if I could include the other repository bit. The goal was to make it available, if there's a better way to go about that (like just having it on the forum and lead to the forum post instead let me know!)

muellerzr:
(Promise we're good to go here, triple quadruple checked myself :) )

sgugger:
Don't forget to run `tools/run-after-git-clone` in the repo because your notebooks aren't stripped out of their metadata.

muellerzr:
Fixed the stripout issue.

sgugger:
Looking good, thanks a lot!

PR #2236, opened 2019-07-21
add delimiter to ImageList.from_csv

KeremTurgutlu:
Adding delimiter to `ImageList.from_csv` method to be consistent with other `from_csv` methods in data block API. This will allow to create a image list from csv files with different delimiters than the pandas default, e.g. '\t'.  

Ran `pytest test_data_block.py` for testing.


sgugger:
Thanks!

Issue #2235, opened 2019-07-19
Move registering `callback_fns` with Learner from `.fit()` to `__post_init__()`.


pechyonkin:
**TLDR**: when trying to run `.get_preds()` right after creating a learner, `callback_fns` are not working because they are registered with the learner in the `.fit()` function. I propose to move registering `callback_fns` to initialization of learner so that callbacks can work even without training first.

I am trying to use a pretrained model and analyze activations for some data; in other words, I am not using training, only predicting from an existing model.

I create the data, then the learner, and a test hook callback that prints when a batch is done:

```python
class TestHook(HookCallback):
    def hook(self, m, i, o):
        pass
    def on_batch_end(self, train, **kwargs):
        print("Batch Done!")
        
learner = cnn_learner(data, models.resnet18, callback_fns=TestHook)
```

When I run `learner.fit_one_cycle(1)`, the callback is invoked after each batch and `"Batch Done!" is printed correctly.

However, when I run `learner.get_preds(data.train_ds)`, nothing happens. After digging into the source code, I found the reason: callbacks are registered with the learner in the [`fit()`](https://github.com/fastai/fastai/blob/bd430ef40e7759057aec3e607e7f7834fe26a110/fastai/basic_train.py#L198) function. This means that after creating the learner, callbacks are still not registered and therefore will not work when running `.get_preds()` immediately after creating the learner.

I was able to solve this the following way, and callback works correctly, but this looks too hacky:

```python
class TestHook(HookCallback):
    def hook(self, m, i, o):
        pass
    def on_batch_end(self, train, **kwargs):
        print("Batch Done!")     
        
learner = cnn_learner(data, models.resnet18)
test_hook_callback = TestHook(learner)
learner.callbacks += [test_hook_callback]

learner.get_preds(data.train_ds)
```

Please let me know if this is worth pursuing. If yes, I can try to change the code and create a pull request.

Some more context is in this [post](https://forums.fast.ai/t/callbacks-for-prediction-step/50667) of fastai forums.

sgugger:
We can certainly try to move this to post_init and see if it breaks anything. It would make more sense to have them initialized there.

pechyonkin:
I will work on it and will submit a PR when ready.

pechyonkin:
I did some research and just moving callback registering creates conflicts and failing tests. I have described some of them [on the forums](https://forums.fast.ai/t/moving-registering-callback-fns-with-learner-from-fit-to-post-init/50946).

PR #2237, opened 2019-07-21
Register `callback_fns` in `.get_preds()` if they haven't been registered previously.

pechyonkin:
defaults.extra_callbacks should be registered in __post_init__
in order to allow using them with Learner without training.

This is to address [this issue](https://github.com/fastai/fastai/issues/2235). It is only a small part of the problem. With this small fix, all existing tests pass.

When I tried to move registering other `callback_fns` to `__post_init__`, some problems arose, that I discussed [here](https://forums.fast.ai/t/moving-registering-callback-fns-with-learner-from-fit-to-post-init/50946).

### DIFF
--- a/fastai/basic_train.py
+++ b/fastai/basic_train.py
@@ -169,6 +169,7 @@ class Learner():
         self.callbacks = listify(self.callbacks)
         if self.silent is None: self.silent = defaults.silent
         self.callback_fns = [partial(Recorder, add_time=self.add_time, silent=self.silent)] + listify(self.callback_fns)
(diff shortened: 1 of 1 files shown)


pechyonkin:
In the second commit (fe0cdba) I introduced a boolean flag `cb_fns_registered`. It is used in `get_preds()` to run `callback_fns` registration with the learner if that has not been done previously. For example, when `get_preds()` is called immediately after creating the learner.

Resolves: #2235

sgugger:
Ok this seems a nice temp fix. I'm not sure I'll investigate the more general issue as we have v2 coming soon and it will initialize callback funcs in init.

PR #2245, opened 2019-07-25
remove two transpose for one in image.pil2tensor

MLSlayer:
I'm not sure if this is going to bug out, but I thought that the operations
```
    a = np.transpose(a, (1, 0, 2))
    a = np.transpose(a, (2, 1, 0))
```
can become just 
`    a = np.transpose(a, (2, 0, 1))`

Maybe the tests pipeline tests will tell me if it's wrong, but I think its worth mentioning


sgugger:
It seems like a waste of code, but it's actually faster this way (we did run benchmarks), so we're not changing it until we are proved wrong :)

MLSlayer:
just what I was looking for :)

PR #2248, opened 2019-07-27
variable `do_blur` unused when creating `UnetBlock` in DynamicUnet Module

pvnieo:
In the `DynamicUnet` Class, while creating the decoder part of the unet network, we created a `do_blur` variable that checks if we want to apply the blur (via the `blur` parameter) and also, if the layer is a final layer or not, but it is **not used** in instantiating the `UnetBlock`, and instead, the `blur' parameter is used !


pvnieo:
In the `DynamicUnet` Class, while creating the decoder part of the unet network, we created a `do_blur` variable that checks if we want to apply the blur (via the `blur` parameter) and also, if the layer is a final layer or not, but it is **not used** in instantiating the `UnetBlock`, and instead, the `blur' parameter is used !

sgugger:
Good catch, thanks!

PR #2249, opened 2019-07-27
Update train.py

AniTho:
There was an error raised by the function fit_one_cycle which is OSError: Image file is truncated after creating a custom dataset using Google. The problem wasn't solved by verify_images or using LOAD_TRUNCATED_IMAGES from PIL library in jupyter notebook.
Screenshot (390)
Screenshot (391)


sgugger:
This can't be added in train as this is generic and used in all applications (text and tabular). You should put this in vision.image

PR #2252, opened 2019-07-30
dice metric: fix for empty images and batch size greater than one

andreikop:
I found that [Dice](https://en.wikipedia.org/wiki/S%C3%B8rensen%E2%80%93Dice_coefficient) is calculated incorrectly for batch sizes > 1 and images which are greater than 1 pixel.
Fixed by this PR

I also found that IOU has 2 bugs. Don't have time to fix and properly test now, but marked in the code with FIXMEs. I can open bug reports when this PR is reviewed.

### DIFF
--- a/fastai/metrics.py
+++ b/fastai/metrics.py
@@ -55,10 +55,45 @@ def dice(input:Tensor, targs:Tensor, iou:bool=False, eps:float=1e-8)->Rank0Tenso
     n = targs.shape[0]
     input = input.argmax(dim=1).view(n,-1)
     targs = targs.view(n,-1)
-    intersect = (input * targs).sum().float()
-    union = (input+targs).sum().float()
-    if not iou: return (2. * intersect / union if union > 0 else union.new([1.]).squeeze())
-    else: return (intersect / (union-intersect+eps) if union > 0 else union.new([1.]).squeeze())
+    if iou:  # TODO move IOU to separate API function?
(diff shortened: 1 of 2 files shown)


sgugger:
Thanks a lot for your PR. We don't comment code in fastai like you did however, and leave all those bits you explained for the [docs](https://docs.fast.ai/). Could you amend your PR to only leave the code on its own in metrics.py and add the details around the formula in `docs_src/metrics.ipynb`? Thanks in advance.

andreikop:
@sgugger , I'm not sure this comments should be put to the docs?

If I understand correctly docs are intended for API users to explain how to use the metric function and how value is calculated.
And this comments rather explain to library developers why the code is written this way.
To make the docs more clear we can add a link to Wikipedia article and a note that Dice = 1 if both label and predicted mask is empty.

sgugger:
Explaining the code explains how the function works and how the value is computed, so this is great for the documentation. fastai is primarily aimed at coders, not mathematicians, so they will probably prefer your comments to the wikipedia page.

As for the source code, not including any code comment and limiting vertical space is pretty much the only part of our [style guide](https://docs.fast.ai/dev/style.html) we strictly enforce.

sgugger:
Closing for now since no news in a week and there are needed changes. Please don't hesitate to reopen (or open a new PR) with the requested changes.

PR #2254, opened 2019-07-31
Fixing ram overload issue

schwobr:
Fixing RAM overload that occurred when `iteration % self.loss_iters == 0` or `if iteration % self.hist_iters == 0`. At validation time, the writer kept requesting to push the same loss values and histograms that got stuck in queue.


sgugger:
Thanks!

PR #2256, opened 2019-08-01
Update vision.models.html Repeated "ResNet50"

c2js:
Spelling correction. Remove extra repeated word of "resnet50"


sgugger:
Thanks for your PR, but as mentioned [here](https://docs.fast.ai/gen_doc_main.html), the docs are auto-generated from the notebooks in docs_src. If I merge your PR, the change will be overwritten next time we build the docs.
Would you mind making a new PR with the file vision.models.ipynb changed instead?

PR #2257, opened 2019-08-01
Allow to override defaults of special case in Tokenizer with an empty list.

StatisticDean:
Following : https://forums.fast.ai/t/tokenizer-with-no-special-cases/51627

At the moment calling `Tokenizer(special_cases=[])` will create a Tokenizer whose `special_cases` attribute will be overriden by defaults, this PR changes that by using default only if special_cases is set to `None` (its default value).


sgugger:
Thanks a lot! (tests fail for another reason, fixed in an additional commit)

PR #2261, opened 2019-08-04
fix typo in PixelShuffle_ICNR

pvnieo:
In PixelShuffle_ICNR, the `blur` parameter is not used, because we create the variable `self.blur = nn.AvgPool2d(2, stride=1)` and we use it to check if we want to use blur, but the latter is always true.


sgugger:
Good catch, thanks!

PR #2265, opened 2019-08-07
Remove variable which was assigned but never used

StatisticDean:
`ds` is assigned but never used in the call of `beam_search` and `predict` in the `LanguageLearner` class in fastai.text.learner.

:relieved: Vertical space !


sgugger:
It was probably used in an earlier version but then we forgot to remove that line. Thanks for catching this error!

PR #2268, opened 2019-08-08
heatmap by default is False

pyko:
Latest version of `plot_top_losses` has default value of `heatmap` will be `False` (see https://github.com/fastai/fastai/commit/9370be5c20c3e669ab08ce8d29b4aea1352decb1)

Note: prior to the above commit, default is `None` which will means value of `heatmap` will depend on `_test_cnn(self.learn.model)`


sgugger:
Thanks!

PR #2270, opened 2019-08-09
Enable followlinks on os.walk for recursive walk

octavflorescu:
I am loading several ImageDataBunch.from_folder having the same validation set, but experimenting with different training sets.
In order not to have to copy-paste the validation images to each dataset, I have created a symlink (ln -s ../valid ./valid) in each dataset, pointing to the common validation set, but fastai did not use this folder without having set the followlinks param to true.

### DIFF
--- a/fastai/data_block.py
+++ b/fastai/data_block.py
@@ -28,11 +28,11 @@ def _get_files(parent, p, f, extensions):
     return res
 
 def get_files(path:PathOrStr, extensions:Collection[str]=None, recurse:bool=False,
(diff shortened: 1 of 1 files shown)


sgugger:
Seems a little bit dangerous and it makes the tests fail. Even with the test succeeding, I think this should be an additional param that defaults to `False`.

sgugger:
Thanks! Tests were failing due to some unrelated bug with the release of PyTorch 1.2.0.

PR #2275, opened 2019-08-14
ImageSegment save fix

mjhough:
Issue discussed on the Fast AI Forums: https://forums.fast.ai/t/imagesegment-save-method-saves-masks-incorrectly/52459/2

This fix allows the saving of masks properly. Since the current `ImageSegment` `save()` method is inherited from `Image`.


sgugger:
Thanks!

PR #2277, opened 2019-08-14
Fix missing embedding in tensorboard callback

gnoparus:
I've discussed in the forum (https://forums.fast.ai/t/tensorboard-integration/38023/12) 

Embedding is missing from LearnerTensorboardWriter Callback.
Can I add a few lines of code to LearnerTensorboardWriter.on_epoch_end ?

And I've made a test in (https://github.com/gnoparus/fastai_dev/blob/master/dev/42_tensorboard_callback_embedding.ipynb) but because it's tensorboard I don't know how to do assert. 

Before patch you run notebook there is no projector tab in tensorboard, after apply patch you can run notebook to see the projector tab appear.


sgugger:
The test seems tricky to implement in the CI as we don't run anything with notebooks. Thanks for the addition!

PR #2278, opened 2019-08-15
fix for dice

justinblaber:
This should compute dice batch-wise and then return the mean. I believe the previous implementation was computing dice over the entire batch. I found this while trying to use fastai for the pneumothorax kaggle challenge. 

The added test for dice should fail for the previous implementation and work the new one.


sgugger:
Thanks a lot!

PR #2279, opened 2019-08-16
tensorboard callback docs

gnoparus:
After last pull request about fixing bug missing embedding in tensorboard callback.  (https://github.com/fastai/fastai/pull/2277)

I also would like to submit docs about it.


sgugger:
Looking good, thanks for your PR!

PR #2280, opened 2019-08-16
DOC: add from fastai.vision import * to README example

raybellwaves:
As I am first time visitor of https://docs.fast.ai/index.html

I quickly installed the package using the conda example (https://docs.fast.ai/index.html#Installation-and-updating)

I then wanted to try the example at the top. I had to look at https://github.com/fastai/fastai/blob/master/examples/vision.ipynb

to get the line ```from fastai.vision import * ```.

Added this line before the example will help newbies run an example by only looking at the README


sgugger:
Good point, thanks!

Issue #2283, opened 2019-08-18
Error with num_workers in one_batch


juvian:
Steps to reproduce:

-  Make a databunch (with over 16 elements, more noticeable if its on images) with batch size 1 and num_workers = 16. Add a custom_collate to add a print statement there to check amount of calls done to it.
- do data.show_batch(1, figsize=(9, 9), ds_type=DatasetType.Train)

16 print statements will appear. All workers are trying to get a batch, even when wanting only 1 batch. 

Digging through the code, I see that in [one batch](https://github.com/fastai/fastai/blob/bd9bc96e05d464a0a4ed8719947fe941d7875e4c/fastai/basic_data.py#L163) num_workers is set to 0 (`self.num_workers = 0`). The problem is self is the databunch, not the dataloader. The correct code would be:

```
w = dl.num_workers
dl.num_workers = 0
try:     x,y = next(iter(dl))
finally: dl.num_workers = w
```

A way to notice the difference is to do:
```
data.dl(DatasetType.Train).num_workers = 0
data.show_batch(1, figsize=(9, 9), ds_type=DatasetType.Train)
```

Only 1 print will appear as it should. Currently the method takes a lot of time the more workers you have.

Bug seems to have been introduced [here](https://github.com/fastai/fastai/commit/ae3c6942114f0e4ef08d7b338117a9c00c43c294#diff-095b5ac2c0200324dcd06569e1245a4a)

sgugger:
Good catch! Would you mind suggesting a PR to fix this since you already found a solution? Thanks!

sgugger:
Fixed by #2285

PR #2285, opened 2019-08-19
fix num workers to apply to dataloader instead of databunch

juvian:
related to https://github.com/fastai/fastai/issues/2283


sgugger:
Thanks!

PR #2287, opened 2019-08-20
OverSamplingCallback not working on inference time

mmauri:
The OverSamplingCallback initializes the weights inside the constructor using train data.
On inference time it crashes when there is only test data available.

Gist with inference crash before fix;
https://gist.github.com/mmauri/eb538d330f0cd7dad6881c4fde4eef00#file-test_oversamplingcallback_inference-ipynb

Gist after fix:
https://gist.github.com/mmauri/0ae954b0ab09d0314e5042e48bf14280


sgugger:
Seems very reasonable, thanks!

PR #2294, opened 2019-08-22
Add test for Normalize tabular proc

oneraghavan:


sgugger:
Thanks for the tests!

PR #2298, opened 2019-08-26
Patch fixing relative import path

erreur404:


sgugger:
Thanks for fixing those imports. I'm not too sure I understand what the test is testing, could you explain a little bit more?

erreur404:
I had this problem that I wanted to import fastai (1.x) as a "local module" (that is, import what's next to your python script). So that it doesn't get confused with my other fastai folder (0.7), I renamed it "fastai1" and did "import fastai1 as fastai" in my python script. Because of this new name some imports did not work.
In the test I tried to reproduce the circumstances of my problem so it can be checked that the imports work better now :)
Almost every imports used this "import ..basic_train" notation, just a few were actually calling "import fastai.basic_train" so I thought I'd change this and contribute.

sgugger:
The CI installs the latest version of fastai, not 0.7, and not in local, so that test doesn't test anything.

erreur404:
okay, should I remove the test that tests nothing to leave only the relative import ? Or shall we simply close it ? :)

sgugger:
No, fixing the import is good. Just remove the test and I'll merge the PR, thanks!

sgugger:
Thanks!

PR #2299, opened 2019-08-28
Adding EfficientNet to fastai vision

Vanrap:
In this (https://ai.googleblog.com/2019/05/efficientnet-improving-accuracy-and.html) paper published by Google, the authors proposed a new neural network architecture they call “EfficientNet”. EfficientNets are a family of image classification models, which achieve state-of-the-art accuracy, yet being an order-of-magnitude smaller and faster than previous models.

A pytorch implementation of EfficientNet can be found here: https://github.com/lukemelas/EfficientNet-PyTorch. Through this, pytorch implementation, we can easily add EfficientNet to fastai.

From the pytorch implementation of EfficientNet:
“EfficientNet PyTorch is a PyTorch re-implementation of EfficientNet. It is consistent with the original TensorFlow implementation(https://github.com/tensorflow/tpu/tree/master/models/official/efficientnet), such that it is easy to load weights from a TensorFlow checkpoint. At the same time, we aim to make our PyTorch implementation as simple, flexible, and extensible as possible.”

The py file added consists of the pytorch implementation of EfficientNet.
Discussion:https://forums.fast.ai/t/adding-efficientnet-to-fastai-vision/53371

Through this pull request, users can directly use EfficientNet without having to import separate packages or writing additional lines of code.


sgugger:
Thanks for your PR! I'm all for adding new models, but it doesn't feel right to take all the code from  https://github.com/lukemelas/EfficientNet-PyTorch and just put it in fastai. Plus, it would mean we'd need to cherrypick any commit made to fix bugs or make the model better.

The package is pip-installable, so we should do something like the [cadene models](https://github.com/fastai/fastai/blob/master/fastai/vision/models/cadene_models.py) which require a separate package and then tries to import it and raises an exception with a clear error message giving the instructions to install.

Would you mind amending your PR?

Vanrap:
Thanks for replying @sgugger! I totally understand that using code from [this repo ](https://github.com/lukemelas/EfficientNet-PyTorch)would not be practical. Can you please elaborate on what exactly you expect in the amended PR? Do you mean I should try to add EfficientNet to Cadene models?

sgugger:
No, we should import the package for efficient net the same way we do it for the cadene models (which is why I linked that module).

Vanrap:
Hmm. Would it be fine if we use 
`pretrainedmodels = try_import('efficientnet-pytorch')
if not pretrainedmodels:
    raise Exception('Error: `efficientnet-pytorch` is needed. `pip install efficientnet-pytorch`')
from efficientnet_pytorch import EfficientNet `

and then define functions for each EfficientNet model? I believe this would keep us up to date on the pytorch repo.

sgugger:
Yes, that would be fine.

Vanrap:
And, would I have to make changes to any other files?

Vanrap:
I assumed I'd have to make changes to _init_, but it keeps throwing errors.

sgugger:
Thanks for amending. I just used core for getting the `try_import` function instead of rewriting it and removed some vertical space to be consistent with the style of the rest of the library. Will merge when the tests pass.

Vanrap:
Awesome! Thank you very much! So does this mean EfficientNets are officially a part of the fast ai vision models? Can I use them in my code now, or would we have to run more trials/tests?

tmabraham:
@Vanrap if you are using the fastai dev version (installing fastai from github), then you can use it. Otherwise, you will have to wait for the next version of fastai so you can install from PyPI and use it.

PR #2300, opened 2019-08-28
show_image support skimage data images

vainaijr:
show_image fails to show images imported from skimage, so I modified it to show such images also.

### DIFF
--- a/fastai/vision/image.py
+++ b/fastai/vision/image.py
@@ -432,8 +432,8 @@ def show_image(img:Image, ax:plt.Axes=None, figsize:tuple=(3,3), hide_axis:bool=
                 alpha:float=None, **kwargs)->plt.Axes:
     "Display `Image` in notebook."
     if ax is None: fig,ax = plt.subplots(figsize=figsize)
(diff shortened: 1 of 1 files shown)


fastai/vision/image.py
```diff
@@ -432,7 +432,8 @@ def show_image(img:Image, ax:plt.Axes=None, figsize:tuple=(3,3), hide_axis:bool=
                 alpha:float=None, **kwargs)->plt.Axes:
     "Display `Image` in notebook."
     if ax is None: fig,ax = plt.subplots(figsize=figsize)
-    ax.imshow(image2np(img.data), cmap=cmap, alpha=alpha, **kwargs)
+    if ('cpu' not in dir(img)): ax.imshow(img)
```
sgugger: We should also pass `cmap`, `alpha`, and `kwargs` there I think. Also, `hasattr(imh, 'cpu')` is a little bit cleaner than checking `dir`.

sgugger:
Thanks for your PR, I made a suggestion with some changes if you don't mind editing it a bit,

sgugger:
Perfect, thanks!

PR #2303, opened 2019-08-29
Include ability to log more items that were not in training set

iamshwin:
Log more (or fewer) items when warning about labels not existing in the training set, but existing in the validation set.


sgugger:
I don't see why not, thanks for your PR.

PR #2304, opened 2019-08-30
Flat Cosine Annealing 

muellerzr:
This PR has the Flat Cosine Annealing Scheduler and fit function that should be used for the new optimizers such as RAdam and Ranger. PR's from those should be soon. See discussion here for relevant information about the "competition" held: [url](https://forums.fast.ai/t/how-we-beat-the-5-epoch-imagewoof-leaderboard-score-some-new-techniques-to-consider/53453)

Documentation to come soonish. If it is easier to keep it as one PR then it can wait

### DIFF
--- a/fastai/callbacks/flat_cos_anneal.py
+++ b/fastai/callbacks/flat_cos_anneal.py
@@ -9,29 +9,22 @@ __all__ = ['FlatCosAnnealScheduler']
 # The brain child of Mikhail Grankin aimed for use of the new optimizers
 # For more information: https://forums.fast.ai/t/how-we-beat-the-5-epoch-imagewoof-leaderboard-score-some-new-techniques-to-consider/53453
 
-class FlatCosAnnealScheduler(LearnerCallback):
-  """
-  Manage FCFit training as found in the ImageNette experiments
-  https://forums.fast.ai/t/how-we-beat-the-5-epoch-imagewoof-leaderboard-score-some-new-techniques-to-consider/53453
-  Based on idea by Mikhail Grankin
-  """
-  def __init__(self, learn:Learner, lr:float=4e-3, tot_epochs:int=1, moms:Floats=(0.95,0.999),
-             start_pct:float=0.72, curve='cosine'):
-      super().__init__(learn)
-      n = len(learn.data.train_dl)
-      self.tot_epochs = tot_epochs
-      self.anneal_start = int(n * self.tot_epochs * start_pct)
-      self.batch_finish = (n * self.tot_epochs - self.anneal_start)
-      if curve=="cosine":
-          curve_type=annealing_cos
-      elif curve=="linear":
-          curve_type=annealing_linear
-      elif curve=="exponential":
-          curve_type=annealing_exp
-      else:
-          raiseValueError(f"annealing type not supported {curve}")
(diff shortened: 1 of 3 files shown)


fastai/callbacks/flat_cos_anneal.py
```diff
@@ -0,0 +1,64 @@
...
+        self.batch_finish = (n * tot_epochs - self.anneal_start)
+        if curve=="cosine":
+            curve_type=annealing_cos
+        elif curve=="linear":
+            curve_type=annealing_linear
+        elif curve=="exponential":
+            curve_type=annealing_exp
+        else:
+            raiseValueError(f"annealing type not supported {curve}")
+        phase0 = TrainingPhase(self.anneal_start).schedule_hp('lr', lr).schedule_hp('mom', moms[0])
```
sgugger: Just wondering, since you're using the phases API, can't we just create a function that returns a `GeneralScheduler`?

muellerzr: Yes, I will look into it

muellerzr:
I believe I simplified it per your request. The callback should just have an __init__ to set up the scheduler correct?

muellerzr:
Fixes are all done, let me know if that's close to what you are talking about via the API

fastai/callbacks/flat_cos_anneal.py
```diff
@@ -0,0 +1,37 @@
...
+from ..core import *
+from ..callback import *
+from ..basic_train import Learner, LearnerCallback
+
+__all__ = ['FlatCosAnnealScheduler']
+
+# The brain child of Mikhail Grankin aimed for use of the new optimizers
+# For more information: https://forums.fast.ai/t/how-we-beat-the-5-epoch-imagewoof-leaderboard-score-some-new-techniques-to-consider/53453
+
+class FlatCosAnnealScheduler(LearnerCallback):
```
sgugger: I think you should make this a regular function that returns `GeneralScheduler(learn, phases)`. Keep the name capitalized for consistency with the other callbacks.

Second comment: please keep the docstring to one line and add any other info in the doc notebooks when you add the documentation. The link and citation to you are already in the source code in the commented lines 9 and 10, so we don't need it here. You should absolutely put them again in the documentation of this when you make that second PR.

sgugger:
We are on the right track. I think we can simplify it a little bit more.

Also, I have no idea why that is, but the diff on the train.py file shows the entire file as changed, not just the things you added. Do you have any idea why that is? It makes reviewing the PR a bit hard for me.

fastai/callbacks/flat_cos_anneal.py
muellerzr: Adjusted it so.

muellerzr:
I looked into it and tried a few things but I have absolutely no idea why. At first I thought it was because I had moved `fit_fc` to the start, but that did not do anything when I had moved it to the very end. I simplified it a bit more for you, and also got rid of the scheduler function as I don't believe that would've been beneficial here since it's just a function now. But let me know if we should have kept it.

lgrz:
> We are on the right track. I think we can simplify it a little bit more.
> 
> Also, I have no idea why that is, but the diff on the train.py file shows the entire file as changed, not just the things you added. Do you have any idea why that is? It makes reviewing the PR a bit hard for me.

FYI, it was because <LF> was replaced with <CRLF> (windows line endings)

PR #2306, opened 2019-08-30
Adding VGG11 and VGG13 with batch normalization

Vanrap:
I noticed that fast ai currently has VGG16 and VGG19 with batch normalization. Why not complete the family of VGG neural networks with the addition of VGG11 and VGG13?

### DIFF
--- a/fastai/vision/learner.py
+++ b/fastai/vision/learner.py
@@ -37,7 +37,7 @@ model_meta = {
 
     models.densenet121:{**_densenet_meta}, models.densenet169:{**_densenet_meta},
     models.densenet201:{**_densenet_meta}, models.densenet161:{**_densenet_meta},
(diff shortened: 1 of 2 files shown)


sgugger:
Thanks!. You need to add it [here](https://github.com/fastai/fastai/blob/fdc79857745498135aa60b1137c7e137f5d757c0/fastai/vision/learner.py#L40) too if you want users to be able to use it in ` cnn_learner`.

Vanrap:
@sgugger I've made a few changes, please go through them and tell me whether there are any more changes required. 

Also, I have a few questions for you which I want to talk about in private. Is there a way I can message/email you in private? You can also send a mail to pranavpulijala@outlook.com, and I will reply.

sgugger:
Changes are perfect, thanks!
You can DM me on the forum for private questions.

PR #2312, opened 2019-09-02
Fixed import issue

muellerzr:
Just noticed, it should be from ..callbacks to grab the TrainingPhase


muellerzr:
Don't merge this yet though @sgugger I tried running this on another dataset and I'm getting CUDA errors. This needs to be solved before we do anything.

muellerzr:
Looks to be tied to Seb's self attention layer, not this, so we're good.

sgugger:
Thanks!

PR #2313, opened 2019-09-02
added flush and fsync to on_epoch_end()

floleuerer:
CSVLogger writes (closes) the log file at the end of training (on_train_end()). To get frequent updates on the training status, I've added file.flush() and os.fsync() to on_epoch_end() as discussed on the CSVLogger thread  https://forums.fast.ai/t/csv-logger-callback/26971/19 .


sgugger:
Thanks a lot!

PR #2314, opened 2019-09-03
Bug Fix: `plot_top_losses`

muellerzr:
There was a bug happening where since they had the same name, tabular plot_top_losses was taking priority to vision's plot_top_losses. To adjust for this I mimicked how it is for the `plot_multi_top_losses` and just renamed the function to `plot_tab_top_losses`. I had originally thought that this would not have been an issue at the time of making the original PR but since they all have the same super interp class I think the functions were at conflict of which to use sometimes. This resolves the issue by just including a third plot option for tabular.

### DIFF
--- a/fastai/tabular/models.py
+++ b/fastai/tabular/models.py
@@ -48,7 +48,7 @@ def _cl_int_from_learner(cls, learn:Learner, ds_type=DatasetType.Valid, activ:nn
     preds = learn.get_preds(ds_type=ds_type, activ=activ, with_loss=True)
     return cls(learn, *preds, ds_type=ds_type)
 
-def _cl_int_plot_top_losses(self, k, largest:bool=True, return_table:bool=False)->Optional[plt.Figure]:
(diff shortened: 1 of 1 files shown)


sgugger:
The issue is caused by importing fastai.widgets, which imports TabularDatabunch in tabular.data, which runs the code in models and replaces the monkey-patched `plot_top_losses`.

I don't like giving a different name for the same functionality and will solve the issue later this morning with a new `tabular.learner` module where the monkey-patch will happen without being touched by `fastai.widgets`.

sgugger:
Ugh, annoying. Even with the reorganization, the monkey-patch is seen while importing from tabular.data, guess we'll have to go with the different name.

muellerzr:
What about something like the suggestion here? 

https://forums.fast.ai/t/plot-top-losses-throwing-attributeerror-imagelist-object-has-no-attribute-cat-names/51239/8?u=muellerzr

sgugger:
It's probably possible, though I don't have time to implement something like this.

muellerzr:
I'll look into it when I have some time available and see (as I need to learn how to do it first).

PR #2315, opened 2019-09-03
It needed both imports - Sorry

muellerzr:
It needed both `..callbacks` and `..callback` (we grabbed Gen Scheduler and the annealing_cos function) sorry!


sgugger:
Ok, let's hope it works now ;)

PR #2321, opened 2019-09-07
use scikit-image transform image resize as imresize

fivejjs:


fivejjs:
imresize change

sgugger:
Thanks!

PR #2327, opened 2019-09-13
Add fix_ds to fix missing attribute error

csaroff:
Previously, `interp = learn_cln.interpret(DatasetType.Fix)` would throw an attribute error, saying that fix_ds does not exist.  This is important when looking at top_losses or plot_confusion_matrix.
 `learn_cln.interpret(DatasetType.Train)` gives a ClassificationInterpretation which is not meaningful because of the lack of ordering in DatasetType.Train.  This change allows users to meaningfully interpret the training dataset.


sgugger:
Thanks!

PR #2331, opened 2019-09-17
Add example of using batch prediction for text classification

iamshwin:
I hope others find this useful as its a way more efficient way of generating predictions when you have lots of data for inference.

Cheers, Ashwin


sgugger:
Thanks!

PR #2333, opened 2019-09-17
Update CPU-only installation instructions

nihil-admirari:
PyTorch CPU-only installation has changed, greatly simplifying `conda` installs.


sgugger:
Thanks!

PR #2335, opened 2019-09-17
PredictionsCorrector widget

nihil-admirari:
Existing `ImageCleaner` widget makes it possible to relabel images in the dataset. This technique can also be used to manually fix mispredicted images in a test set, e.g. before making a submission in a competition.

Test set usually doesn't have any labels, and due to its size only a subset of it can be looked at. `DatasetFormatter.from_most_unsure` returns the indices of the images whose classes the model is the least sure about, i.e. difference in probabilities between the most probable and second most probable classes is minimal. This method adds to already existing `from_toplosses` and `from_similars`.

`PredictionsCorrector` is a new widget to be used in conjunction with `from_most_unsure`. Since it's very similar to an existing `ImageCleaner`, a base class has been extracted: `BasicImageWidget`.

Example usage on an MNIST dataset: https://www.kaggle.com/niladmirari/digits-recognizer-with-fast-ai.


fastai/widgets/image_cleaner.py
```diff
@@ -94,117 +99,158 @@ def sort_idxs(cls, similarities):
...
-        if self._duplicates:
-            chunked_idxs = chunks(fns_idxs, 2)
-            chunked_idxs = [chunk for chunk in chunked_idxs if Path(items[chunk[0]]).is_file() and Path(items[chunk[1]]).is_file()]
-            return  [(dataset.x[i]._repr_jpeg_(), items[i], self._labels[dataset.y[i].data]) for chunk in chunked_idxs for i in chunk]
-        else:
-            return [(dataset.x[i]._repr_jpeg_(), items[i], self._labels[dataset.y[i].data]) for i in fns_idxs if
-                    Path(items[i]).is_file()]
+        items = self._dataset.x.items
+        idxs = ((i for i in fns_idxs if items[i].is_file())
+                if not drop_batch_on_nonfile
```
nihil-admirari: If `ImageCleaner` is split into two: cleaner and duplicate finder, `create_image_list` in the base class can be simplified by removing `drop_batch_on_nonfile`. Duplicate finder widget will simply override `create_image_list` instead of passing a constructor argument.

sgugger:
Thanks for the PR. You should present your widget on the [forum](https://forums.fast.ai/) so that people know it exists and start using it.

PR #2347, opened 2019-09-24
Minor fix for ImageCleaner with data from CSV (fastai#2344)

stmandl:
This PR contains a simple fix and small test for the problem described in fastai#2344


sgugger:
Thanks!

PR #2351, opened 2019-09-24
Expose tfm_y argument in load_learner

JoshVarty:
As discussed [here](https://forums.fast.ai/t/inference-using-load-learner/38694/28?u=joshvarty) currently there is no way to specify `tfm_y` when loading a learner with `load_learner()`. This can cause an exception when the learner tries to apply transforms against an `EmptyLabel`.

When loading an exported learner, we want callers to be able to choose whether or not to apply transforms against y.


sgugger:
Thanks!

PR #2352, opened 2019-09-26
[WIP] Enable image normalization to be performed on multiple arguments

ntoxeg:
I have basically ported what is done in batch application of the model to normalization - the input is a list of at least one tensor and you map the normalization over them.

Models get invoked with `*x` so this should be transparent to them.


sgugger:
There is a problem for the denomalization then (see failing tests). You can run the tests locally by just typing pytest.

ntoxeg:
I was afraid of that part getting broken, it would be easy to change `denormalize` to work in a similar fashion to `_normalize_batch` (I think I would also move the functionality to `normalize` to maintain greater symmetry between the two).
All of that still leaves that singleton list getting returned from `one_batch` though. I am not sure if it makes sense to add a check for it that flattens that singleton (and possibly in other places in the code) or to give up on this whole thing and keep using my hacked version.

sgugger:
Yeah this is going to be terribly hacky. v2 will handle normalization better for tuple of images, so I suggest closing this for now and wait for the v2 release :)

PR #2353, opened 2019-09-27
Update mnist stats

thomasbrandon:
Updates MNIST stats as discussed on [forum](https://forums.fast.ai/t/mnist-stats-look-wrong/55407/4). I've also changed them to be 1D as otherwise they broadcast 1 channel images up to 3 channel. I put up the [notebook ](https://nbviewer.jupyter.org/gist/thomasbrandon/2e8e365086ed20cef655dac11cd8c365) with the code to calculate which shows the issue with broadcasting.


sgugger:
Thanks!

PR #2355, opened 2019-09-29
Add min_lr parameter to ReduceLROnPlateauCallback

rohitgr7:
Add `min_lr` for lower bound on learning rate in `ReduceLROnPlateauCallback`.

### DIFF
--- a/fastai/callbacks/tracker.py
+++ b/fastai/callbacks/tracker.py
@@ -70,7 +70,7 @@ class EarlyStoppingCallback(TrackerCallback):
             self.best,self.wait = current,0
         else:
             self.wait += 1
-            if self.wait > self.patience:
+            if self.wait >= self.patience:
                 print(f'Epoch {epoch}: early stopping')
                 return {"stop_training":True}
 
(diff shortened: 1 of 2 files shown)


fastai/callbacks/tracker.py
```diff
@@ -70,7 +70,7 @@ def on_epoch_end(self, epoch, **kwargs:Any)->None:
             self.best,self.wait = current,0
         else:
             self.wait += 1
-            if self.wait > self.patience:
+            if self.wait >= self.patience:
```
sgugger: I wouldn't change this behavior as a. it's the one users know, b. it's consistent with other callbacks. We can clarify in the documentation it takes patience full epochs before applying the change at patience+1.

fastai/callbacks/tracker.py
```diff
@@ -124,7 +124,7 @@ def on_epoch_end(self, epoch, **kwargs:Any)->None:
         if self.operator(current - self.min_delta, self.best): self.best,self.wait = current,0
         else:
             self.wait += 1
-            if self.wait > self.patience:
+            if self.wait >= self.patience and self.opt.lr > self.min_lr:
```
sgugger: Same here

sgugger:
Thanks for your PR. I don't think we should change the behavior with patience though.

fastai/callbacks/tracker.py
rohitgr7: Changed as suggested.

sgugger:
Thanks!

Issue #2344, opened 2019-09-22
ImageCleaner bug with data from CSV


stmandl:
When following [lesson2-download](https://github.com/fastai/course-v3/blob/master/nbs/dl1/lesson2-download.ipynb) in the fast.ai course-v3, I ran into a problem, when using the ImageCleaner to remove duplicate images. I get this error message:

```text
AttributeError: 'numpy.str_' object has no attribute 'is_file'
```

I'm using fastai that's built from the github master branch.
This code reproduces the problem:

```python
from fastai.vision import *
from fastai.widgets import *

path = untar_data(URLs.MNIST_TINY)
data_from_csv = ImageList.from_csv(path, csv_name='labels.csv').split_none().label_from_df().transform(get_transforms(), size=224).databunch()
learn_cln = cnn_learner(data_from_csv, models.resnet34, metrics=error_rate)
ds, idxs = DatasetFormatter().from_similars(learn_cln)
ImageCleaner(ds, idxs, path=path, duplicates=True)
```

I also have a trivial fix for the problem, but as I'm a total fast.ai newbie (I started using it today), I wanted to make sure that it is an actual bug in the library code and not an error on my part using the library.

**Installation details**

```text
=== Software === 
python        : 3.7.3
fastai        : 1.0.58.dev0
fastprogress  : 0.1.21
torch         : 1.2.0
nvidia driver : 430.26
torch cuda    : 10.0.130 / is available
torch cudnn   : 7602 / is enabled

=== Hardware === 
nvidia gpus   : 1
torch devices : 1
  - gpu0      : 7979MB | GeForce RTX 2070

=== Environment === 
platform      : Linux-5.0.0-29-generic-x86_64-with-debian-buster-sid
distro        : Ubuntu 18.04 bionic
conda env     : base
python        : /home/stm/anaconda3/bin/python
sys.path      : 
/home/stm/anaconda3/lib/python37.zip
/home/stm/anaconda3/lib/python3.7
/home/stm/anaconda3/lib/python3.7/lib-dynload
/home/stm/anaconda3/lib/python3.7/site-packages
/home/stm/fastai/fastai

Sun Sep 22 22:52:06 2019       
+-----------------------------------------------------------------------------+
| NVIDIA-SMI 430.26       Driver Version: 430.26       CUDA Version: 10.2     |
|-------------------------------+----------------------+----------------------+
| GPU  Name        Persistence-M| Bus-Id        Disp.A | Volatile Uncorr. ECC |
[...]

PR #2358, opened 2019-10-01
Fix AttributeError in ImageCleaner (duplicates=False case)

nihil-admirari:
When refactoring `ImageCleaner` (https://github.com/fastai/fastai/pull/2335), I removed `Path(items[i])` conversions, since in cases I checked, `items[i]` was already a `Path`, and conversion seemed superfluous. Turns out it's not: https://github.com/fastai/fastai/issues/2344.

https://github.com/fastai/fastai/pull/2347 fixed duplicates finder, this PR fixes plain `ImageCleaner`.

### DIFF
--- a/fastai/test_registry.json
+++ b/fastai/test_registry.json
@@ -1898,14 +1898,14 @@
         },
         {
             "file": "tests/test_widgets_image_cleaner.py",
-            "line": 36,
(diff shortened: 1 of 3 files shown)


sgugger:
Thanks!

PR #2360, opened 2019-10-04
Adding collab_learner and from_df to this_tests

sinemetu1:
I noticed these weren't showing up in the test doc links [here](https://docs.fast.ai/collab.html#CollabLearner). So this is just supposed to expose that link.


sgugger:
Thanks!

Issue #2359, opened 2019-10-04
PyTest Warning: Importing ABCs from collections deprecated


ZaydH:
**Describe the bug**

Running `pytest` on a module that imports `DataBunch` yields the warning:

```
<my_python>/site-packages/fastai/imports/core.py:9
  <my_python>/site-packages/fastai/imports/core.py:9: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
    from collections import abc,  Counter, defaultdict, Iterable, namedtuple, OrderedDict

-- Docs: https://docs.pytest.org/en/latest/warnings.html
```

**Provide your installation details**
```text
=== Software ===
python       : 3.7.1
fastai       : 1.0.58
fastprogress : 0.1.21
torch        : 1.2.0
torch cuda   : None / is **Not available**

=== Hardware ===
No GPUs available

=== Environment ===
platform     : Darwin-18.7.0-x86_64-i386-64bit
conda env    : Unknown
python       : /Users/zayd/.pyenv/versions/3.7.1/Python.framework/Versions/3.7/bin/python
sys.path     : /Users/zayd/repos/fastai_pytest_testcase
/Users/zayd/.pyenv/versions/3.7.1/Python.framework/Versions/3.7/lib/python37.zip
/Users/zayd/.pyenv/versions/3.7.1/Python.framework/Versions/3.7/lib/python3.7
/Users/zayd/.pyenv/versions/3.7.1/Python.framework/Versions/3.7/lib/python3.7/lib-dynload
/Users/zayd/Library/Python/3.7/lib/python/site-packages
/Users/zayd/.pyenv/versions/3.7.1/Python.framework/Versions/3.7/lib/python3.7/site-packages
no supported gpus found on this system
```

`tox`: 3.14.0

**To Reproduce**

I created a [test case github repo](https://github.com/ZaydH/fastai_pytest_testcase).  Running the `tox` command in that directory should yield the warning.

**Expected behavior**

No warning in PyTest.

**Additional context**

Other large packages have ran into this issue.  If this is acknowledged as an issue, I can work to get you a PR quickly to address it.

sgugger:
I'm guessing the fix is just to import the right abc. Is it compatible with python 3.6? That's the minimal version we require.

ZaydH:
[...]

PR #2363, opened 2019-10-06
Fix for Issue 2359: Update core Imports to Prevent pytest Warning

ZaydH:
Updated core imports so Iterable is imported from collections.abc. Removed superfluous abc import


ZaydH:
Cross reference with #2359.

sgugger:
Tests are passing so it doesn't seem like it destroys anything :)
Thanks for the fix!

PR #2374, opened 2019-10-14
Add space between commas

raybellwaves:
I've added spaces between the commas in ``predict``. I feel in helps me with the legibility.

### DIFF
--- a/fastai/basic_train.py
+++ b/fastai/basic_train.py
@@ -367,11 +367,11 @@ class Learner():
         "Return predicted class, label and probabilities for `item`."
         batch = self.data.one_item(item)
         res = self.pred_batch(batch=batch, with_dropout=with_dropout)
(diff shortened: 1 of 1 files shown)


sgugger:
I understand why you may like it more this way but in general, please refrain from PRs just changing the style of the code like this (see the [style note](https://docs.fast.ai/dev/style.html) for more thoughts of Jeremy about that).

While I'd agree with some of this spacing (deconstruting a tuple never has a space in fastai: `a,b = c,d`, the other changes are fine by me) the choice is ultimately on the person who originally wrote the code and we would like to spend time coding and not discussing minor stile issues (see the first note in the contributing file "Contributing to Parkinson’s law of triviality has negative consequences for a project. Let’s focus on deep learning!")

raybellwaves:
No problem. Thanks for the link.

PR #2378, opened 2019-10-14
Fix OversamplingCallback

thomasbrandon:
Fix for OversamplingCallback in PyTorch 1.3 as per[ forum discussion](https://forums.fast.ai/t/pytorch-1-3-breaks-oversamplingcallback/56488).


sgugger:
Thank you!

PR #2380, opened 2019-10-16
Update show_top_losses function

DarshanPatel11:
Update show_top_losses function using learn object as the old version using interp was giving wrong output.

 
 
### DIFF
--- a/fastai/text/interpret.py
+++ b/fastai/text/interpret.py
@@ -75,26 +75,28 @@ class TextClassificationInterpretation(ClassificationInterpretation):
         text, attn = self.intrinsic_attention(text, class_id)
         show_piece_attn(text.text.split(), to_np(attn), **kwargs)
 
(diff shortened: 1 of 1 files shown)


sgugger:
I don't understand your PR: what is this `learn` you use everywhere? It's not an argument of the function...

sgugger:
That function doesn't know the learner object, so this can't work. The function will only give the correct output if the `Learner` used to create it is called `learn`, you can't rely on that kind of behavior.

I don't understand the bug you are trying to fix so maybe start with a small reproducible example so we can address it properly.

sgugger:
Ok, should be fixed by [this commit](https://github.com/fastai/fastai/commit/531e0f0ddeaba8dff93e79c3a9e9deb807755c17). The problem was that there was another `TextClassificationInterpretation` defined in models.awd_lstm and this one didn't have the `ordered=True` in the call to get_preds.

PR #2384, opened 2019-10-18
Add activ to pred_batch

rohitgr7:
Added `activ` parameter in case we want a custom activation instead of the [predefined one from the loss function.](https://github.com/fastai/fastai/blob/85dca5af46e62c48f9574d79d3de46b409707453/fastai/basic_train.py#L350).


sgugger:
This was an oversight on our side, thanks for fixing!

PR #2389, opened 2019-10-25
added apply_tfms to EmptyLabel

tcapelle:
This PR aims to fix applying transforms to `EmptyLabelLists`.
```
from fastai import *
from fastai.datasets import *
from fastai.vision import *

path = untar_data(URLs.MNIST)

bs=64

src = (ImageImageList.from_folder(path)
       .split_by_folder('training', 'testing')
       .label_from_func(lambda x: x)
       .add_test_folder(path)
       .transform(get_transforms(), tfm_y=True)
       .databunch(bs=bs).normalize(imagenet_stats, do_y=True)
      )
```
this code fails because it tries to apply transforms to the labels that are `EmptyLabelLists`. My solution is probably not the best way to solve this. @sgugger  probably a more elegant solution would be to modify `transform` method from `LabelList` or the `_check_kwargs`.

### DIFF
--- a/fastai/core.py
+++ b/fastai/core.py
@@ -303,6 +303,7 @@ class EmptyLabel(ItemBase):
     def __init__(self): self.obj,self.data = 0,0
     def __str__(self):  return ''
     def __hash__(self): return hash(str(self))
+    def apply_tfms(self, tfms:Collection, **kwargs): return self
 
 class Category(ItemBase):
     "Basic class for single classification labels."


sgugger:
This step fails on purpose as it is a sign something is wrong: trying to apply transforms to an input when they should also be applied to the target, which means they should be reversed on the predicted output to get proper predictions.
I don't think letting them pass is a good idea, at least not without a proper warning.

tcapelle:
It's very usual that the test set does not have labels, and we want to produce those labels. So you get a `LabelList` with `EmptyLabel`, how to solve this?
Maybe there is another way of doing this that I am not seeing?

sgugger:
You can't apply transforms at inference without post-processing your results. I don't know what your transforms are but if you crop your inputs for instance, then the predictions you get are only for that cropped part and not the image in its totality.
The error is there to make you realize something is wrong (though the meassage could be clearer). You have to disable those transforms by creating your test set passing `tfms=None, tfm_y=False`.

tcapelle:
My hope was to have a default behavior when passing a `test_set` without labels, overriding the `tfm_y=True`.

sgugger:
You can even add the test set after creating your DataBunch (it has a test method), but yes, in this case you need to add your test set *after* the transform step.

sgugger:
I've made the error message clearer in [this commit](https://github.com/fastai/fastai/commit/f89e59bc2f9baa1551f252b3f84fb5ecea71a20c), so closing this PR.

PR #2392, opened 2019-10-30
Fixes pillow version mismatch

lgvaz:
Directly following the install instructions on README causes the error:
```
from fastai.vision import *
ImportError: The _imaging extension was built for another version of Pillow or PIL:
Core version: 6.0.0.post0
Pillow version: 6.2.0
```

In the proposed changes I just specify pillow version to be 6.0.0

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ Note that JPEG decoding can be a bottleneck, particularly if you have a fast GPU
 
 ```bash
 conda uninstall --force jpeg libtiff -y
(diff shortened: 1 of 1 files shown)


sgugger:
Yes, it looks like pillow-simd didn't catch up with the latest pillow release. Can you check from time to time what version pillow simd is? I'll do the same and when they release their 6.2 version, we can remove this.

lgvaz:
Sure, if I catch the update before you I'll post it here =)

PR #2393, opened 2019-10-31
Use barrier synchronization before torch.load() in distributed training mode. 

philtrade:
Use barrier synchronization to avoid processes loading model or data file before it was written completely by the master process, when in distributed training mode.

**History**
First noticed spurious load() error when converting the lesson3-camvid Jupyter notebook to run in distributed mode in one go.  Then found discussion on SaveModelCallback error here: https://forums.fast.ai/t/to-distributed-with-savemodelcallback/41384

Experimented a fix and discussed it with @sgugger:

 https://forums.fast.ai/t/race-conditions-in-savemodelcallback-when-in-distributed-training-mode/57504

The gist to simulate the error and test the fix: https://gist.github.com/philtrade/e0ecbfd33eed7dad80642318e806a90f

**Fix**
Introduce ```fastai.torch_core.distrib_barrier()``` convenience routine.  Insert it in front of torch.load() in ```basic_data.load_data(), basic_train.load(), and text.learner.load_encoder()```.

In ```callbacks.tracker.SaveModelCallback.on_train_end()```, the ```is_file()``` check may not work: when the master process is lagging behind other processes (speed skew) noticeably, the check risks skipping over the entire load().  With barrier synchronization in place, this check can be omitted.

### DIFF
--- a/fastai/basic_data.py
+++ b/fastai/basic_data.py
@@ -277,6 +277,7 @@ def load_data(path:PathOrStr, file:PathLikeOrBinaryStream='data_save.pkl', bs:in
               no_check:bool=False, **kwargs)->DataBunch:
     "Load a saved `DataBunch` from `path/file`. `file` can be file-like (file or buffer)"
     source = Path(path)/file if is_pathlike(file) else file
+    distrib_barrier()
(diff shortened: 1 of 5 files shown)


sgugger:
Looks good, thanks a lot!

philtrade:
You're welcome, and thank you for reviewing the change!

Phil

PR #2400, opened 2019-11-07
add `DataBunch.slice(...)` for quick playing with massive datasets

koyo922:

### DIFF
--- a/fastai/basic_data.py
+++ b/fastai/basic_data.py
@@ -272,6 +272,19 @@ class DataBunch():
             warn(message)
             print(final_message)
 
+    def slice(self, slice_train:slice, slice_valid:slice=slice(None), slice_test:slice=slice(None)):
+        "Return sliced version of `self`, for quick experimenting with massive datasets"
+        dls = []
+        for ds, slc in zip((self.train_ds, self.valid_ds, self.test_ds), (slice_train, slice_valid, slice_test)):
+            if ds is None: dls.append(None)
+            else: dls.append(DataLoader(ds[slc], batch_size=self.batch_size))
+        return type(self)(*dls[:2], test_dl=dls[2],
+                          device=self.device, dl_tfms=self.dl_tfms, collate_fn=self.collate_fn, path=self.path)
+
+    def __getitem__(self, item):
+        assert isinstance(item, slice)
+        return self.slice(item, item, item)
+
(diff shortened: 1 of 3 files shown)


fastai/basic_data.py
```diff
@@ -272,6 +272,19 @@ def sanity_check(self):
             warn(message)
             print(final_message)
 
+    def slice(self, slice_train:slice, slice_valid:slice=slice(None), slice_test:slice=slice(None)):
+        "Return sliced version of `self`, for quick experimenting with massive datasets"
+        dls = []
+        for ds, slc in zip((self.train_ds, self.valid_ds, self.test_ds), (slice_train, slice_valid, slice_test)):
+            if ds is None: dls.append(None)
+            else: dls.append(DataLoader(ds[slc], batch_size=self.batch_size))
```
sgugger: There are plenty of other arguments to pass to `DataLoader`, and it could also be a subclass of a `DataLoader` here.

tests/test_basic_data.py
```diff
@@ -117,3 +117,19 @@ def __len__(self):
...
+
+def test_DataBunch_slice():
+    x_train,y_train = fake_basedata(n_in=3, batch_size=6),fake_basedata(n_in=3, batch_size=6)
+    x_valid,y_valid = fake_basedata(n_in=3, batch_size=3),fake_basedata(n_in=3, batch_size=3)
+    train_ds,valid_ds = TensorDataset(x_train, y_train),TensorDataset(x_valid, y_valid)
+    data = DataBunch.create(train_ds, valid_ds, bs=5)
+    this_tests(data.slice)
+
+    small_data = data.slice(slice(None,2),slice(None,1)) # explicit slice
+    assert 2 == len(small_data.train_ds[0]) # len of `x` part
```
Zaxtim565: 2

fastai/basic_data.py
koyo922: Thank you for the reply, the `use_partial_data` is what I wanted.

I've watch the full courses at https://course.fast.ai/ but found no relating contents.
please publish a reading list, so that new users can get familiar enough to known such features.

sgugger: This function is an external contribution, which was added after the course material was prepared, that's why it's not mentioned there. The [docs](https://docs.fast.ai/data_block.html#ItemList.use_partial_data) are the best place to get information though it's true this feature could be better advertised.

Closing the PR if you're satisfied with `use_partial_data` for now. If you want to update it with the changes I mentioned, I'd be happy to merge it.

PR #2401, opened 2019-11-09
Fix typo in docs `StopAfterNBatches`

lgrz:
The callback does not take `Learner` as an argument to the constructor.


sgugger:
Thanks!

PR #2411, opened 2019-11-16
calculating memory metrics for cpu usage

lscarpato:
The CpuPeakMemMetric Callabck is useful for when we are using fast.ai in cpu, without gpu. In these cases can occur memory craches and is necessary to control the cpu memory usage

There is already the PeakMemMetric calback, but is only for gpu usage so I created a very similar callback for cpu usage.

The code is really similar to PeakMemMetric.

Thank you.

### DIFF
--- a/fastai/callbacks/cpu_mem.py
+++ b/fastai/callbacks/cpu_mem.py
@@ -10,9 +10,6 @@ class CpuPeakMemMetric(LearnerCallback):
 
     _order = -20  # Needs to run before the recorder
 
-    def __init__(self, learn: Learner):
-        super().__init__(learn)
-
     def peak_monitor_start(self):
         self.peak_monitoring = True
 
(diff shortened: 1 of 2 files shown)


fastai/callbacks/cpu_mem.py
```diff
@@ -0,0 +1,49 @@
...
+from ..utils.mem import *
+from ..basic_train import *
+from ..torch_core import *
+
+class CpuPeakMemMetric(LearnerCallback):
+    "Callback that measures used and peaked general and CPU memory."
+
+    _order = -20  # Needs to run before the recorder
+
+    def __init__(self, learn: Learner):
```
sgugger: There is no need for an init if it only calls the super class.

fastai/callbacks/cpu_mem.py
```diff
@@ -0,0 +1,49 @@
...
+        self.cpu_mem_used_peak = -1
+
+        while True:
+            if not self.peak_monitoring: break
+            time.sleep(0.001)  # 1msec
+
+    def on_train_begin(self, **kwargs):
+        self.learn.recorder.add_metric_names(['cpu used', 'cpu_peak'])
+
+    def on_epoch_begin(self, **kwargs):
```
sgugger: Can all be on the same line since it's a very short function.

fastai/callbacks/cpu_mem.py
luscarpato: Hi @sgugger, I already fixed it. Thank you for your attention.

fastai/callbacks/cpu_mem.py
luscarpato: Hi @sgugger, I already fixed it. 

Thank you.

PR #2412, opened 2019-11-18
Fix type hints in text.learner, text.models 

tpietruszka:
 
Tinkering in text.learner, once again I had trouble figuring out (/remembering) parameter and return types in a few places. Found some outdated/erroneous type hints (+ my linter found some more) on the way.

This PR should not really break things, as it only changes type hints.

I also felt like a comment was due in `MultiBatchEncoder.concat` - I know `fast.ai ` is against comments, but I really do not think that one-liner operating on a list of lists of 3d tensors is very clear by itself. 

Maybe the comment would be better in the docstring? But then again I do not see multi-line docstings anywhere.

### DIFF
--- a/fastai/callbacks/rnn.py
+++ b/fastai/callbacks/rnn.py
@@ -16,7 +16,7 @@ class RNNTrainer(LearnerCallback):
         "Reset the hidden state of the model."
         self.learn.model.reset()
 
-    def on_loss_begin(self, last_output:Tuple[Tensor,Tensor,Tensor], **kwargs):
+    def on_loss_begin(self, last_output:Tuple[Tensor,Sequence[Tensor],Sequence[Tensor]], **kwargs):
         "Save the extra outputs for later and only returns the true output."
         self.raw_out,self.out = last_output[1],last_output[2]
         return {'last_output': last_output[0]}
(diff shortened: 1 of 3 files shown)


sgugger:
Thanks for your PR! All the type annotations are good. For the concat method, I would rather not add those comments. For one, the method is private-ish so the end user doesn't need to know how it works and then I personally find the source code more helpful than those comments. In any case, any bit of extra documentation should go in the corresponding doc notebook.

tpietruszka:
OK, removed the comments. They were not the most elegant, sure.

Although I must add that having worked with `fastai.text` a few times at different points in time, every time I come back to it I find it a bit confusing. Several lists of (lists of) tensors get passed around, in ways and for reasons that are not very clear. 

I do get it, a lot of it is for `RNNTrainer`'s benefit, but neither the reason not the exact way seems very apparent from the code. I think for people that are not using a `learner` as a black box, but also haven't spent much time developing the library, some amount of in-code explanations would be useful.  Unfortunately, I do not know how to do it exactly within `fast.ai`'s coding style.

Sorry for the ranty message, maybe you (@sgugger) have some idea how to make this code more approachable? (it's internal structure, the interfaces seem completely fine...)

tpietruszka:
... and type hints fixed in one more place.

sgugger:
Thanks!
I think all the inner details can be added in the doc notebook, especially as it can be done in some interacting way. And the behavior of this concat function can be further explained there.

PR #2413, opened 2019-11-18
Extend EarlyStoppingCallback to support multiple metrics

dreamflasher:
Currently EarlyStoppingCallback only supports tracking one metric. Using multiple EarlyStoppingCallback will result in early stopping if any of the metrics did not improve. What has been missing (and is solved with this PR) is an "union" operator on early stopping. With this code change early stopping will occur when neither of the supplied metrics improved for patience epochs.

### DIFF
--- a/fastai/callbacks/tracker.py
+++ b/fastai/callbacks/tracker.py
@@ -50,29 +50,57 @@ class TrackerCallback(LearnerCallback):
             warn(f'{self.__class__} conditioned on metric `{self.monitor}` which is not available. Available metrics are: {", ".join(map(str, self.learn.recorder.names[1:-1]))}')
         return values.get(self.monitor)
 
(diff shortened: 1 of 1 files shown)


sgugger:
Thanks for your PR.
It's a lot of changes for this simple callback, so I think it should have its own separate class. I'm closing this for now as we try to stabilize the core of v1 before the v2 release, and especially since it would make a lot of sense to have this be part of a fastai-contrib repo we will build soon for v2. In the meantime, do not hesitate to make a gist out of your code and share it on the forum so that people can use it as it seems like a useful feature. Thanks again!

dreamflasher:
@sgugger Thank you, I understand. Yet, have a look at the code again: Github makes it look like a lot changed, but it's just changing single values to lists and iterating over them.

PR #2417, opened 2019-11-19
Update the default "ordered" for "get_preds"

jpzhangvincent:
It makes more sense to set the "ordered" default "True" to get the results from the validation/test set.

The issue was discussed in
https://forums.fast.ai/t/does-get-preds-randomize-shuffle-the-order-of-the-predictions/44329/2


sgugger:
I don't see why not, thanks for the PR!

PR #2421, opened 2019-11-26
fix roc_curve cuda vs cpu error

wassname:
Minor change to roc_curve. 

Previously it was always using the default cpu device type. This causes an error when trying to add a cuda and cpu array.


sgugger:
There is a `new_zeros` method for tensor that would this more nicely. I think `zer = fps.new_zeros(1)` would work and be nicer, if you don't mind changing your code.

wassname:
Good idea, I added it the the PR

sgugger:
Thanks!

PR #2423, opened 2019-11-26
Fix: default argument syntax used instead of type hint

hlibbabii:
This would prevent an unclear error that would happen if `last_metrics` were not passed to `on_epoch_end`.


sgugger:
Oh good catch, thanks!

PR #2424, opened 2019-11-26
fix transform typo

colinquirk:
Just fixing a small typo I came across. Hopefully I've done everything correctly but let me know if that's not the case.

Thanks for a great package/class!


sgugger:
Thanks! Everything was correct indeed :)

PR #2427, opened 2019-12-04
Tweak `_maybe_squeeze` helper to behave properly with single-row dataframes too

mstewart141:
Hello all,

When calling `learner.get_preds` on a `tabular.Learner` there is currently a failure case for single-row dataframes (the `basic_data.DataBunch` for said learner is a single-row DF).

This PR makes a minor tweak to the one-line utility fn `_maybe_squeeze`. After this tweak, the previous functionality is preserved while single-row DFs in the data bunch are also supported.

The edge-case boils down to zero-dim array semantics in numpy:
```
np.array(0).shape
Out[19]: ()
np.array(0).ravel().shape
Out[20]: (1,)
np.squeeze(np.array(0)).shape
Out[21]: ()
```

Note that ravel properly translates shape `() -> (1,)` while squeeze does not. 

Thank you for the great library. Feedback appreciated; I'm happy to add a test case too as desired.

### DIFF
--- a/fastai/data_block.py
+++ b/fastai/data_block.py
@@ -10,7 +10,10 @@ __all__ = ['ItemList', 'CategoryList', 'MultiCategoryList', 'MultiCategoryProces
 def _decode(df):
     return np.array([[df.columns[i] for i,t in enumerate(x) if t==1] for x in df.values], dtype=np.object)
 
-def _maybe_squeeze(arr): return (arr if is1d(arr) else np.array(arr).ravel())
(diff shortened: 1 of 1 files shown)


sgugger:
Thanks for your PR! I think we can tweak it to make everything fit in one line using something like
```
return a if cond1 else b if cond2 else c
```
Would you mind adjusting your code?

mstewart141:
No problem, I have done so.

sgugger:
Thanks!

PR #2429, opened 2019-12-07
Fix incorrect param type

mess-lelouch:

While developing stuff I found out the type of a param was incorrect. My grain of sand :D


sgugger:
Thanks!

PR #2431, opened 2019-12-10
IMAGE{NETTE|WOOF}2{|_160|_320} URLs added

kdorichev:
URLs for the updated `Imagenette` and `Imagewoof` datasets, as mentioned [here](https://github.com/fastai/imagenette/blob/master/README.md).


sgugger:
Thanks! Let me just check with Jeremy if we want to erase the old urls or have two different sets.

sgugger:
So after talking with Jeremy we'd like to replace the old urls, not add new names. Can you amend your PR to do that? Thanks!

kdorichev:
> So after talking with Jeremy we'd like to replace the old urls, not add new names. Can you amend your PR to do that? Thanks!

Done in this [PR](https://github.com/fastai/fastai/pull/2432).

PR #2437, opened 2019-12-18
Matching darknet architecture from Yolov3 paper

jasondemeter:
My reasoning for this change is [on the fastai forum.](https://forums.fast.ai/t/project-implement-yolo-v3-backbone-and-preact-resnet/14613/66?u=jdemeter)

 
 

sgugger:
Thanks! I can't remember why we set it this way but it seems safer to use the official implementation indeed.

PR #2440, opened 2019-12-28
Remove auto_update kwarg from progress_bar call

dimpavloff:
The kwarg was removed in the November release of progressbar ( https://github.com/fastai/fastprogress/commit/c7c91a1b410a2381a8901e40021aa538ee2a8ec2 ). 
This then breaks for fresh `fastai` installs (e.g. https://forums.fast.ai/t/fastprogress-auto-update/58830/6).

Also call `pbar.update(0)` because the `progress_bar.update()` requires it atm -- it assumes `progress_bar.wait_for` exists.
It would probably be better if the `progress_bar.__init__` set this to 1 or for `update` to handle this gracefully but I'm new to the libraries and don't have the time for a larger PR :)

 
 
### DIFF
--- a/fastai/core.py
+++ b/fastai/core.py
@@ -208,8 +208,9 @@ def download_url(url:str, dest:str, overwrite:bool=False, pbar:ProgressBar=None,
 
     with open(dest, 'wb') as f:
         nbytes = 0
-        if show_progress: pbar = progress_bar(range(file_size), auto_update=False, leave=False, parent=pbar)
+        if show_progress: pbar = progress_bar(range(file_size), leave=False, parent=pbar)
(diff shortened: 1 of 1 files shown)


sgugger:
Thanks for the PR! Failure in the tests is independent so I will merge. 
It's normal you have to manually call update(0), as it signifies you are starting the iteration (which is different than the init) when not directly iterating through the progressbar.

dimpavloff:
> It's normal you have to manually call update(0), as it signifies you are starting the iteration (which is different than the init) when not directly iterating through the progressbar.
Thanks for clarifying and the quick merge

Issue #2448, opened 2020-01-06
Bug report, in lesson4-mnist_sgd.ipynb


charlie4fun:
**Describe the bug**
Hello, I am doing [this course](http://course18.fast.ai/ml) from fast AI and in [this notebook](https://github.com/fastai/fastai/blob/master/courses/ml1/lesson4-mnist_sgd.ipynb) while fitting the model I get the error, which could be find below on screenshot. I will push my fix now also, but it is a bit ugly, though I would be glad to improve it if you would find it reasonable.

**Screenshots**
![Screenshot from 2020-01-06 20-42-27](https://user-images.githubusercontent.com/19307345/71843966-633f3a00-30c5-11ea-8ee4-953e46fb6f8e.png)

charlie4fun:
I guess error appeared because numpy got updated and what used to work with old numpy does not function now. I'm using numpy==1.17.4

sgugger:
Thanks for flagging. I went for another fix (similar to yours but putting in the one place it comes from instead of two different places).

PR #2449, opened 2020-01-06
issue #2448 bug fix

charlie4fun:
Pull request is fixing [this issue](https://github.com/fastai/fastai/issues/2448), maybe it is not the best way to fix it, but the error occures in 0.7 version of library which is not used so much now, as I understand, so maybe it would be fine as it is. Though I can correct it if you would give me a hint on how to do it better


sgugger:
Went for an alternative fix, so closing this.

PR #2450, opened 2020-01-08
Typo in documentation for Learner.save() argument: name -> file

kennysong:
"If argument `name` is a pathlib object that's an absolute path" 
to
"If argument `file` is a pathlib object that's an absolute path"
 

sgugger:
Thanks!

PR #2452, opened 2020-01-09
made centering optionnal in pca

nestordemeure:
Added a parameter to the pca method to make centering optional (useful when the position of the origin is meaningful).


sgugger:
Thanks!

PR #2457, opened 2020-01-16
Handling breaking change in pytorch grid_sample

akarazniewicz:
Disclaimer: 
I cannot create a proper test case for this fix; this, unfortunately depends on environment and pytorch version.

Backgrund:
Pytorch in 1.3.0 started to change a default behavior grid_sample from being True to False. In version 1.3.0 and 1.3.1 pytorch basically warned that this behavior will change in 1.4.0. Now, recently released Pytorch 1.4.0 indeed introduced this breaking change in grid_sample (https://pytorch.org/docs/stable/nn.functional.html?highlight=grid_sample#torch.nn.functional.grid_sample). Now, fastai,with pytorch >1.4.0 will simply have improper behavior everywhere where grid_sample is used. 

This simple PR keeps correct behavior, regardless of pytorch version.


sgugger:
Thanks!

PR #2458, opened 2020-01-17
Include Optuna integration badge on README.md

Crissman:
Hi!

Optuna is a new hyperparameter optimization library, and we’ve written an Optuna integration module for fast.ai to make it easy to search for good hyperparameter settings and prune unpromising trials. We’re looking for ways to let fast.ai users know this is available and thought a badge to show Optuna integration is available would be helpful.

Here’s our example of using the pruning integration with fast.ai: https://github.com/optuna/optuna/blob/master/examples/pruning/keras_integration.py

If there are other ways you would recommend reaching out to fast.ai users to let them know about Optuna, please let us know.

Thanks!


sgugger:
Hi there, thanks for you PR. The link is for a Keras example however, not fastai so I'm not really sure there is support. You can reach to fastai users on the [forum](https://forums.fast.ai/), they'll read that more than the REAMDE in any case.

PR #2461, opened 2020-01-20
Update datasets.py

Anku5hk:


Anku5hk:
Imagenette dataset names have some conflict.

Anku5hk:
with untar_data(URLs.IMAGENETTE_320) it gets an error

sgugger:
No, those are the rights URLs, you would change to the old versions of the datasets with this. Please open an issue with your error so we can fix this.

PR #2463, opened 2020-01-22
allow the test set to be shuffled

dreamflasher:


sgugger:
This is dangerous as it makes your predictions be in a different order. You can always do it with
```
dbunch.test_dl = dbunch.test_dl.new(shuffle=True)
```
but I don't think exposing the argument is a good idea.

PR #2464, opened 2020-01-22
Fix SP vocab and model names

manycoding:
SPProcessor was saving files under the wrong filenames, so I fixed it


sgugger:
I don't understand what this PR changes, could you elaborate? Thanks!

manycoding:
@sgugger 

As of now, SP saves 2 files - `path/'"cache_dir".vocab' and path/'"cache_dir".model'` and returns `path/tmp/spm` path. When the latter is used to read vocab again, it won't be found.

So I fixed the paths, making it consistent with `load`

sgugger:
Got it, it's the thing returned that is different. Thanks!

PR #2465, opened 2020-01-22
fixes exception before model save

dreamflasher:
If an exception happens before the model is saved, then this raised an FileNotFoundError which is overriding the previous error, see e.g. https://forums.fast.ai/t/to-distributed-with-savemodelcallback/41384/46


sgugger:
Thanks!

PR #2466, opened 2020-01-23
Scale default value is too low, 0.08 instead of 0.8 in _presize funct…

nofreewill42:
…ion.

Having a little object in the photo, the object often wasn't even present in the transformed data with the default 0.08 value.
0.8 gives reasonable augmentation, I think that the original intent also was 0.8, but an extra zero was typed in mistakenly or something.

 
 

sgugger:
No, this is the default from [torchvision](https://pytorch.org/docs/stable/torchvision/transforms.html#torchvision.transforms.RandomResizedCrop), there is no error there. I agree this is low, a value we often use is 0.35 but I think the default should be the same as the underlying transform.

nofreewill42:
Thank you for letting me know about this.

PR #2469, opened 2020-01-24
improved monitoring of metrics

maxmatical:
Currently, `SaveModelCallback` only saves model when the monitored value increases, which leads to the learner saving the worst model when monitoring things like loss or mse. Expanded the monitor option to `'increase'` and `'decrease'` so that the callback will only save the mode based on increase/decrease in the monitored metric


sgugger:
Thanks for your PR, but there is a mode that determines if improvement should be understood as greater or less. That mode is automatically set to less if there is `loss` or `error` in the name of the metric, which is all done in the base class `TrackerCallback`. 
You can set that mode to `min` or `max` if the automatic pick does not work for your problem.

PR #2472, opened 2020-01-27
Worked on issue 2471: Updated tutorial.data.html file

arp95:
Updated the accuracy number in tutorial.data.html from 74% to 91% as mentioned in the issue.


sgugger:
The docs file are auto-generated from the notebooks, so you should put that change in [here](https://github.com/fastai/fastai/blob/master/docs_src/tutorial.data.ipynb) and it will propagate to the docs once we rebuild them (otherwise your change will be erased next time we build the documentation). Closing this PR as it's easier if you do it in a new one.

Issue #2474, opened 2020-01-27
Adding signature to datapath4file method 


arp95:
As seen in the screenshot:
![fastai2](https://user-images.githubusercontent.com/8268392/73220469-79707100-412c-11ea-9960-efa13d6b78ba.png)

The signature of filename is "filename:str" which I understood after going through the url2path and url2name functions. Should I change the signature and add "filename:str" in the code also?
The reason is while debugging I had to see the fastai documentation and understand the type of filename before I could debug the entire untar_data function.

sgugger:
Yes, that would be good.

sgugger:
Closed by #2475

PR #2475, opened 2020-01-28
Worked on issue #2474

arp95:
Modified datasets.py file. Added ":str" statement to datapath4file and modelpath4file functions.


sgugger:
Thanks!

PR #2478, opened 2020-01-29
examples/tabular: Fixed 'split_idx' not defined

drscotthawley:
As per forum post https://forums.fast.ai/t/examples-tabular-ipynb-split-idx-is-not-defined/62459/4, reverting back by un-commenting necessary code.

 
 
 - [x ] Read the [contributing docs](https://github.com/fastai/fastai/blob/master/CONTRIBUTING.md)
 - [x ] Add details about your PR.

The variable 'split_idx' was being used without being defined.  
I posted to the forum first, where Zachary Mueller suggested the fix, and pointed toward a similar code in course-v3 where it's corrected: https://github.com/fastai/course-v3/blob/master/nbs/dl1/lesson4-tabular.ipynb

So this submission is simply reverting to where the `list(range())` syntax is uncommented and replacing the `split_idx`.

### DIFF
--- a/examples/tabular.ipynb
+++ b/examples/tabular.ipynb
@@ -9,7 +9,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": null,
    "metadata": {},
    "outputs": [],
    "source": [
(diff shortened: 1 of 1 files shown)


drscotthawley:
Ok...nbstripout failed.  Will re-run and re-submit in a bit...

sgugger:
Thanks!

Issue #2481, opened 2020-01-31
Minor documentation update (performance.md)


celynwalters:
This relates to [docs/performance.md](https://github.com/fastai/fastai/blob/ed3635d17a52af7ead2805b802f20b99fa7a7b28/docs/performance.md) (and I hope this affects the [documentation webpage](https://docs.fast.ai/performance.html))

Since (I believe) version 7.0.0, Pillow has a different command.

```bash
python -c "from PIL import Image; print(Image.PILLOW_VERSION)"
3.2.0.post3
```

is now:

```bash
python -c "from PIL import Image; print(Image.__version__)"
7.0.0.post3
```

The rest of the `Pillow-SIMD` instructions seem to be correct.

sgugger:
The thing is since they did not offer backward compatibility, it's actually one of those two commands... I'll write an util function to grab the version when I have a bit of time.

PR #2482, opened 2020-02-01
Update performance.md

krishnakatyal:
version 7.0.0, Pillow has a different command.

python -c "from PIL import Image; print(Image.PILLOW_VERSION)"
3.2.0.post3
is now:

python -c "from PIL import Image; print(Image.__version__)"
7.0.0.post3

 
 
### DIFF
--- a/docs/performance.md
+++ b/docs/performance.md
@@ -134,8 +134,8 @@ Here are the detailed instructions, with an optional `TIFF` support:
 #### How to check whether you're running `Pillow` or `Pillow-SIMD`?
 
 ```
-python -c "from PIL import Image; print(Image.PILLOW_VERSION)"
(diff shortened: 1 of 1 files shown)


krishnakatyal:
@jph00  KINDLY REVIEW MY PR

stas00:
Thank you.

Perhaps leave the old one as a note, for those who still run older versions? As the newer version you suggested will fail for the older one.

sgugger:
Thanks for your PR. As I commented on #2481, we need a util function that works for both cases, this will fail for any user with an earlier version of Pillow, so this is not the solution.

stas00:
FYI, this has been fixed in master (docs too).

PR #2483, opened 2020-02-01
Fix SaveModelCallback

manycoding:
I added `cpu()` casting. 
When `get_monitor_value` is called in SaveModelCallback it fails since the value it returns for valid loss still sits on gpu.

I checked [valid_loss](https://github.com/fastai/fastai/blob/ed3635d17a52af7ead2805b802f20b99fa7a7b28/fastai/callbacks/tracker.py#L44) in the parent TrackerCallback but it was on gpu. Perhaps it's changed somewhere else.

**Update:**
Above was true for `language_model_learner.fit_one_cycle` after `unfreeze`. However, `text_classifier_learner` in the same situation already has `valid_loss` on cpu. So I added type check.

 
 
### DIFF
--- a/fastai/callbacks/tracker.py
+++ b/fastai/callbacks/tracker.py
@@ -93,9 +93,7 @@ class SaveModelCallback(TrackerCallback):
         "Compare the value monitored to its best score and maybe save the model."
         if self.every=="epoch": self.learn.save(f'{self.name}_{epoch}')
         else: #every="improvement"
-            current = self.get_monitor_value()
-            if isinstance(current, torch.Tensor):
-                current = current.cpu()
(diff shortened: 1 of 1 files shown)


sgugger:
Thanks for your PR! Since the call to `.cpu()` already does that test inside, I simplified a little bit the code you added.

manycoding:
@sgugger The problem is that `self.get_monitor_value()` returns a numpy array during lm learner training, so it will fail since numpy doesn't have `cpu()`

That's why I added the check.

sgugger:
Ah, my mistake. Will push a fix in a few minutes and revert to your code.

Issue #2485, opened 2020-02-04
Make lr_find() usable in a distributed data parallel mode, a  step towards enabling notebook to run distributed training.


philtrade:
In the process of building a tool to facilitate distributed training + `fastai v1` in Jupyter notebook, I realized `lr_find()` doesn't work in distributed mode, and it doesn't need a distributed learner to work either.

Although I got it to work by dynamically intercepting`Learner.lr_find()`, I have distilled that monkey wrench/patch down to a backward compatible, fairly simple change.  I'd like to submit it here for your consideration.

The net outcome is that: one can use a Learner which had been `to_distributed()`,  and run `lr_find()` on it without any complication.  It is a beneficial feature to fastai's Jupyter notebook users, IMHO.

If it pleases you, I'll be happy to submit a PR, and help porting it over to `fastai v2` as well, so that `fastai v2` lesson notebooks will be *distributed-training ready* early on, a feature that was asked and answered "no" many times for `fastai v1`.

The change takes 2 steps:

1. Only allows the master proc (rank 0) to run `lr_find()`, like how `basic_train.save()` does it, other GPUs aren't helpful anyway to this training run, are they?

2. Temporarily take the `learner` out of the distributed mode, do the usual `lr_find()` dance, and restore it back to the distributed condition afterwards.

Below is the diff, I have tested it *with and without distributed training, with one GPU or multiple GPUs DDP*.  Comments are welcome!  I'll be formalize it into a PR if it's acceptable to you. 

* ["The diff for *Temporarily suspend distributed mode when doing lr_find*"](https://github.com/fastai/fastai/compare/master...philtrade:master?diff=split)

Thank you.

Note: [1]. the reason DDP doesn't play well for `lr_find()` is that, the multiple GPUs each trains on its own and may early-stop at different batches, causing spurious deadlocks, as the earlier quitters and late comers wait for each other at different rendezvous points.  Hence the above solution.
[2]  The tool I built, [The iPython extension `Ddip`](https://github.com/philtrade/Ddip), which I intend to add support to `fastai v2` soon, and would love to contribute it to  the evolving `fastai` code base.

sgugger:
Running lr find in process 0 only will change the effective batch size compared to the distributed training. I think the check should be to let the training run its full course on all processes for this, to avoid the dead-lock, then just showing the loss of process 0 up to the point this process would have stopped.

philtrade:
[...]

PR #2487, opened 2020-02-08
Make lr_find() DDP-safe

philtrade:
See Issue #2485 - `lr_find()` may deadlock in distributed mode due to early stopping in the middle of the epoch.

`LRFinder.on_batch_end(), on_epoch_end()`: 
Don't use `{ 'stop_epoch' : True }`. Mark the best loss batch number, let the epoch finish.  Returns `{'stop_training' : True } in `on_epoch_end()`, when all the GPUs would be in-sync, having consumed all the batches in this epoch.

`train.lr_find()`:
In distributed mode, multiply the number of epochs by the # of GPUs to give `lr_find()` a long enough runway to search for loss expansion.  Also suggest any `skip_end` to use for `plot()`.

Tests:

**1. Deadlocked on first run**:  `bs=2`, one GPU finished at 00:27, timestamp garbled with the progress bar output, while cell status is `[ * ]` -- still busy.  Only possible to capture such screen when cell is stuck.
bs=2 DDP deadlock never reached end of on_train_end()

**2.  Passed 100 runs without deadlock:** `bs=1`**, at which previously would deadlock easily.
Screen Shot 2020-02-07 at 3 59 47 PM


fastai/callbacks/lr_finder.py
```diff
@@ -24,11 +25,16 @@ def on_train_begin(self, pbar, **kwargs:Any)->None:
 
     def on_batch_end(self, iteration:int, smooth_loss:TensorOrNumber, **kwargs:Any)->None:
         "Determine if loss has runaway and we should stop."
+        self.batch_num += 1
```
sgugger: Why are you not using `iteration`? It is the same thing (starts at 0 at the beginning of training and is incremented at each batch).

fastai/train.py
```diff
@@ -37,7 +37,7 @@ def lr_find(learn:Learner, start_lr:Floats=1e-7, end_lr:Floats=10, num_it:int=10
     end_lr = learn.lr_range(end_lr)
     end_lr = np.array(end_lr) if is_listy(end_lr) else end_lr
     cb = LRFinder(learn, start_lr, end_lr, num_it, stop_div)
-    epochs = int(np.ceil(num_it/len(learn.data.train_dl)))
+    epochs = int(np.ceil(num_it/len(learn.data.train_dl))) * (num_distrib() or 1)
```
sgugger: Why do you need to multiply by `num_distrib` here?

fastai/callbacks/lr_finder.py
philtrade: Hello Sylvain, ha, glad to learn something new --- I wasn't aware what  `iteration` does.  Sure enough that's the piece to use.  Thank you!!  Upon digging I also learned about `epoch` and `num_batch` in the implicit parameters passed from `state_dict`.

I have updated the PR to use these three existing entries to do what I intended.

philtrade:
Strange, the tests failed with "HTTP Error 403: Forbidden", which seems unrelated to my change....

https://dev.azure.com/fastdotai/fastai/_build/results?buildId=6859&view=logs&j=d1113ce6-f179-58af-fca7-fd734ada1ab9&t=c6941af0-1990-549e-1833-f8e1570af56d&l=547

But I'm able to download the model https://download.pytorch.org/models/resnet34-333f7ec4.pth from a browser..... could it be a transient error?

fastai/train.py
philtrade: Hi Sylvain, thank you for the question.  This line compensates for inadequate number of epochs when attempting to `lr_find()` in distributed training mode.

As `LRFinder` steps through a range of learning rate only inside `on_batch_end()`, the more the # of batches it processes, the more learning rates it tries.  If the **total # of batches processed** is not large enough, `lr_find()` may finish without reaching a learning rate at which the loss started to explode.

In distributed training with multiple GPUs: the # of batches per epoch is scaled down, but the above line 40 **does not** scale up the **# of epochs** enough.    The change I propose compensate for that, by scaling up **# of epochs by # of GPUs**, so that the **total # of batches** will be big enough for `lr_find()`.  Early-stopping still works with this change.

Does this make sense?

------- evidence from experiments below -------
See the three graphs below for comparison of a successful `lr_find` in single GPU mode, and an unsuccessful one in distributed training, using the same `bs`, then lastly with the change, a successful `lr_find()` that has up to 150 batches to find the exploding-loss learning rate.

E.g. **lesson3-camvid** in course v3, dl1, single GPU (GTX 1080ti), and `bs=8`.

1. Graph 1 --- the number of batches per epoch is determined  to be **75**, and it will find the loss-explosion after at around batch # ~68 - 72-ish, the **range of learning rate** covered is `1e-06` -- `1e-02`:

2. Graph 2 ---   Scaling to use 3 GPUs in distributed training, `lr_find()` would run each epoch with only **75/3 = 25** batches, but the original code only sets "epochs" to 2.  As such, `LRFinder.on_batch_end()`only gets to process **25 x 2 = 50** batches, and the **range of learning rates** searched ended before`1e-03`, much shorter than case 1 above, without finding the learning rate at which loss exploded.

3. Graph 3 --- Then with this PR, in distributed training, `epochs * (num_distrib() or 1)` changes it from 2 to 6, and that gives a max # of batches of `6 * 25 = 150`.  See the range of learning rate covered is good again.

**Graph 1: single GPU, bs=8, max # of batches to try = 75, lr covered: 1e-06 -- 1e-02**
bs=8, non-DDP

**Graph 2: three GPUs, bs=8, max # of batches to try = 25 x 2 = 50 only, , lr covered: 1e-06 -- 7e-03***
bs=8 DDP bn=25, epoch = 2


**Graph 3: three GPUs, bs=8, max # of batches to try = 25 x 6 = 150, lr covered: 1e-06 -- 1e-02**
Screen Shot 2020-02-12 at 2 58 34 AM

sgugger: Got it, thanks for explaining it to me!

sgugger:
Thanks, this looks good to me!

Issue #2490, opened 2020-02-13
download_images does not filter empty strings


ozgurshn:
**Describe the bug**
download_images method does not filter the empty strings after reading the CSV file and splitting into lines array. I don't know whether it should be solved when creating CSV but I fixed it by filtering the string array in my fork https://github.com/ozgurshn/fastai/commits/master. Should I send a PR for this fix?

it originates from this line 
https://github.com/fastai/fastai/blob/ac46636d156639716b2f88feb21c31aac8be6695/fastai/vision/data.py#L194

Changing this to filtering as below will fix this issue.
urls = list(filter(None, open(urls).read().strip().split("\n")))[:max_pics]

**Provide your installation details**
```text
=== Software === 
python       : 3.6.9
fastai       : 1.0.60
fastprogress : 0.2.2
torch        : 1.4.0
torch cuda   : 10.1 / is **Not available** 

=== Hardware === 
No GPUs available 

=== Environment === 
platform     : Linux-4.14.137+-x86_64-with-Ubuntu-18.04-bionic
distro       : #1 SMP Thu Aug 8 02:47:02 PDT 2019
conda env    : Unknown
python       : /usr/bin/python3
sys.path     : 
/env/python
/usr/lib/python36.zip
/usr/lib/python3.6
/usr/lib/python3.6/lib-dynload
/usr/local/lib/python3.6/dist-packages
/usr/lib/python3/dist-packages
/usr/local/lib/python3.6/dist-packages/IPython/extensions
/root/.ipython
no supported gpus found on this system
```

**To Reproduce**
1) Create CSV file of Google images that contain at leasn one empty link ('').
2) use download_images for your CSV 
`download_images('download (8).csv',dest='images2',max_workers=0)
`
**Expected behavior**
Downloading without showing any error

**Screenshots**
<img width="736" alt="Screen Shot 2020-02-13 at 21 30 19" src="https://user-images.githubusercontent.com/2404074/74466330-05fa8d80-4ea8-11ea-9931-b79ee4747f79.png">

sgugger:
One could argue that you need to clean up the csv file as a separate steps (it won't work if you don't put links but random text, obviously), but I don't the see the harm in filtering for empty strings. Don't hesitate to suggest a PR.

ozgurshn:
Just suggested a [PR](https://github.com/fastai/fastai/pull/2492)

sgugger:
Fixed by #2492

PR #2492, opened 2020-02-14
Filtering empty strings in download_images function

ozgurshn:
Using the script below in Chome Dev Tools to download images from Google Images may include empty strings that resulted in errors for the download_images method. It might be solved in this script but I wanted to filter these empty strings in download_images method. This PR is fixing the [issue](https://github.com/fastai/fastai/issues/2490).

```
urls=Array.from(document.querySelectorAll('.rg_i')).map(el=> el.hasAttribute('data-src')?el.getAttribute('data-src'):el.getAttribute('data-iurl'));
window.open('data:text/csv;charset=utf-8,' + escape(urls.join('\n')));

```
You can see the error [here](https://forums.fast.ai/t/problems-fetching-urls-from-google-images/62933/12?u=ozgur) 

It seems download_images method reads CSV file, splits it with “\n” to read line by line, but does not filter empty strings in the string array.

`urls = open(urls).read().strip().split("\n")[:max_pics]
`
Changing this line as below will fix this issue.
`urls = list(filter(None, open(urls).read().strip().split("\n")))[:max_pics]
`

### DIFF
--- a/fastai/vision/data.py
+++ b/fastai/vision/data.py
@@ -191,7 +191,7 @@ def _download_image_inner(dest, url, i, timeout=4):
 
 def download_images(urls:Union[Path, str], dest:PathOrStr, max_pics:int=1000, max_workers:int=8, timeout=4):
     "Download images listed in text file `urls` to path `dest`, at most `max_pics`"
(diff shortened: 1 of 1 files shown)


sgugger:
Thanks!

Issue #2496, opened 2020-02-17
DistributedDataParallel trainer interferes with GANTrainer.switch()


philtrade:
**Describe the bug**

Training wgan in Distributed Data Parallel mode would crash, because it cannot find the `switch()` method in the `DistributedDataParallel` object.

**Provide your installation details**
```
fastai v1, python 3.7, PyTorch v1.2 and v1.4, 3 cuda GPUs (gtx 1080 ti)
platform      : Linux-4.15.0-62-generic-x86_64-with-Ubuntu-18.04-bionic
```

**To Reproduce**
Run the scripted version of `lesson7-wgan` notebook in DDP mode, even with only 1 GPU:  https://gist.github.com/philtrade/cb9953079a99afc522fa777574f2549d
by `python -m torch.distributed.launch --nproc_per_node=1 wgan_ddp.py`
will produce:
```text
  File "/home/ndim1/Dropbox/fastai/fastai/basic_train.py", line 98, in fit
    cb_handler.on_epoch_begin()
  File "/home/ndim1/Dropbox/fastai/fastai/callback.py", line 273, in on_epoch_begin
    self('epoch_begin')
  File "/home/ndim1/Dropbox/fastai/fastai/callback.py", line 251, in __call__
    for cb in self.callbacks: self._call_and_update(cb, cb_name, **kwargs)
  File "/home/ndim1/Dropbox/fastai/fastai/callback.py", line 241, in _call_and_update
    new = ifnone(getattr(cb, f'on_{cb_name}')(**self.state_dict, **kwargs), dict())
  File "/home/ndim1/Dropbox/fastai/fastai/vision/gan.py", line 134, in on_epoch_begin
    self.switch(self.gen_mode)
  File "/home/ndim1/Dropbox/fastai/fastai/vision/gan.py", line 154, in switch
    self.model.switch(gen_mode)
  File "/usr/local/lib/python3.7/dist-packages/torch/nn/modules/module.py", line 576, in __getattr__
    type(self).__name__, name))
AttributeError: 'DistributedDataParallel' object has no attribute 'switch'
```

**Expected behavior**
Training should proceed, e.g.:
```
epoch     train_loss  valid_loss  gen_loss  disc_loss  time    
Epoch 1/3 : |███---------------------------------------------------------------| 5.92% [10/169 00:26<07:00 -0.1393
```

**Triage**
[...]

PR #2498, opened 2020-02-18
GANTrainer: Create reference to model.switch() in constructor  (fix for issue #2496)

philtrade:
Create a reference to `self.model.switch()` for later use in `GANTrainer.switch()`, as `DistributedDataParallel()` interferes with `self.model`.  See issue #2496 .


sgugger:
Thanks a lot for the fix!

philtrade:
You are so welcome.  I must thank you and Jeremy for the open platform and
opportunity to learn and help.

PR #2503, opened 2020-02-19
Add train_cats() which existed in v0.7

EzzEddin:
Hi, while following the notebook of the first lesson on [fast.ai course](https://course18.fast.ai/lessonsml1/lesson1.html) I found *train_cats* function does not exist in the latest version, I tried to install v0.7 but failed so alternatively I moved *train_cats* from v0.7 to be in the latest version. So now when you use this: `from fastai.tabular import *` and then use `train_cats(<data_frame>)` it should work.

### DIFF
--- a/fastai/tabular/transform.py
+++ b/fastai/tabular/transform.py
@@ -1,6 +1,6 @@
 "Cleaning and feature engineering functions for structured data"
 from ..torch_core import *
-from pandas.api.types import is_numeric_dtype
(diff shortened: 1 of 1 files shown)


sgugger:
The old course is not compatible with v1, as is written at the top of every notebook. That's why the function `train_cat` is not in v1, there is another approach used that you can discover on the [new course](https://course.fast.ai/).

EzzEddin:
Okay, is there an alternative function for `proc_df` in the new course as well?

PR #2507, opened 2020-02-29
[Bug] Update train_sentencepiece to accept paths with spaces

angertdevsingh:
`train_sentencepiece`  uses f-strings to pass paths to `SentencePieceTrainer`.

If a path has a space in the name, it should be in quotes.


sgugger:
Indeed, thanks!

PR #2513, opened 2020-03-07
Fix Imputer Import

aminvakil:

from sklearn.impute import SimpleImputer will work because of the following

DeprecationWarning: Class Imputer is deprecated; Imputer was deprecated in version 0.20 and will be removed in 0.22. Import impute.SimpleImputer from sklearn instead.

https://github.com/mindsdb/lightwood/issues/75#issuecomment-565657292


sgugger:
Thanks for your PR! There is no plan to keep the old course up to date with the new versions of the libraries (your code will fail with all people having an old version btw).
We can constrain the version of sklearn to a maximum, but not make changes like this.

PR #2516, opened 2020-03-09
edited the computation of ConfusionMatrix for large speedup

arjunbgupta:
### Describe the change
The method `on_batch_end()` of class `ConfusionMatrix` is very slow when the data has a large number of target classes.

The line that causes the computation to be slow is the following :
```cm = ((preds==self.x[:, None]) & (targs==self.x[:, None, None])).sum(dim=2, dtype=torch.float32)```
This is because we initialize an MxM matrix at each call (where M is the number of labels). We propose to initialize this matrix only once at the start of the Confusion Matrix computation and to leverage the indexing capabilities offered by PyTorch (seen in the line `self.cm[targs, preds] += 1`)

### Context
I stumbled over this issue when when doing a single label text classification task on WordNet definitions with over 10,000 target classes. The training was working well but the validation was tremendously slow. I opened a thread on the fastai forums describing my problem here :
[https://forums.fast.ai/t/learner-uses-gpu-with-training-set-but-not-validation/64698/6](url)

After doing some memory profiling, I pinpointed that the source of the problem came from the `on_batch_end()` method.

I was using an OpenStack VM of 16GB of RAM and 4 cores when this issue arose. A colleague and I reproduced the problem and proposed the above change in a Google Colab environment to make sure that the issue wasn't hardware related.

### Testing and Reproducing
In an interactive python environment (we used Google Colab for this), initialize a sample output:
```
in_batch = torch.rand((12, 10000)).to(torch.device('cpu'))
b = torch.tensor([254]).repeat(12)
b[5] = 128

in_batch[5, 128] = np.inf
```
`in_batch` represents a model output (e.g. softmax). `b`  is a vector of predictions that we consider to be the ground truth. We set the 5th element of B to 128 and the 128th element of the 5th batch of the output to `np.inf` so that this predictions matches it's ground truth.
```
def on_batch_end_original(last_output, last_target, cm=None, n_classes=0, **kwargs):
        preds = last_output.argmax(-1).view(-1).cpu()
        targs = last_target.cpu()
        if n_classes == 0:
            n_classes = last_output.shape[-1]
            x = torch.arange(0, n_classes)
        cm_tmp = ((preds==x[:, None]) & (targs==x[:, None, None])).sum(dim=2, dtype=torch.float32)
        if cm is None: cm = cm_tmp
        else:               cm += cm
        return cm

  def on_batch_end(last_output, last_target, cm=None, n_classes=0, **kwargs):
        preds = last_output.argmax(-1).view(-1).cpu()
        targs = last_target.cpu()
        if n_classes == 0:
            n_classes = last_output.shape[-1]
            #x = torch.arange(0, n_classes)
        if cm is None: cm = torch.zeros((n_classes, n_classes))
        #cm = ((preds==self.x[:, None]) & (targs==self.x[:, None, None])).sum(dim=2, dtype=torch.float32)
        cm[targs, preds] += 1
        return cm
```
The function `on_batch_end_original()` mimics the original method found in the `Confusionmatrix` class. `on_batch_end()` mimics the behavior of our change.
We time each functions with:
```
%%timeit
cm = None
for i in range(1):
  cm = on_batch_end_original(in_batch, b, cm)
```
`[output] 1 loop, best of 3: 5.46 s per loop`

```
%%timeit
cm2 = None
for i in range(100):
  cm2 = on_batch_end(in_batch, b, cm2)
```
`[output] 10 loops, best of 3: 102 ms per loop`

As we can see, our propose change is able to compute a 100 times the confusion matrix  in a fraction of the time it takes the original function to compute it once.
We make sure our confusion matrices are equal:
```(cm == cm2).all()```
`[output] tensor(True)`

Finally,
```cm.diag().argmax()```
`[output] tensor(128)`
verifies that the confusion matrices recorded a correct prediction for label 128.

#### Note:
It is possible that this change is slower on tasks with only few target classes but it seems to us that the large speedup benefits of this change outweigh a small change of compute time. This might need further testing if our change does not seem convincing to you.

**Co-authored-by: Martin Sotir <martinsotir@gmail.com>**

### DIFF
--- a/fastai/metrics.py
+++ b/fastai/metrics.py
@@ -155,10 +155,8 @@ class ConfusionMatrix(Callback):
         targs = last_target.cpu()
         if self.n_classes == 0:
             self.n_classes = last_output.shape[-1]
-            self.x = torch.arange(0, self.n_classes)
-        cm = ((preds==self.x[:, None]) & (targs==self.x[:, None, None])).sum(dim=2, dtype=torch.float32)
-        if self.cm is None: self.cm =  cm
-        else:               self.cm += cm
+        if self.cm is None: self.cm = torch.zeros((n_classes, n_classes))
+        self.cm[targs, preds] += 1
 
     def on_epoch_end(self, **kwargs):
         self.metric = self.cm
--- a/fastai/metrics.py
+++ b/fastai/metrics.py
@@ -155,8 +155,10 @@ class ConfusionMatrix(Callback):
         targs = last_target.cpu()
         if self.n_classes == 0:
             self.n_classes = last_output.shape[-1]
-        if self.cm is None: self.cm = torch.zeros((n_classes, n_classes))
-        self.cm[targs, preds] += 1
+        if self.cm is None: self.cm = torch.zeros((n_classes, n_classes), device=torch.device('cpu'), dtype=torch.int32)
+        cm_temp_numpy = self.cm.numpy()
+        np.add.at(cm_temp_numpy, (targs ,preds), 1)
+        self.cm = torch.from_numpy(cm_temp_numpy)
 
     def on_epoch_end(self, **kwargs):
         self.metric = self.cm
--- a/fastai/metrics.py
+++ b/fastai/metrics.py
@@ -155,7 +155,7 @@ class ConfusionMatrix(Callback):
         targs = last_target.cpu()
         if self.n_classes == 0:
             self.n_classes = last_output.shape[-1]
-        if self.cm is None: self.cm = torch.zeros((n_classes, n_classes), device=torch.device('cpu'), dtype=torch.int32)
+        if self.cm is None: self.cm = torch.zeros((n_classes, n_classes), device=torch.device('cpu'))
         cm_temp_numpy = self.cm.numpy()
         np.add.at(cm_temp_numpy, (targs ,preds), 1)
         self.cm = torch.from_numpy(cm_temp_numpy)

### CODE REVIEW

**Comment by sgugger** [comment]
I understand the problem and the slowdown, but your suggested change does not produce the same result:
```
targ = tensor([0,2,1,3,0])
pred = tensor([0,1,1,3,0])
cm[targ, pred] += 1
```
gives a +1 on cm[0,0], but it should give a +2, given the fact there are 2 targets and predictions 0.

Would it be acceptable to compute the confusion matrix at the end of the phase instead of at the end of each batch?

**Comment by Shadeswalker** [comment]
> I understand the problem and the slowdown, but your suggested change does not produce the same result:
> 
> ```
> targ = tensor([0,2,1,3,0])
> pred = tensor([0,1,1,3,0])
> cm[targ, pred] += 1
> ```
> 
> gives a +1 on cm[0,0], but it should give a +2, given the fact there are 2 targets and predictions 0.

Ah yes, my bad!

> Would it be acceptable to compute the confusion matrix at the end of the phase instead of at the end of each batch?

If the initialization of the CM is not done at each batch I don't see any issues with that.

**Comment by martinsotir** [comment]
Hi @sgugger , I have been working with @Shadeswalker on this issue.

We encounter this slowdown using Recall and Precision metrics with a Fastai learner (each metric computation will hang the CPU for ~4s for each batch when N_labels=10k). To my knowledge this issue has been fixed in fastaiV2 by using scikit-learn metrics. For this PR we were looking for the least invasive patch.

> Would it be acceptable to compute the confusion matrix at the end of the phase instead of at the end of each batch?

Seems a good solution to us, though it may be disruptive depending on how this metric used by fastai users ? (I am relatively new to the fastai library :) ). This would also require storing predictions and targets for all the validation datasets and it may increase memory usage when N_val_samples >> N_labels^2.

> gives a +1 on cm[0,0], but it should give a +2, given the fact there are 2 targets and predictions 0.

Good catch. Instead of indexing `cm` directly we could use a loop, or use `np.add.at(cm, (targ, pred), 1)` if we use a numpy array instead of the CPU torch.tensor for `cm` (`np.add.at` is [not available for pythorch when indices are multidimensional](https://pytorch.org/docs/stable/tensors.html#torch.Tensor.index_add_)).

**Comment by sgugger** [comment]
The loop would probably be very slow. If numpy has a function that does the trick, maybe the conversion to numpy would be the least disruptive (we don't really want to make substantial changes to v1 right now, since v2 is our main focus).

**Comment by martinsotir** [comment]
A quick bench:

```python
n_classes = 20 # 2000
batch_size = 256

targs = torch.randint(low=0, high=n_classes, size=(batch_size,))
preds = torch.randint(low=0, high=n_classes, size=(batch_size,))
```
### Option 0: With tensor broadcasting (orignal source)

```python
cm = torch.zeros((n_classes, n_classes), device=torch.device('cpu'), dtype=torch.int32)
x = torch.arange(0, n_classes)
cm = ((preds==x[:, None]) & (targs==x[:, None, None])).sum(dim=2, dtype=torch.float32)
```

### Option 1:  with loop indexing

```python
%%timeit
cm = torch.zeros((n_classes, n_classes), device=torch.device('cpu'), dtype=torch.int32)
for i in range(batch_size):
  cm[targs[i], preds[i]] += 1
```

### Option 2.a:  With np.add.at + numpy array

```python
%%timeit
cm = np.zeros((n_classes, n_classes), dtype=np.int32)
np.add.at(cm, (targs ,preds), 1)
```

### Option 2.b With np.add.at + tensor -> numpy -> tensor conversion

```python
%%timeit
cm = torch.zeros((n_classes, n_classes), device=torch.device('cpu'), dtype=torch.int32)
cm_temp_numpy = cm.numpy()
np.add.at(cm_temp_numpy, (targs ,preds), 1)
cm = torch.from_numpy(cm_temp_numpy)
````

### Results

|            | `n_classes= 20` | `n_classes= 2000` |
|------------|-------------------|---------------------|
| Option 0   | 480 µs            | 3.65 s              |
| Option 1   | 6.26 ms           | 18.2 ms             |
| Option 2.a | 86.9 µs           | 12.1 ms             |
| Option 2.b | 95.5 µs           | 11.8 ms             |

`np.add.at` is fast in both situations. If we want `cm` to remain a torch tensor (keeping the API unchanged), the torch-> numpy -> torch conversion seems to be quick enough (no copy in most cases I presume ?).

sgugger:
I think cm can become a numpy array, with the conversion to torch tensor done only once at `on_epoch_end`. Then, all subclasses would just need to call `super().on_epoch_end()` at the start of their `super().on_epoch_end`.

arjunbgupta:
I implemented option 2.b. of @martinsotir's benchmark. I didn't feel familiar enough with the code and coding guidelines to change `cm` to a numpy array and then change all the subclasses. If it is the solution that you prefer, feel free to change it and close this request. You could also disregard any of these changes but I think some people could benefit from them.

sgugger:
Ok with this. I will revisit and implement my way to avoid the multiple conversions if I have time someday.
Thanks a lot for investigating and fixing this bug!

PR #2517, opened 2020-03-10
Update awd_lstm.py

garethmd:
When the weights in an embedding layer are initialised all of the weights are updated, including those set to zero at the padding index. This means that the weights at the padding idx are now non zero, which I don't think can be desirable. The change I am proposing just reinstates the zeros at the padding index.

 
 
### DIFF
--- a/fastai/text/models/awd_lstm.py
+++ b/fastai/text/models/awd_lstm.py
@@ -97,6 +97,8 @@ class AWD_LSTM(Module):
             self.rnns = [WeightDropout(rnn, weight_p) for rnn in self.rnns]
         self.rnns = nn.ModuleList(self.rnns)
         self.encoder.weight.data.uniform_(-self.initrange, self.initrange)
+        if self.encoder.padding_idx is not None:
+                self.encoder.weight.data[self.encoder.padding_idx] = 0.
         self.input_dp = RNNDropout(input_p)
         self.hidden_dps = nn.ModuleList([RNNDropout(hidden_p) for l in range(n_layers)])
 


sgugger:
Note sure this is desirable: the network can learn something about the padding index that could be helpful (otherwise it would leave it to zero). I'd leave this as is, unless there is compelling evidence it changes anything.

garethmd:
Hi @sgugger thanks for the reply. Maybe I am missing something, but I can't think of a scenario where the network should learn padding parameters. In fact I thought the entire point of them was to allow the network to handle tensors of variable lengths, with the padding used to fill up the space of the longest tensor. As such the padded space caries no information relating to the input and its gradients should be zero, which cannot be the case unless their values are zero ( or if Pytorch sets the gradients to be zero for the padding index internally). If they are non zero then any subsequent affine layer gets affected and I can't see how that could possibly be correct?

sgugger:
> As such the padded space caries no information relating to the input and its gradients should be zero, which cannot be the case unless their values are zero

Well if that is the case, then the weights will never be changed, so the PR has no use there either. But if the model does encounter a padding token for which the outputs are not discarded, then it will compute a gradient that is meant to make the loss lower. So in that case, I don't see why updating those weights would be hurtful, the model will just learn a semantic representation of the padding token, the same way it learns a representation of the BOS token.

garethmd:
Hi @sgugger 

> Well if that is the case, then the weights will never be changed, so the PR has no use there either.

This is the issue that the PR fixes. When a Pytorch embedding is created with a padding idx then the weights at the padding index are set to zero. If you subsequently initialise the embedding then those zeros in the padding index are overwritten with something that is non zero. The PR just sets those weights back to zero. I've attached a screenshot of some code in case what I am saying is not clear.

Screenshot 2020-03-10 at 15 35 20

The BOS token I can understand. Then network needs to learn its representation and every record the data set will have one, but I think this is different for padding. The number of padding tokens used is dependent on the length of the sequence relative to the longest sequence in the batch.

at inference why should this
`<BOS><THE><FILM><WAS><GREAT><PAD><PAD><PAD>`

be different from this?
`<BOS><THE><FILM><WAS><GREAT>`

If those padding tokens are non-zero I don't see how they can be equivalent.

sgugger:
Ah my bad, I though you were doing this at each forward pass cause I dind't read properly. We can certainly add this. Sorry for the misunderstanding!

PR #2525, opened 2020-03-24
Fix #2524

### LINKED ISSUE #2524: GANTrainer doesn't denormalize the image shown at the end of each epoch
**Describe the bug**

The images shown by GANTrainer at the end of each epoch are darker than the images actually generated. 

**Provide your installation details**

```
=== Software === 
python       : 3.6.9
fastai       : 1.0.61.dev0
fastprogress : 0.2.2
torch        : 1.4.0
torch cuda   : 10.1 / is **Not available** 

=== Hardware === 
No GPUs available 

=== Environment === 
platform     : Linux-5.3.0-42-generic-x86_64-with-Ubuntu-18.04-bionic
distro       : #34~18.04.1-Ubuntu SMP Fri Feb 28 13:42:26 UTC 2020
conda env    : Unknown
python       : /home/david/Envs/tmpfastai/bin/python
sys.path     : 
/usr/lib/python36.zip
/usr/lib/python3.6
/usr/lib/python3.6/lib-dynload
/home/david/Envs/tmpfastai/lib/python3.6/site-packages
/home/david/Envs/tmpfastai/lib/python3.6/site-packages/IPython/extensions
no supported gpus found on this system
```

**To Reproduce**

1. Train a GANLearner with a **normalized** databunch. For example,

```
bs, n_channels, img_size = 128, 3, 16
path = untar_data("http://vis-www.cs.umass.edu/lfw/lfw")
data = (GANItemList.from_folder(path)
               .split_none()
               .label_from_func(noop)
               .transform(tfms=[[crop_pad(size=img_size, row_pct=(0,1), col_pct=(0,1))], []], size=img_size, tfm_y=True)
               .databunch(bs=bs)
               .normalize(stats = [torch.tensor([0.5,0.5,0.5]), torch.tensor([0.5,0.5,0.5])], do_x=False, do_y=True))
critic = basic_critic(img_size, n_channels, n_extra_layers=1)
generator = basic_generator(img_size, n_channels, n_extra_layers=1)
learner = GANLearner.wgan(data, generator, critic, switch_eval=False, 
                          opt_func = partial(optim.Adam, betas = (0.,0.99)), wd=0.)
learner.fit(1, 5e-4)
```

2. Show results:

`learner.show_results(ds_type=DatasetType.Train)`

The images shown now seem brighter that the one(s) shown during training, at the end of each epoch. It's doesn't seem a casual difference between specific images, the images shown during training are always darker.

#
**Expected behavior**

The images shown during training should have similar brightness to those ouput by `learner.show_results()`.

**Screenshots**

**Additional context**

I think any RGB value between 0 and 127 (or -1. and 0. for tensors) is being displayed as 0.
If I'm not wrong the reason is the **image isn't properly denormalized** in `GANTrainer.on_epoch_end`, so when the weights are clamped (from [-1, 1] to [0, 1]) by calling `data.train_ds.y.reconstruct()` (which is `ImageList.reconstruct()`), the darker values are lost and mapped to just black (0) for its channel.

I guess `do_x=True` needs to be enforced only for this particular call to `data.denorm` as in the case of GAN data the input images aren't stored as `x`, which is the input noise, but `y`.

I've made a (dirty) gist too: https://gist.github.com/davidleonfdez/8ff8e2ecc746c085f173debb2e114f85

I've submitted a PR. (#2525)

Thanks in advance.

davidleonfdez:

It's not easily testable as it depends on an intermediate result used in `GANTrainer.on_epoch_end`. It would require caching the intermediate result given by `data.denorm` just for the purpose of testing or coding some kind of heuristic about the pixel distribution of GANTrainer.imgs; probably not worth it.

It just enforces denormalization of the generated image shown at the end of any training epoch by `GANTrainer.on_epoch_end`. The call to `data.denorm()` was doing nothing. More details are given in the linked issue.
This is my first PR, hope it's ok! Let me know otherwise.

### DIFF
--- a/fastai/vision/gan.py
+++ b/fastai/vision/gan.py
@@ -138,7 +138,7 @@ class GANTrainer(LearnerCallback):
         data = self.learn.data
         img = self.last_gen[0]
         norm = getattr(data,'norm',False)
-        if norm and norm.keywords.get('do_y',False): img = data.denorm(img)
+        if norm and norm.keywords.get('do_y',False): img = data.denorm(img, do_x=True)
         img = data.train_ds.y.reconstruct(img)
         self.imgs.append(img)
         self.titles.append(f'Epoch {epoch}')


sgugger:
That should do the trick, thanks!

Issue #2527, opened 2020-03-27
sklearn.preprocessing does not have Imputer. Thus throws NoModuleError


aayushagrawal135:
The import `from sklearn.preprocessing import Imputer` fails in `fastai/old/fastai/structured.py` since `Imputer` in `sklearn` is moved to `sklearn/impute`

sgugger:
Closed by #2527

PR #2528, opened 2020-03-27
Change import directory for Imputer (#2527)

aayushagrawal135:
The import `from sklearn.preprocessing import Imputer` fails in `fastai/old/fastai/structured.py` since `Imputer` in `sklearn` is moved to `sklearn/impute`


sgugger:
Thanks for making this up to date!

PR #2530, opened 2020-03-30
Update setup.py

embecka:
Spacy dependency is turned off by default when installing fastai on Python 3.8 since Python 3.8 is not currently supported by Spacy.
This is the only dependency that prohibits installing fastai on Python 3.8.


sgugger:
Thanks!

PR #2534, opened 2020-04-05
Visible grids makes measuring LearningRate easier.

ParthikB:

 - [ I just started learning fastai, and as most of the people use a notebook, it's really hard to measure the learning rate from the plot { learner.recorder.plot() }. Thus, I thought that this minor addition would surely be a lot of help to everyone!] Add details about your PR.

### DIFF
--- a/fastai/basic_train.py
+++ b/fastai/basic_train.py
@@ -537,7 +537,7 @@ class Recorder(LearnerCallback):
         ys = spl(xs)
         return ys
 
-    def plot(self, skip_start:int=10, skip_end:int=5, suggestion:bool=False, return_fig:bool=None,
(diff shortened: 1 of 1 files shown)


sgugger:
Thanks but since it changes the behavior of the plot, I would like this to be controlled by a flag that defaults to False to not surprise anyone (I just tried it and it's actually less readable for me, probably because I'm used to those plots). Could you add that?

ParthikB:
> Thanks but since it changes the behavior of the plot, I would like this to be controlled by a flag that defaults to False to not surprise anyone (I just tried it and it's actually less readable for me, probably because I'm used to those plots). Could you add that?

Of course why not! I'm on it..

sgugger:
Thanks! Just changed to fit our usual style (if statements with only one line are set on the same line in fastai, to save vertical space).

PR #2538, opened 2020-04-07
Undersampling

nathanhubens:
I recently competed in this [Kaggle competition](https://www.kaggle.com/c/deepfake-detection-challenge), where one of the most used class balancing technique was UnderSampling. Indeed the dataset was such that OverSampling still lead to overfitting. My teammates and I used UnderSampling as a Callback to train our models so I wanted to add it to fastai because that might help other people as well.


sgugger:
Hi there, thanks for your PR! It looks good but I was thinking we should maybe put back the original training dataloader at the end of training to clean up? Something like
```
self.data.train_dl = self.data.train_dl.new(shuffle=True, drop_last=True, sampler=None)
```
What do you think?

nathanhubens:
Yes I agree, it was taken from the `OverSamplingCallback` but I also think it would look cleaner to remove the temporary assignment to `ds, dl` and have this: 

```python    
         self.labels = self.data.train_dl.y.items
         assert np.issubdtype(self.labels.dtype, np.integer), "Can only undersample integer values"
         _,self.label_counts = np.unique(self.labels,return_counts=True)
         if self.weights is None: self.weights = torch.DoubleTensor((1/self.label_counts)[self.labels])
         self.total_len_undersample = int(self.data.c*np.min(self.label_counts))
         sampler = WeightedRandomSampler(self.weights, self.total_len_undersample)
         self.data.train_dl = self.data.train_dl.new(shuffle=False, sampler=sampler)
```

If you agree with that change, I can also modify the `OverSamplingCallback` accordingly. Also I noticed that you added `drop_last=True`, should I add that ?

sgugger:
Yes, adding to `Oversampling` should be good. I was trying to reproduce the default `train_dl` but now that I think of it, it's probably better to just keep a reference to the train_dl before changing it, storing it and then set it back at the end (in case the train_dl was not built with the defaults).

nathanhubens:
> it's probably better to just keep a reference to the train_dl before changing it, storing it and then set it back at the end (in case the train_dl was not built with the defaults)

I'm not sure to fully understand what you advise there. You prefer the version I already committed where the `self.train_dl` was stored in the intermediate `dl` variable then re-assigned at the end to `self.train_dl` ? If so that's already what `OverSampling` is using so nothing's left to change 😊

sgugger:
Now I mean store it in the before fit event:
```
self.old_dl = dl
```
and then set it back in the after fit event:
```
self.data.train_dl = self.old_dl
```

nathanhubens:
Pushed a new version, tell me if everything's fine for you

sgugger:
Thanks a lot!

PR #2539, opened 2020-04-07
Added Verbose

ParthikB:
While downloading URLs.IMDB (which is a huge data-set), I didn't knew that it was being extracted after being downloaded, because no such information was printed to console of file being extracted. And as I was using Google Colab (which I think most of the students prefer), it takes a lot of time to extract the file contents. And I cancelled the process 4 times before acknowledging what's going on.

I think it's better that we just print the the on-going process for everyone's benefit.
(I added a flag just not to surprise anyone, though I don't think it's necessary)

 
 
### DIFF
--- a/fastai/datasets.py
+++ b/fastai/datasets.py
@@ -233,7 +233,7 @@ def untar_data(url:str, fname:PathOrStr=None, dest:PathOrStr=None, data=True, fo
         fname = download_data(url, fname=fname, data=data)
         if url in _checks:
             assert _check_file(fname) == _checks[url], f"Downloaded file {fname} does not match checksum expected! Remove that file from {Config().data_archive_path()} and try your code again."
-        if verbose: print('.tgz file downloaded. Extracting the contents...)
+        if verbose: print('.tgz file downloaded. Extracting the contents...')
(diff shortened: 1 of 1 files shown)


ParthikB:
With verbose=True
fastai

fastai/datasets.py
```diff
@@ -233,5 +233,7 @@ def untar_data(url:str, fname:PathOrStr=None, dest:PathOrStr=None, data=True, fo
         fname = download_data(url, fname=fname, data=data)
         if url in _checks:
             assert _check_file(fname) == _checks[url], f"Downloaded file {fname} does not match checksum expected! Remove that file from {Config().data_archive_path()} and try your code again."
+        if verbose: print('.tgz file downloaded. Extracting the contents...)
```
sgugger, line 236: You are missing a quote here.

sgugger:
Thanks for your PR, but there is a syntax error. Please fix before we can merge.

fastai/datasets.py
ParthikB: oops. On it.

sgugger:
Thanks!

PR #2540, opened 2020-04-08
Reading plots is easier with Grids

ParthikB:
Added a flag to show/hide grids in the plot. It makes reading the plots a lot easier.

 
 
Plots with and without the flag on:
fastai2


sgugger:
Thanks!

PR #2544, opened 2020-04-14
Update __init__.py

ketan-b:
fastai.vision import * -> name ‘learner’ is not defined issue resolved

 
 

sgugger:
No, you probably have some other problems in your install as the current init works without problem for many many users. Look at the [forum](https://forums.fast.ai/) to help debug your install.

PR #2546, opened 2020-04-20
BUG: ResBlock with stride>2

scottclowe:
The pooling kernel on the identity block needs to match the stride on the conv block so the shapes match up. Previously it was hard-coded to be a pooling kernel of 2 if the conv stride was not 1.

This bug has probably not been seen in the wild because ResBlock is only ever called with stride of 1 or 2.

 
 

sgugger:
Indeed, thanks for fixing!

PR #2556, opened 2020-04-29
Fix ImageList.from_df failure for data frame with one row

### LINKED ISSUE #2555: ImageList.from_df fails for data frame with one row
**Describe the bug**
`ImageList.from_df` fails for data frame with one single row before checking if the file path is valid.

**Provide your installation details**
```text
=== Software === 
python       : 3.7.4
fastai       : 1.0.60
fastprogress : 0.2.2
torch        : 1.4.0
torch cuda   : None / is **Not available** 

=== Hardware === 
No GPUs available 

=== Environment === 
platform     : Darwin-19.4.0-x86_64-i386-64bit
conda env    : base
python       : /Users/jiahuawu/opt/anaconda3/bin/python3
sys.path     : 
/Users/jiahuawu/opt/anaconda3/lib/python37.zip
/Users/jiahuawu/opt/anaconda3/lib/python3.7
/Users/jiahuawu/opt/anaconda3/lib/python3.7/lib-dynload
/Users/jiahuawu/opt/anaconda3/lib/python3.7/site-packages
/Users/jiahuawu/opt/anaconda3/lib/python3.7/site-packages/aeosa
/Users/jiahuawu/opt/anaconda3/lib/python3.7/site-packages/IPython/extensions
no supported gpus found on this system
```
**To Reproduce**
```
>>> from fastai.vision import *
>>> df = pd.DataFrame(["123.png"], columns=["names"])
>>> ImageList.from_df(path=".", df=df)
  File "/Users/jiahuawu/opt/anaconda3/lib/python3.7/site-packages/fastai/data_block.py", line 77, in __repr__
    items = [self[i] for i in range(min(5,len(self.items)))]
TypeError: len() of unsized object
```

**Expected behavior**
An ImageList with one item should be created if the path is valid.

**Additional context**
Based on my investigation in the code base, the problem is that [`_maybe_squeeze`](https://github.com/fastai/fastai/blob/master/fastai/data_block.py#L13) returns a 0d array for data frame with one single row while 1d array for dataframe with multiple rows.

JiahuaWU:
Fix #2555 and add test to prove the fix.

### DIFF
--- a/fastai/data_block.py
+++ b/fastai/data_block.py
@@ -135,7 +135,8 @@ class ItemList():
         "Create an `ItemList` in `path` from the inputs in the `cols` of `df`."
         inputs = df.iloc[:,df_names_to_idx(cols, df)]
         assert not inputs.isna().any().any(), f"You have NaN values in column(s) {cols} of your dataframe, please fix it."
-        res = cls(items=_maybe_squeeze(inputs.values), path=path, inner_df=df, processor=processor, **kwargs)
+        items = _maybe_squeeze(inputs.values) if len(df) > 1 else (inputs.values[0] if not isinstance(cols, Collection) or len(cols) == 1 else inputs.values)
+        res = cls(items=items, path=path, inner_df=df, processor=processor, **kwargs)
         return res
 
     @classmethod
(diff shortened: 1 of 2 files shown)


sgugger:
Thanks, in particular for adding the test! (MacOS tests have been broken for a while but tests are passing otherwise)

PR #2558, opened 2020-05-01
Add num channels parameter

cheeseblubber:
Want to be able to customize the number of channels in the cnn_learner method
 - [x ] Read the [contributing docs](https://github.com/fastai/fastai/blob/master/CONTRIBUTING.md)


cheeseblubber:
Noticed it was taking num channel from .c on th edata

PR #2562, opened 2020-05-07
Patch 1

jss367:
I downloaded both the PASCAL_2007 and PASCAL_2012 datasets and my checksums don't match the ones in the code and my file sizes appear to be slightly larger. I also downloaded the Skin Lesion dataset to test and my checksum matched, so I believe there is something different with the PASCAL downloads now. Can someone else try it and see if theirs matches mine or the original code's?

 - [X ] Read the [contributing docs](https://github.com/fastai/fastai/blob/master/CONTRIBUTING.md)
 - [X ] Add details about your PR.


sgugger:
I double-checked and those are the correct values indeed. I think that's because we added some csv files in those dataset to make it easy to use pascal for multi-label problems and forgot to update those checks.

Thanks for fixing!

PR #2565, opened 2020-05-08
Fixed a bug in learner.py:

tomersa:
AttributeError: backwards

 
 

sgugger:
No, this does not check if the attribute is `True` anymore, just that it exists. So it's not a real fix. It should be something like `getattr(data, 'backward', False)` to check if it's there with a default to `False` if it's not.

sgugger:
Please reopen if you want to address the comments up there, closing since there hasn't been any activity in a while.

PR #2566, opened 2020-05-08
update data.py

sakamoto-gintoki:
the original says that  list is unhashable when we call TabularDataBunch.from_df().

 
 

sgugger:
Not sure it will work fully with a list of dependent variables, but this is a first step, thanks!

PR #2569, opened 2020-05-09
fixed a bug in vision data.py

atul006:
while creating dataset "from_folder", the function picks training images from folders other than train/valid, if present and assign that folder name as a class. Changed it to pick training images only from 'train'.
 
 

sgugger:
No, this breaks everything in the library. Note that if the factory methods of `ImageDataBunch` do not help you in your problem, you should switch to the data block API. Those factory methods are not guaranteed to work in every use case you may have, they are just tools for easy data creation for beginners.

PR #2573, opened 2020-05-11
 Fix for the type mismatch error that occurs at the call to _th_addmm_out.

rraminen:
The error, " RuntimeError: Expected object of scalar type Double but got scalar type Float for argument #3 'mat2' in call to _th_addmm_out" occurs in the fasatai dl1 notebook: lesson3-head-pose.ipynb.
image

This occurs when c.flow is a tensor of type Double, m[:2,2] and a are tensors of type Float. 
So defining **flow.flow**  as Float ( in the file: https://github.com/fastai/fastai/blob/master/fastai/vision/image.py) as in the below picture helps to resolve this error.
image


sgugger:
Good catch, thanks!

PR #2585, opened 2020-05-31
Make EfficientNet compatible with cnn_leaner

JiahuaWU:
Although EfficientNet has been made available in [this pull request](https://github.com/fastai/fastai/pull/2299), it's not in form of an usual sequential model and it cannot be employed via `cnn_learner`.

This PR aims at transforming the pytorch implementation of efficientnet into a sequential model and therefore making it work with `cnn_learner`.

Through this pull request, users can directly use EfficientNet via `cnn_learner` like other well-supported networks. In particular, they can initialise a `cnn_learner` with EfficientNet:
 ```learn = cnn_learner(data, models.EfficientNetB0, pretrained=True, cut=-2)```
and use full functionalities such as freeze/unfreeze and split with desired effects.

Thanks in advance for your review/comments!


jph00:
Thanks for this PR! We're not adding new functionality to v1 - but if you're interested, we'd be happy to consider this functionality for v2, after it's released in a few weeks. BTW, you might be interested in checking out the implementations here:
https://github.com/rwightman/pytorch-image-models

PR #2589, opened 2020-06-11
Enhance the dynamicUnet to support efficientnet

Flyfoxs:
- [X] Read the [contributing docs](https://github.com/fastai/fastai/blob/master/CONTRIBUTING.md)
 - [X] Include a test in your PR that fails without your code, and passes with it, as well as a test of a case that already worked without your code (and still works with it). Currently fastai has poor test coverage, so don't take the current tests as a role model - we're all working to fix it together!
 - [X] Add details about your PR.

DynamicUnet design is very cool, but it is not support EfficientNet as decoder, even it can get better score in most of the scenarios. 

The reason is the structure of EfficientNet is diff from normal structure.
- EfficientNet has ModuleList block, which can not forward.
- in_channels can not calculate the input channel for EfficientNet

Here is one of the simple test case to validate it

`
import torch
from torch import nn
from fastai.vision import models
from efficientnet_pytorch import EfficientNet

def eff():
    class EfficientNet_(EfficientNet):
        def __init__(self, *args, **kwargs):
            super().__init__(*args, **kwargs)
        def forward(self, inputs):
            x = self.extract_features(inputs)
            return x
    return EfficientNet_.from_pretrained(f'efficientnet-b0', in_channels=3)

for encoder in [eff(), 
                nn.Sequential(*list(models.resnet34().children())[:-3])]:
    unet =  models.unet.DynamicUnet(encoder, n_classes=5, img_size=(224, 224)) 
    print(type(encoder), unet(torch.rand(1,3,224,224)).shape)

`

### DIFF
--- a/fastai/torch_core.py
+++ b/fastai/torch_core.py
@@ -258,7 +258,7 @@ def apply_init(m, init_func:LayerFunc):
 
 def in_channels(m:nn.Module) -> List[int]:
     "Return the shape of the first weight layer in `m`."
-    for l in m.children():
+    for l in m.modules():
         if hasattr(l, 'weight'): return l.weight.shape[1]
     raise Exception('No weight layer')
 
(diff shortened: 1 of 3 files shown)


jph00:
We're not adding new features to v1, sorry! But please consider contributing this functionality to v2 when it's out in a few weeks.

PR #2592, opened 2020-06-15
Fix/improve accuracy calculation:

ttong-ai:
1. view(n,-1) will generate an nx1 2D Tensor.  This will cause problem if passed in targs is len(n) 1D Tensor, which is typically the case.
2. Add cases to support whether input and targs are in one-hot vectorized labels or already with argmax operation. In the return statement, both `input` and `targs` are len(n) 1D Tensor.

 
 

jph00:
Could you please include tests in your PR that check the behavior of all expected tensor shapes?

PR #2623, opened 2020-08-10
Fix misidentification of in channels when using grouped convolution

yasrynn:
 - [x] Include a test in your PR that fails without your code, and passes with it.

Torch's `Conv2d` class allows a `groups` flag for grouped convolution to segregate channels into groups, for example to run parallel images at once.  When using the flag, the `weight` tensor is `out_channels` x `in_channels // groups` x `kernel_size` x `kernel_size`, while the module expects input with `in_channels` channels. This means the `in_channels` function will misidentify the channels the module is expecting when `groups` is not 1. This can cause errors when doing a `dummy_eval`, for example.

This PR provides a fix for that, as well as a test that reveals the misidentification.


jph00:
Thank you :)

PR #2632, opened 2020-08-18
Fix min version for fastcore requirement

albertvillanova:
Renaming of `Tuple` to `fastuple` was tagged as version 0.1.35.


jph00:
Good spot!

Issue #2633, opened 2020-08-18
add_datepart() fails if there are NaT rows in a datetime column


kevinbird15:
Please confirm you have the latest versions of fastai, fastcore, fastscript, and nbdev prior to reporting a bug (delete one): YES
fastai2 0.0.30
fastcore 0.1.38
fastscript 0.1.13
nbdev 0.2.40

**Describe the bug**
If a datetime column has any null rows, the add_datepart function will fail.  

**To Reproduce**
1. See code:
```
import pandas as pd
import numpy as np
from fastai2.tabular.all import *

tmp_df = pd.DataFrame([['8/17/2020'],[np.nan],['8/18/2020']], columns=["datefield"])
add_datepart(tmp_df, 'datefield')
```
**Expected behavior**
I would expect everything to populate with None values but maybe other people would expect different behavior. 

**Error with full stack trace**

Place between these lines with triple backticks:

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-1-b327ea20eb9a> in <module>
      4 
      5 tmp_df = pd.DataFrame([['8/17/2020'],[np.nan],['8/18/2020']], columns=["datefield"])
----> 6 add_datepart(tmp_df, 'datefield')

/home/kevin/fastai/fastai2/tabular/core.py in add_datepart(df, field_name, prefix, drop, time)
     32     if time: attr = attr + ['Hour', 'Minute', 'Second']
     33     for n in attr: df[prefix + n] = getattr(field.dt, n.lower())
---> 34     df[prefix + 'Elapsed'] = field.astype(np.int64) // 10 ** 9
     35     if drop: df.drop(field_name, axis=1, inplace=True)
     36     return df

/opt/anaconda3/envs/fastai2/lib/python3.8/site-packages/pandas/core/generic.py in astype(self, dtype, copy, errors)
   5535         else:
   5536             # else, only a single dtype is given
-> 5537             new_data = self._mgr.astype(dtype=dtype, copy=copy, errors=errors,)
   5538             return self._constructor(new_data).__finalize__(self, method="astype")
   5539 

/opt/anaconda3/envs/fastai2/lib/python3.8/site-packages/pandas/core/internals/managers.py in astype(self, dtype, copy, errors)
    565         self, dtype, copy: bool = False, errors: str = "raise"
    566     ) -> "BlockManager":
--> 567         return self.apply("astype", dtype=dtype, copy=copy, errors=errors)
    568 
    569     def convert(

[...]

PR #2640, opened 2020-08-19
modifying add_datepart to work with null rows

kevinbird15:
This PR is created to fix the issue brought up in #2633.  A test was also added test that things worked as expected.


marii-moe:
Seems we are getting the same error in 'CI / test-nbdev-sync (pull_request) '. If you figure out what is causing it I would like to know too. I'm on discord.

jph00:
Looks good!

PR #2643, opened 2020-08-20
Fix typo in tutorial.ipynb

joe-bender:
Changed 'if' to 'it'


jph00:
I'll change it locally since we have a conflict.

PR #2646, opened 2020-08-21
Update CI to reference fastai instead of fastai2

hamelsmu:


hamelsmu:
@jph00 CI has been switched to the new containers pointing at `fastai` instead of `fastai2`

Issue #2636, opened 2020-08-19
metrics and predictions inconsistent when using to_distributed(...) 


antorsae:
DataDistributedParallel(DDP) funcionality is incomplete in fastai2:

fastai1 has its own recorder for DDP https://github.com/fastai/fastai1/blob/master/fastai/distributed.py#L49 but fastai2 doesn't https://github.com/fastai/fastai/blob/master/nbs/20a_distributed.ipynb as a metrics do not work in DDP in fastai2 (and by extension if you rely on them to save best epoch for example they misguide you).

jph00:
I'll try to look at this next week - feel free to do a PR in the meantime if you're interested.

antorsae:
I think I reported this in error. `to_detach(...)` gave me an error when in DDP inside `AccumMetric` so I mistakenly did `.detach().cpu()` which disabled the offending code bit that gave the error:

`torch.distributed.all_gather(res, x if ndim > 0 else x[None])`

but it is needed in DDP mode.

Now that I understand all this a bit better I've change all gather to make `x` contiguous and it works.

I am hesitant to do a PR b/c it is 1 line change in https://github.com/fastai/fastai/blob/master/nbs/03_data.core.ipynb but I don't know how to do proper test case for this:

```
#export
def maybe_gather(x, axis=0):
    "Gather copies of `x` on `axis` (if training is distributed)"
    if num_distrib()<=1: return x
    ndim = x.ndim
    res = [x.new_zeros(*x.shape if ndim > 0 else (1,)) for _ in range(num_distrib())]
    torch.distributed.all_gather(res, x.contiguous() if ndim > 0 else x[None])
    return torch.cat(res, dim=axis) if ndim > 0 else torch.cat(res, dim=axis).mean()
```

Also, less impactful but the validation metrics are still not perfect b/c the DDP wrapping DL `DistributedDL` pads batches to be evenly divisible:

```
    def sample(self):
        idxs = self.get_idxs()
        if self.shuffle: idxs = self.shuffle_fn(idxs)
        # add extra samples to make it evenly divisible
        idxs += idxs[:(self.total_n - len(idxs))]
        # subsample
        idxs = idxs[self.rank:self.total_n:self.world_size]
        return (b for i,b in enumerate(idxs) if i//(self.bs or 1)%self.nw==self.offs)
```

and this is not accounted for (as far as I can see) anywhere else.

jph00:
[...]

PR #2649, opened 2020-08-21
DistributedDataParallel fix (gather non-contiguous tensor) and improvement (sync'd batch norm)

antorsae:
This PR fixes https://github.com/fastai/fastai/issues/2636 and also enables and turns on by default synched batch normalization in DDP mode.


jph00:
Nice job!

PR #2653, opened 2020-08-22
Fix link to quick start

rohitpaulk:
👋


jph00:
Needs to be fixed in index.ipynb

PR #2657, opened 2020-08-22
Expand on tabular documentation, clean up predict, and allow TabularModel to have custom activation functions

muellerzr:
As the title says, expands on most of the tabular documentation. Predict was cleaned up to only need a few lines, and a slight adjustment to `TabularModel` was performed to allow for custom activation functions (in case people want to experiment easily). Along with this, Tutorial Tabular now also includes using `TabularPandas` to integrate with other libraries such as `xgboost`

Let me know if there are any adjustments that should be made!


jph00:
Looks good!

PR #2673, opened 2020-08-24
Solves "FutureWarning: Series.dt.weekofyear and Series.dt.week have been deprecated.  Please use Series.dt.isocalendar().week instead."

KushajveerSingh:
In `add_datepart`, we get this warning. This is because `df[date].dt.week` will be moved to `df[date].dt.isocalendar().week`. This PR solves the issue. Also, the order of the columns is preserved i.e. `Week` column after `Month` and before `Day`.


jph00:
Thank you :)

jph00:
Thank you :)

Issue #2675, opened 2020-08-24
Add docs to DataLoader


jph00:
fastai's DataLoader currently is missing documentation for every method! Use `add_docs` to add it.

san7988:
Hi Jeremy,

I believe we are talking about fastai/data/load.py. Is that correct?
 Also, your last commit removed a line of documentation from above file and the corresponding notebook as well. Was that documentation redundant?

Thanks

PR #2676, opened 2020-08-24
Add docs for DataLoader

KushajveerSingh:
#2675 

Started working on this. I will try to complete the remaining functions tomorrow. Also, added a separate section in the documentation showing how to use *iterable style dataset*. 

@jph00 Can you review some of the changes, to see if I am going in the right direction and the implementation is correct?


jph00:
Thanks for this! Take a look at the docs for any of the other classes, such as `L` in `fastcore`, to see how things should look. Each method name and its docstring needs to fit on one line. (Also there's some conflicts to resolve)

antorsae:
A few comments:

> shuffle (bool): If `True`, then data is shuffled after every epoch.
Instead of every epoch, I'd say every time dataloader is fully read/iterated.

> get_idxs = Returns a list of indices to be used in the next epoch. If `shuffle=True`, then the indices are passed to `shuffle_fn` which returns shuffled indices. In case of iterable-style dataset, it returns a list of `[None]*n`.
Returns a list of indices to reference the dataset. It is responsible of calling `shuffle_fn` internally if `shuffle=True`. 

> shuffle_fn = Returns a random permutation of `idxs`. `idxs` is a list of items (can be indices, or anything).
care must be taken to use python's `random` functionality (vs numpy or pytorch's) due to the way random seeds are sync'd between subprocesses inside `_FakeLoader`.

KushajveerSingh:
> care must be taken to use python's random functionality (vs numpy or pytorch's) due to the way random seeds are sync'd between subprocesses inside _FakeLoader.

@antorsae Didn't know that. Is it a pytorch internal thing? Can you guide me on how you know about this thing?

Also, do you know what is the use of `wif` method. I am not able to find a link on its usage.

jph00:
Good points from @antorsae . Some of those should be in markdown prose, not in docstrings, since docstrings should be just a brief sentence on one line in fast.ai libs.

antorsae:
`wif` is an abbreviation for `worker_initialization_fn` (yes I know what you're thinking, wlcme to fstai)

This is all internal stuff inspired by pytorch multiprocessing; some background: https://pytorch.org/docs/stable/data.html#single-and-multi-process-data-loading.

I correct what I said about about "care must be taken to use python's random functionality (vs numpy or pytorch's) due to the way random seeds are sync'd between subprocesses inside _FakeLoader."; what happens is:

a) when/if subprocesses are created (in _FakeLoader) each subprocess has its own different seed, which gets initialized via `set_seed(...)`:

```
def _wif(worker_id):
    set_num_threads(1)
    info = get_worker_info()
    ds = info.dataset.d
    ds.nw,ds.offs = info.num_workers,info.id
    set_seed(info.seed)
    ds.wif()
```

note that that `info.seed` is actually different for each worker/subprocess, so each subprocesses cannot shuffle the indexes carelessly b/c that would result in duplicated items; so:

b) fastai DataLoader has a `self.rng` random number generator initialized and maintained in the main process. calls to `self.randomize()` which happens in the main process.

c) when/if subprocesses are created inside `__iter__` the main process `self.rng` is copied to forked subprocesses, which may call `shuffle_fn` which uses that consistent random number generator which is the same across subprocesses; hence rendering consistent behavior.

A lot of care must be put when shuffling/sampling b/c behavior may silently break if not careful. I am going to check if `WeightedDL` is OK or not.

KushajveerSingh:
> Some of those should be in markdown prose, not in docstrings, since docstrings should be just a brief sentence on one line in fast.ai libs.

Had started working on this. I realized the better approach later.

Thanks @antorsae for that explanation. I just have one more question regarding fastai implementation of `DataLoader`. `DataLoader` does not use `wif`. `_FakeLoader` calls `_wif`, which sets the worker info that you mentioned in your previous message. But the `wif` function passed in `DataLoader` is not used in the code or is passed to `_FakeLoader`, so how is it happening in this case? (Am I still missing something?)

jph00:
On Tue, Aug 25, 2020, at 9:51 AM, antorsae wrote:
> `wif` is an abbreviation for `worker_initialization_fn` (yes I know what you're thinking, wlcme to fstai)


We should at least document that abbreviation. If we wanted to use a longer name, I'd go with ``worker_init_func`, since `init` is a well known abbrev, and `fn` is ambiguous (we normally use it for `filename` and not `function`).`

jph00:
> A lot of care must be put when shuffling/sampling b/c behavior may silently break if not careful. I am going to check if `WeightedDL` is OK or not.


By the way I'm not sure anyone has actually looked carefully at my DataLoader code before, so if you think of any approaches that might make things simpler and more reliable, do say so! I'm not happy with how subtle things are right now. The user API for sampling really needs to be improved so that users don't need to know or care about this inner details.

jph00:
> Thanks @antorsae <https://github.com/antorsae> for that explanation. I just have one more question regarding fastai implementation of `DataLoader`. `DataLoader` does not use `wif`. `_FakeLoader` calls `_wif`, which sets the worker info that you mentioned in your previous message. But the `wif` function passed in `DataLoader` is not used in the code or is passed to `_FakeLoader`, so how is it happening in this case? (Am I still missing something?)


It's called in the line `ds.wif()`, which is park of `FakeLoader`.

KushajveerSingh:
@jph00 Thanks for the pointer. I realized what I was missing. Each worker contains the information of the `DataLoader`. Simple things that don't point out, when you don't know stuff.

jph00:
Yeah it's complex code, because the PyTorch DataLoader foundations are not at all well decoupled, so trying to use them externally is a big ugly hack :(

antorsae:
OK, my fear materialized:

Mind my ugly `print` for debugging:
```
#export
@log_args(but_as=TfmdDL.__init__)
@delegates()
class WeightedDL(TfmdDL):
    def __init__(self, dataset=None, bs=None, wgts=None, **kwargs):
        super().__init__(dataset=dataset, bs=bs, **kwargs)
        wgts = array([1.]*len(dataset) if wgts is None else wgts)
        self.wgts = wgts/wgts.sum()

    def get_idxs(self):
        if self.n==0: return []
        if not self.shuffle: return super().get_idxs()
        r = list(np.random.choice(self.n, self.n, p=self.wgts))
        print(f"{os.getpid()} => {r}")
        return r
```
```
list(WeightedDL(list(range(50)),wgts=[1] * 50,bs=32,shuffle=True,num_workers=3))
```
As pointed out `get_idxs` gets executed in each subprocesses, and yields different indexes for each subprocess:
```
116188 => [25, 39, 49, 35, 27, 47, 14, 31, 45, 47, 43, 3, 41, 22, 19, 28, 49, 31, 21, 31, 23, 47, 38, 27, 31, 46, 36, 12, 28, 5, 16, 7, 22, 33, 3, 21, 34, 13, 41, 0, 40, 7, 14, 3, 13, 49, 6, 4, 15, 42]
116191 => [37, 10, 2, 38, 33, 32, 40, 29, 10, 13, 24, 42, 11, 40, 39, 27, 24, 8, 21, 19, 4, 34, 12, 45, 33, 46, 5, 32, 36, 31, 20, 9, 35, 2, 15, 21, 17, 36, 48, 43, 48, 15, 25, 6, 8, 37, 11, 17, 37, 8]
116196 => [14, 46, 33, 42, 3, 9, 45, 17, 27, 30, 23, 19, 43, 37, 21, 12, 10, 38, 16, 11, 31, 6, 40, 8, 9, 37, 8, 14, 38, 21, 43, 6, 24, 34, 45, 46, 24, 49, 45, 38, 39, 11, 13, 23, 20, 33, 3, 8, 16, 18]
```
...which is essentially the inconsistencies pytorch documentation warns you against when using `seed`. 

The reason why this happens is:
a) `get_idxs` did not use `self.rng`
b) `get_idxs` is a called in the context of _each_ subprocess.

I'm going to submit a PR to fix this. I do not think it is realistic to impose such a low-level and subtle detail to `get_idxs`.

Issue #2637, opened 2020-08-19
to_distributed() wrapped DataLoader ignores get_idxs() so WeightedDL behaves inconsistently in DDP mode


antorsae:
If your dataloader  has its own get_idxs eg.  `WeightedDL` it gets ignored when `to_distributed(...)`, so as a result same code that e.g. does minority oversampling (using `WeightedDL`) will silently NOT do it in DDP mode, b/c DDP callback recreates wrapped dataloader from its Datasource directly ignoring any overridden `get_idxs` function you may had.

jph00:
Thanks will check this out.

PR #2679, opened 2020-08-25
DistributedDL to be get_idxs aware (so it works properly with WeightedDL)

antorsae:
This fixes https://github.com/fastai/fastai/issues/2637

Instead of trying to have all ranks be in sync wrt get_idxs, it avoids the problem by computing get_idxs in rank 0 and broadcasting get_idxs to all ranks (this only happens once at the beggining, so not perf impact); the only assumption about get_idxs is that it must returl the same number of items across executions.


jph00:
Well done!

PR #2682, opened 2020-08-25
Update main.yml

hamelsmu:
@jph00 lets try this and see if it works


hamelsmu:
I couldn't find a cleaner way to filter runs than this

PR #2687, opened 2020-08-25
Consistent sampling behavior when overriding get_idxs when using >1 num_workers

antorsae:
Before this PR, this:

```
class AdamantDL(DataLoader):
    def get_idxs(self):
        r=random.randint(0,self.n-1)
        return [r] * self.n
test_eq(torch.cat(tuple(AdamantDL((list(range(50))),bs=16,num_workers=4))).unique().numel(),1)
```
would fail b/c `get_idxs` was being invoked within each worker. This PR makes sure `get_idxs` gets called from main process so no special considerations are needed to make sure workers maintain consistent indices. This simplifies `DistributedDL` a bit to, since it had to deal w/ this issue in its own way.


antorsae:
Included in https://github.com/fastai/fastai/pull/2729

PR #2651, opened 2020-08-22
update the show function so that you view all frames if the .dcm image has more than 1 frame per file

asvcode:
Most .dcm datasets have 1 frame per file but some datasets have more than 1 frame per .dcm file. In this case the show function fails with a type error.  This update allows the show function to check for the number of ndims and specify how many frames to view (default being set at 1).

PR #2690, opened 2020-08-25
update the show function so that you view all frames if the .dcm image has more than 1 frame per file

asvcode:
This is a re-submission of `#2651` to resolve conflicts

Most .dcm datasets have 1 frame per file but some datasets have more than 1 frame per .dcm file. In this case the show function fails with a type error. This update allows the show function to check for the number of ndims and specify how many frames to view (default being set at 1).


hamelsmu:
@jph00 failing to pull data from files.fast.ai again

jph00:
Thank you!

PR #2696, opened 2020-08-26
Not existing autocast causing explosion of cuda memory.

richarddwang:
When using `native_fp16()` and try to `get_preds` on a test set which **with no labels**, it will run into `if len(self.yb)==0: return` and not do `self.autocast.__exist__`. And as batches iterated, the occupied cuda memory is keeping increasing and then cuda out of memory.

After this fix, it now consumes constant memory.


jph00:
Many thanks - this is very helpful. I've fixed it ( believe) in a more general way:
https://github.com/fastai/fastai/commit/2c18fe5e03a8628625b436b7141313674c6d73d0

PR #2698, opened 2020-08-26
deal with non-Boolean selectors in `ColSplitter()` 

jerzydziewierz:
There is an unexpected behaviour in `ColSplitter()` in `vision/transforms.py` :

``` python
valid_idx = (o.iloc[:,col] if isinstance(col, int) else o[col]).values`
```

this means that if we have a DataFrame:
``` python
df = pd.DataFrame({'a': [5,6,7,8,9], 'is_valid': [1,0,1,1,0]})
splits_by_int = ColSplitter('is_valid')(df)
```
then it would fail without raising an error. It would create a broken validation set with only two observations `5, 6` repeated 5 times: `[6, 5, 6, 6, 5]`. This is unexpected.

**Expected output:** `[5,7,8]`

**Fix:** Make sure that the selector is boolean by casting to boolean.

**Alternate fix:** raise a helpful error when the selector is non-boolean.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.idea/
 docs/
 conda/
 tmp/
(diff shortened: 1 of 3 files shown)


jerzydziewierz:
sorry, need to fix one thing still

PR #2700, opened 2020-08-26
Allow for `get_preds` to return back the classes if `with_decoded` is True

muellerzr:
Let me know your thoughts as to if this could be done better, but this scenario seems the best?

Also what would you rather have `dec_out` be named? `y`?


muellerzr:
(Don't merge yet, tabular's `predict` fails, looking into it)

jph00:
No we can't really do it like this, with hard-coded callback names and other hacks. We need to come up with a flexible decoupled API. Feel free to create an issue explaining exactly what you're looking for, with some examples, and we can always get to it once we look at new features.

muellerzr:
Sure, sounds good. Apologies!

PR #2708, opened 2020-08-27
Adding a test for lr_find model weights

kevinbird15:
Adding a test to check that the model weights are the same before and after an lr_find are ran.


jph00:
Good idea

Issue #2652, opened 2020-08-22
Readme links return 404 and need to be updated


robmarkcole:
Links from the [readme](https://github.com/fastai/fastai/blob/master/README.md) which are 404:
* https://github.com/fastai/fastai/blob/master/quick_start
* https://github.com/fastai/fastai/blob/master/tutorial
* https://github.com/fastai/fastai/blob/master/migrating_pytorch (under `Plain PyTorch`)
* https://github.com/fastai/fastai/blob/master/migrating_ignite (under `Ignite`)
* https://github.com/fastai/fastai/blob/master/migrating_lightning (under `Lightning`)
* https://github.com/fastai/fastai/blob/master/migrating_catalys (under `Catalyste`)

hamelsmu:
We should try to fix nbdev so that the right links are shown rather than referencing the absolute URLS as I attempted to do in #2680  

cc: @jph00

tyoc213:
@jph00 links on https://github.com/fastai/fastai/blob/master/README.md are OK

jph00:
Thanks!

PR #2710, opened 2020-08-27
fix links to migration guides

aseifert:


aseifert:
oh snap, i just realized that these links might be as intended (see issue #2652 )

jph00:
Well, they're not as intended, but it is intended that I fix nbdev to make them correct! :)

PR #2714, opened 2020-08-27
updated xresnet to support slicing

### LINKED ISSUE #2677: Can't slice xResnet
In order to control for the issue of local setup, I am running this using the official docker container, and the latest version of the fastai code, up-to-date as of 5 minutes before time of posting. 

Please confirm you have the latest versions of fastai, fastcore, fastscript, and nbdev prior to reporting a bug (delete one): YES

**Describe the bug**
It is not possible to slice xResnet() produced Sequential. 

**To Reproduce**
Run this: 
```
from fastai.vision.models.xresnet import *; xresnet34()[0:2]
```
Gist: https://colab.research.google.com/gist/marii-moe/5a43207416cc3af80831c7df2c0717c8/01_intro.ipynb
Error is slightly different as colab uses older version.

**Expected behavior**
Returning with only the first two modules in xResnet34 without throwing an error.  

```
<ipython-input-13-1ed37b9f1a12> in <module>
----> 1 xresnet34()[0:2]

/opt/conda/lib/python3.7/site-packages/torch/nn/modules/container.py in __getitem__(self, idx)
     79     def __getitem__(self: T, idx) -> T:
     80         if isinstance(idx, slice):
---> 81             return self.__class__(OrderedDict(list(self._modules.items())[idx]))
     82         else:
     83             return self._get_item_by_idx(self._modules.values(), idx)

TypeError: __init__() missing 2 required positional arguments: 'expansion' and 'layers'
```

**Additional context**
I am currently working to fix this one, looking for suggestions on how you would like it fixed. Will probably do a bit of documentation in the area as well after this fix.

marii-moe:
This is just an idea, can rewrite if needed, main goal is to support this test: assert(len(xse_resnet18()[0:2])==2) 

@jph00 

Fixes #2677


jph00:
You should revert the capitalization to `XResNet`, since `ResNet` is how it's spelled in PyTorch.

Why move from using `__init__` to a class method? (In general, we should aim to make most things available through `__init__`, because that's the usual expected way to construct an object, and it's less verbose - but I'm open to using a class method if there's big enough benefits)

marii-moe:
I agree 100%. This was probably something I justified more in my head after spending too much time coding it. I'll try again if I come up with a cleaner solution.

Issue #2715, opened 2020-08-28
TabularDataLoaders.from_csv() has no way to easily strip leading or trailing white spaces


bguan:
Please confirm you have the latest versions of fastai, fastcore, fastscript, and nbdev prior to reporting a bug: YES (fastai 2.0.2)

**Describe the bug**
When trying to debug why my tabular learner is unable to make sensible prediction when following the tutorial learning the adult sample salary dataset, I believe I found a bug in ***TabularDataLoaders.from_csv()*** that calls ***pd.read_csv(csv)***.  The adult CSV has leading spaces after delimiter!  See 
```
> head ~/.fastai/data/adult_sample/adult.csv 
age,workclass,fnlwgt,education,education-num,marital-status,occupation,relationship,race,sex,capital-gain,capital-loss,hours-per-week,native-country,salary
49, Private,101320, Assoc-acdm,12.0, Married-civ-spouse,, Wife, White, Female,0,1902,40, United-States,>=50k
44, Private,236746, Masters,14.0, Divorced, Exec-managerial, Not-in-family, White, Male,10520,0,45, United-States,>=50k
38, Private,96185, HS-grad,, Divorced,, Unmarried, Black, Female,0,0,32, United-States,<50k
38, Self-emp-inc,112847, Prof-school,15.0, Married-civ-spouse, Prof-specialty, Husband, Asian-Pac-Islander, Male,0,0,40, United-States,>=50k
42, Self-emp-not-inc,82297, 7th-8th,, Married-civ-spouse, Other-service, Wife, Black, Female,0,0,50, United-States,<50k
20, Private,63210, HS-grad,9.0, Never-married, Handlers-cleaners, Own-child, White, Male,0,0,15, United-States,<50k
49, Private,44434, Some-college,10.0, Divorced,, Other-relative, White, Male,0,0,35, United-States,<50k
37, Private,138940, 11th,7.0, Married-civ-spouse,, Husband, White, Male,0,0,40, United-States,<50k
46, Private,328216, HS-grad,9.0, Married-civ-spouse, Craft-repair, Husband, White, Male,0,0,40, United-States,>=50k
```
[...]

PR #2716, opened 2020-08-28
Fix for 2715: expose skipinitialspace in TabularDataLoaders.from_csv()

bguan:
As discussed in discord ***#fastai-dev*** for #2715, this fix exposes skipinitialspace which defaults to True, so that panda.read_csv() can trim leading white spaces by default.

I wrote some test using adult.csv as example extending what was already there. I was originally thinking of writing a test using simpler temp csv files with hand populated values so it is not dependent on external data sources. I am open to suggestions on how to make the code better and how to make the test more robust.


nbs/41_tabular.data.ipynb
```diff
@@ -75,9 +75,9 @@
     "        return to.dataloaders(path=path, **kwargs)\n",
```
jph00, line 75: <p>It's better to use markdown prose than comments in this situation. Remember - this becomes documentation, it's not just a test. So perhaps create a markdown cell above this that explains that this is how you can do a prediction with tabular? I'm not sure you need <em>all</em> those <code>test_ne</code>  lines BTW. Perhaps just keep one of them?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/fastai/fastai/pull/2716/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='16'/>

bguan: <p>Got it. I've moved the comments to a markdown cell and cut down the redundant <code>test_ne()</code></p>

PR #2719, opened 2020-08-28
Added new parameter to wandb callback and a new way to log predictions

EtienneT:
When calling learn.get_preds, the _reorder_ parameter can cause problems for models that returns tuples from their forward method.  For example a VAE returns the reconstructed image, mu and logvar so that we can optimize KLD in the loss function.  This tuple really trips with the _reorder_ parameter in get_preds.

WandbCallback did not allow to pass in this _reorder_ parameter which makd it impossible to use with models returning tuples.

Also when logging predictions in WandbCallback, the self.valid_dl.show_results call crashes when the model returns more things than the dataloaders expects (in the dataloaders, my target variable is only an image, but the model returns an image, mu and logvar).

It was much easier to just be able to bypass self.valid_dl.show_results by being able to write a subclass of WandbCallback overwriting my new method log_predictions where I can do other things too.  For example in a VAE model, in log_predictions I can sample from a standard normal and ask the model decoder to generate samples and send that picture to wandb for each epoch.


EtienneT:
Humm, seems like the tests didn't pass in a notebook I did not modify.

jph00:
Sorry about that - temporary DNS issue. Re-checking...

PR #2721, opened 2020-08-28
adjustable workers in `get_preds`, set predict to 0 for the other cases

muellerzr:


muellerzr:
Along with this I'm going to write a tutorial.inference documentation to hit on a few key points.

muellerzr:
I've decided since you're going to completely rework inference to not do the inference guide yet. I'd rather wait until after you make your adjustments @jph00

jph00:
This looks great - thanks!

PR #2729, opened 2020-08-30
DistributedDataParallel proper gather: metrics and get_preds(...) work

antorsae:
- DistributedDataParallel consistent metrics and get_preds: added new function to Learner: to_detach, which will attempt to find a .to_detach method in the Learner's self.dl; conversely DistributedDL provides a `.to_detach` method which keeps track of padding in gathered tensors from all ranks and cut padded tensors appropiately so now metrics and inference (get_preds) work OK.


antorsae:
Tests are limited b/c of difficulting in simulating a proper DDP mode in test. I have tested this in train/valid/inference with `WeightedDL` and works OK in my 1-off tests.

jph00:
This looks great. Did you want to close #2687 since this includes that functionality (AFAICT)?

This will take me a while to review - but it's certainly a priority!

jph00:
Oh also, could you try running tests of a regular (non-weighted) DL to confirm it gives the same results for, say, 5 runs of Imagenette over 20 epochs?

antorsae:
5 runs w/ 3 GPUs DDP -> https://gist.github.com/antorsae/7157da6ec9e5fa40a961806a92fc91d7

jph00:
Many thanks! I'll merge this now. But I don't really like the fact that other callbacks have to know to call `self.learn.to_detach`. I wonder if the underlying problem is that really we need a `DistributedLearner`? That seems like it might be the right place for the `gather` stuff - what do you think?

PR #2733, opened 2020-08-31
MCDropoutCallback to enable Monte Carlo Dropout in fastai.

EtienneT:
A new callback to enable Monte Carlo Dropout in fastai in the get_preds method.  Monte Carlo Dropout is simply enabling dropout during inference.  Calling get_preds multiple times and stacking them yield of a distribution of predictions that you can use to evaluate your prediction uncertainty.


tcapelle:
This is pretty elegant en useful, thanks!

PR #2747, opened 2020-09-02
change the order in if statement for code consistency

deep-diver:
in order to make the code more consistent and readable comparing to the next line

`... if valid_pct else ...` and `... if valid_pct is None else ...`do the same thing, but they look different.


jph00:
Unfortunately, it also makes it incorrect, I believe, since valid_pct None and `0` have very different meanings. Your change would make a valid_pct of `0` be ignored, and replaced with a GrandparentSplitter!

PR #2750, opened 2020-09-02
Add examples of DataLoader methods

KushajveerSingh:
In this PR I added simple examples to show how DataLoader methods work (used simple examples in this case).

I can work on a new PR for a detailed from-scratch implementation of DataBlock, showing how these functions are used in that case. I am trying to document the fastai defaults and this can help me with that also.


jph00:
I'm very grateful for you writing this! :) 

It's more of a code walk-thru than an explanation of what things are actually for, so I think this would be better as the basis for a blog post, rather than part of the docs. I do agree that a tutorial explaining this class is needed, but I think that might need to be written by me, since no-one else yet knows why things are the way they are in this class!...

PR #2751, opened 2020-09-02
fix(wandb): remove unused parameter

borisdayma:
Remove unused parameter which causes issue in future release of wandb


borisdayma:
Actually I'm going to take advantage to implement another small improvement on model logging

borisdayma:
The way to log model should be more robust, mainly if the saved model file changes while being uploaded (for example when using `fine_tune` on short epochs).

borisdayma:
@jph00 Not sure if we're still supposed to tag you on PR's

jph00:
I've got things better set up to notify me now. But feel free to tag me if you don't get a timely response

PR #2752, opened 2020-09-03
Corrected Typos in 23_tutorial.vision.ipynb

0tist:


jph00:
I'd rather not replace active with passive voice like this, or to replace simple language with less readable language.

PR #2732, opened 2020-08-31
Documentation updates to 60_medical imaging notebook

asvcode:
This is a proposed update to notebook `60_medical.imaging.ipynb` to improve some elements of the  documentation.

I will also be submitting additional tutorial notebooks on medical imaging and specifically updating notebook 61 which is the medical imaging tutorial.

PR #2754, opened 2020-09-04
Documentation updates to 60_medical imaging corrections

asvcode:
These are the corrections for  PR #2732.  I was unable to successfully commit the corrections to that branch and hence had to open a new one.

Jeremy, all your proposed corrections were done however I had to change this line after `freqhist_bins`:

`plt.hist(t_bin, bins=t_bin, color='c'); plt.plot(t_bin, torch.linspace(0,1,len(t_bin)));dcm.show(t_bin)`

and get rid of `dcm.show(t_bin)` at the end because the `show` and `hist_scaled` functions are listed further down in the notebook.  I  tried to get this to work but it would mean moving those functions further up in the notebook and was not sure if you wanted any movement of the functions.

I do like the fact that you can see the image though because it changes from this:
hist_one

to this:
hist_two


jph00:
Great! :)

PR #2761, opened 2020-09-05
export docs in data.core

KushajveerSingh:
Exported docs of TfmdDL and TfmdLists to the library.


fastai/data/core.py
```diff
@@ -269,6 +279,17 @@ def __getitem__(self, idx):
         if self._after_item is None: return res
         return self._after_item(res) if is_indexer(idx) else res.map(self._after_item)
 
+# Cell
+add_docs(TfmdLists,
+         setup="Transform setup with self",
+         decode="From `Pipeline",
```
jph00, line 285: Looks like two lines are missing closing backticks here

jph00 (changes requested):
Many thanks! Just one change requested

fastai/data/core.py
KushajveerSingh: Done.

KushajveerSingh:
Done.

PR #2774, opened 2020-09-08
Fix typo in docstring for BCEWithLogitsLossFlat

raphaelr:
It's based on nn.BCEWithLogitsLoss, not on nn.CrossEntropyLoss. Updates the classes' docstring to reflect that.


jph00:
Thanks

PR #2778, opened 2020-09-09
scaled_px update

asvcode:
on some occassions `scaled_px` will result in the following error:

```
<ipython-input-9-e3b67e46a2c5> in <module>
----> 1 scaled_px(f)

~\Anaconda3\envs\fastai_2020_N\lib\site-packages\fastai\medical\imaging.py in scaled_px(self)
     65     "`pixels` scaled by `RescaleSlope` and `RescaleIntercept`"
     66     img = self.pixels
---> 67     return img if self.Modality == "CR" else img * self.RescaleSlope + self.RescaleIntercept
     68 
     69 # Cell

~\Anaconda3\envs\fastai_2020_N\lib\site-packages\pydicom\dataset.py in __getattr__(self, name)
    781             return {}
    782         # Try the base class attribute getter (fix for issue 332)
--> 783         return object.__getattribute__(self, name)
    784 
    785     @property

AttributeError: 'FileDataset' object has no attribute 'RescaleSlope'
```

This is because there are other modalities such as from [here](https://www.kaggle.com/c/siim-isic-melanoma-classification/data)  which has a modality of 'XC' and no `RescaleSlope` and `RescaleIntercept` .

The proposed change helps to make this a more universal solution, try to calculate the scaled pixels and if not return `self.pixels` instead.   Example of error in [colab](https://colab.research.google.com/drive/1Plp3w_1mNhNk6PmCHU7uXWyRdpzrye5W?usp=sharing)

In all honesty it's unlikely one would use `scaled_px` on these type of dicoms as they typically have a range between `0` and `255` so it wouldn't work.  This work about is just to stop the error from occurring.


jph00 (changes requested):
We shouldn't use exception handlers that catch and ignore all types of exception. Instead, use `hasattr` on each attr before using it.

asvcode:
Thanks for the pointer, I re-wrote the code using `hasattr` and tested it with the same images and it works fine

PR #2791, opened 2020-09-13
Fixed n_worker to default to 0 for predict.

ClarityCoders:


jph00:
Fix is right, but the change needs to be made in the notebook, not the auto-generated module. :)

PR #2792, opened 2020-09-14
fix fastai/fastbook/issues/263

Tagar:
potential fix for fastai/fastbook#263


Tagar:
Looks like all tests passed except this one - 

Screen Shot 2020-09-13 at 10 58 11 PM

The reason is that I didn't regenerate the callback/schedule.py from notebook, and not sure why that new line is now missing.

Tagar:
I tested locally with fastai installed from this branch (https://github.com/fastai/fastbook/issues/263) 
`pip install git+git://github.com/Tagar/fastai.git`

and it solves the [issue]

Tagar:
All tests pass now too.

jph00:
Many thanks!

Issue #2786, opened 2020-09-11
show_batch() function does not show BBoxes correctly.


IRailean:
Please confirm you have the latest versions of fastai, fastcore, fastscript, and nbdev prior to reporting a bug (delete one): YES 

**Describe the bug**
Try to run https://docs.fast.ai/tutorial.datablock#Bounding-boxes tutorial with different parameter sent in Resize function (other than 128). BBoxes with wrong size and location are shown.

**To Reproduce**
1.  Install fastai from git using command:
     !pip install git+https://github.com/fastai/fastai.git
2. Import fastai
```
   from fastai.data.all import *
   from fastai.vision.all import *
```
3. Run the following code from the [tutorial](https://docs.fast.ai/tutorial.datablock#Bounding-boxes), setting the parameter in Resize function other that 128. (128 is the size of a picture in the coco_tiny dataset mentioned in the tutorial)
```
   
   coco_source = untar_data(URLs.COCO_TINY)
   images, lbl_bbox = get_annotations(coco_source/'train.json')
   img2bbox = dict(zip(images, lbl_bbox))
   
   coco = DataBlock(blocks=(ImageBlock, BBoxBlock, BBoxLblBlock),
                 get_items=get_image_files,
                 splitter=RandomSplitter(),
                 get_y=[lambda o: img2bbox[o.name][0], lambda o: img2bbox[o.name][1]], 
                 item_tfms=Resize(256),
                 batch_tfms=aug_transforms(),
                 n_inp=1)

   dls = coco.dataloaders(coco_source)
   dls.show_batch(max_n=9)

```
4. Look at the output of the show_batch() function.

**Expected behavior**
BBoxes shown correctly.

**Additional context**
This may happen because of applying PointScaler, Resize transforms incorrectly to TensorBBox.

marii-moe:
Adding my gist to this one: https://gist.github.com/marii-moe/be625c65e6d8aa783370cc70c7bc7ae2

I am continuing to look into this one, but I am fairly stuck on it.

IRailean:
Take a look at _unscale_pnts function in vision/core.py The problem may be that for unscaling it uses the original size of a picture not the resized one. 
I hardcoded sz variable in _unscale_pnts to a resized picture size: 
sz = fastuple((700, 700)) (In DataBlock API I used Resize(700, method=ResizeMethod.Squish)
and it worked! (Working also with other ResizeMethods.

So this may be the issue.

marii-moe:
[...]

PR #2799, opened 2020-09-15
`img_size` attribute for `TensorPoint` is not updated properly

IRailean:
Solving the issue reported in #2786 
As resized image size will be saved in PointScaler self.sz variable, and 'img_size' attribute for TensorPoint is not updated properly -> prioritize self.sz over sz in PointScaler _get_sz function.


jph00:
Could you please add a test in the notebook that previously would have failed, but this change makes pass?

jph00 (changes requested):
Could you please add a test in the notebook that previously would have failed, but this change makes pass?

IRailean:
I have updated the last test from 09_vision.augment.ipynb with a new transform Resize(256) - it is clearly seen that previously bboxes are not covering the objects. 
Note: With the old implementation everything works fine if run only this notebook. If run after installing fastai (pip | from git | using local setup.py), bboxes are not shown correctly.

jph00:
> With the old implementation everything works fine if run only this notebook. If run after installing fastai (pip | from git | using local setup.py), bboxes are not shown correctly.

Oh interesting - can you tell me more about this? Is it a fastcore change or fastai change which broke things? Do you know what version the breakage occurred from?

marii-moe:
I think I was the one who originally stated that a previous version existed that did not have the issue. I tried to track down the issue, but didn't get to the bottom of it yet. This notebook shows the behavior I found that was suspicious: https://gist.github.com/marii-moe/a8e7a330fbd4d9b0dc326e396ef17e36

Note, in the above notebook the gist is from a recent version of fastai, I copied and pasted the results from the previous version for a comparison. 

commit that this is from is 94e1568f9d9f7718a8c925436fc4b25452f626b9

jph00:
Thanks to you both! I'll just go ahead and merge this then, but I don't actually know if it's the right fix, so I'd love it if you folks could take a look at some image augmentation outputs and confirm visually that it all looks ok to you! :)

muellerzr:
Can confirm this is working. Also fixes any issues with `aug_transforms` along with it. Nice work guys :)

Issue #2738, opened 2020-09-01
Typo in replace_maj documentation


kafji:
**Describe the bug**
Typo in replace_maj documentation, `Replace tokens in ALL CAPS by their lower version and add TK_UP before.`

**To Reproduce**
See rendered documentation https://docs.fast.ai/text.core#replace_maj, source code for the documentation/notebook https://github.com/fastai/fastai/blob/master/nbs/30_text.core.ipynb?short_path=b6012f0#L350, and source code for the function itself https://github.com/fastai/fastai/blob/13bf45874516a78068c671d0f1e0618a888457ed/fastai/text/core.py#L84

**Expected behavior**
Function documentation should be `Replace capitalized tokens by their lower version and add TK_MAJ before.`

PR #2803, opened 2020-09-15
#2738 typo `TK_MAJ` instead of `TK_UP`

tyoc213:
Just a typo reported in #2738


jph00 (changes requested):
Thanks! Actually there's a 2nd problem here. Could you replace "Replace tokens in ALL CAPS" with "Replace tokens in Sentence Case" in the docstring?

tyoc213:
@jph00 I wonder this 2 are phrased exactly the same... both of them should be "Replace tokens in Sentence Case"?

image

And

image

jph00:
No, `replace_all_caps` is correct. It's only `replace_maj` that needs to be changed.

PR #2810, opened 2020-09-17
Separate out the loss functions inside layer.py into losses.py, verbose the documentation some, and update the sidebar 

muellerzr:


muellerzr:
Don't merge yet, I know it's failing on the imports from layers. Let me do some digging/refactoring some

muellerzr:
Needed to include it inside of `torch_basics`

PR #2816, opened 2020-09-18
Clamp the outputs for points similar to bounding boxes and expand the related docs

muellerzr:
Adds a clamping mechanism to `PointBlock` and for `TensorPoints`, preventing points to go out-of-scope at the end of data augmentation, allowing for more intense augmentations to be utilized. Along with this the `bb_pad` and other similar functionalities documentations were updated to include examples from a high and low-level integration 

Oh, and of course, tests!


fastai/vision/data.py
```diff
@@ -20,6 +20,12 @@ def get_grid(n, nrows=None, ncols=None, add_vert=0, figsize=None, double=False,
     if title is not None: fig.suptitle(title, weight='bold', size=14)
     return (fig,axs) if return_fig else axs
 
+# Cell
+def clamp_points(samples):
+    "Clamps points to an image border"
+    def _f(imgs, pnts): return imgs,torch.clamp(pnts, -1, 1)
+    return [_f(*s) for s in samples]
```
jph00, line 27: I don't see the benefit of defining `_f`? Why not just put that inline into the list comprehension? Or alternatively keep `_f`, and use `samples.starmap(_f)`

jph00 (changes requested):
Thanks. One minor suggestion.

How well tested is this? It's a fairly significant change so can you ensure you've tried a few different datasets, transforms, models, etc, to see it works well?

fastai/vision/data.py
muellerzr: I was mostly looking at the bounding box code for that one, so I can look into modifying it. I can also then potentially do the same for the bbox code too.

muellerzr:
It's extremely well tested (by others doing keypoint datasets), and provides a much more stable training. Along with this, in any article I've read involving data augmentation with points, this is being done, as well as in papers. What most libraries do when they instead opt for a conv-approach (heatmap) is they have a third variable they keep track of that checks if a point went off-screen (it's called vis), however this is a major API change to how keypoints work, so simply doing this approach allows for more stable training while enabling transforms to be done. 

I had to modify my Walk with fastai notebook to include a ClampBatch transform because it was leading to instability during training thanks to out-of-boundary transforms. As you may recall many moons ago we had a discussion over trying to figure out why the results of my cat keypoint model was centering rather than giving the points, and it was due to them going off-screen. While this was initially a data issue (as some of the datapoints were already going off-picture), after applying the common transforms the pattern came back. Clamping fixed this issue. 

Otherwise, without clamping you can't really remove outliers well, and your model will train for "ghost points", which during inference can be hard to weed out and narrow down (IE the model winds up guessing where an arm joint could be, even though that's not an option or what it should be looking for). Instead we look for extremes on the boundaries.

jph00:
Sorry to clarify, I'm not asking about the general approach, but the testing of this specific code.

muellerzr:
Ahhh sorry, I understand now. I'm updating the Walk with fastai keypoint notebook with it, but I did my testing with the Kaggle Cats dataset at first to provide hints towards generating a decent test. IE the current test here checks for combinations of out of scope negatives and positives (`-2,.5`), etc, which I can all easily reproduce by adding transforms like `aug_transforms` to my `DataBlock`. Does this help some? I can also run this with the headpose dataset too if that would be beneficial.

fastai/vision/data.py
muellerzr: I don't believe they come in as an `L` though, so that could throw an issue? Doing the following:

```python
def print_types(s): print(type(s))
dls= dblock.dataloaders(path/'images', bs=64, before_batch=print_types)
```
Shows that it winds up coming in as a `list`, unless you are suggesting something like:
```python
return L(samples).starmap(_f)
```

muellerzr:
I see now. While I did test it, I didn't test it thoroughly, as it winds up not really doing what it should be doing on the points (outliers still are > 1.) I'm going to attempt to restructure this as a Transform instead that runs before IntToFloat and Normalize

muellerzr:
The fact that this did not do as it was supposed to also causes some concern over doing the same thing in practice with the bounding boxes. Which is indeed the case, leading to a larger issue. Tested the following with coco:

```python
item_tfms = [Resize(64),]
batch_tfms = [*aug_transforms(), Normalize.from_stats(*imagenet_stats)]
```
Grabbing a batch of data leads to an out of scope y:
```python
[-0.7330, -1.0998, -0.0408, -0.5673],
```
My solution for the keypoints was to do a `Transform` that instead performs this as an order of `50`:
```python
class ClampPoints(Transform):
    order = 50
    def encodes(self, x:TensorPoint): return torch.clamp(x, -1,1)
```
Which would run this after `Normalize`, which I believe should be the solution here too. As any batch transforms looks to mess with it. Does this sound reasonable @jph00 ?

jph00:
Sounds reasonable, although I don't recall all the code fully ATM. So still needs plenty of testing!

jph00:
> Shows that it winds up coming in as a `list`, unless you are suggesting something like:

> return L(samples).starmap(_f)

Yes that.

jph00:
> IE the current test here checks for combinations of out of scope negatives and positives (`-2,.5`), etc, which I can all easily reproduce by adding transforms like `aug_transforms` to my `DataBlock`. Does this help some? I can also run this with the headpose dataset too if that would be beneficial.


Yes this certainly helps some, although just as important is to train some models and do inference to see it looks right.

muellerzr:
Closing the PR for now, as there's more hidden away with issues pertaining to the problem:

Such as [here](https://discuss.pytorch.org/t/need-some-help-on-using-the-new-keypointrcnn/47051)

PR #2818, opened 2020-09-18
nb60_changes

asvcode:
Realized a few inconsistencies since my last PR hence correcting them.  Most predominantly visual. 

- histogram for when `n_bins` is 1 is now correctly showing the right histogram
- removing the `1 frame per file` when using `show` which is not really required if there is only 1 image per frame
- `gauss_blur2d` example now correctly shown


jph00:
Thanks!

PR #2822, opened 2020-09-21
added tensorboard embedding projector support for images

floleuerer:
Added tensor board embedding projector support for images. The Callbacks (TensorBoardCallback(projector=True) and TensorBoardProjectorCallback()) register a hook at model[1][1] to extract the features. Those features will be written as embeddings, and images / labels will be added. 

TensorBoardCallback(projector=True) writes the embeddings for the validation set during training after each epoch.
TensorBoardProjectorCallback() can be used to write the embeddings for a custom dataset (test_dl).


fastai/callback/tensorboard.py
```diff
@@ -13,8 +13,9 @@
 # Cell
 class TensorBoardCallback(Callback):
     "Saves model topology, losses & metrics"
-    def __init__(self, log_dir=None, trace_model=True, log_preds=True, n_preds=9):
-        store_attr('log_dir,trace_model,log_preds,n_preds')
+    def __init__(self, log_dir=None, trace_model=True, log_preds=True, n_preds=9, projector=False, layer=None):
+        store_attr('log_dir,trace_model,log_preds,n_preds,projector,layer')
```
jph00, line 17: Since you're storing everything, you can remove the argument entirely and just say `store_attr()`

fastai/callback/tensorboard.py
```diff
@@ -42,8 +43,98 @@ def after_epoch(self):
             x,y,its,outs = self.dls.valid.show_results(b, out, show=False, max_n=self.n_preds)
             tensorboard_log(x, y, its, outs, self.writer, self.train_iter)
 
+    def before_validate(self):
+        if self.projector:
```
jph00, line 47: Let's reduce nesting by making `if not self.projector: return` the first line of each of these methods

fastai/callback/tensorboard.py
```diff
@@ -42,8 +43,98 @@ def after_epoch(self):
...
+            self.run_projector = True
+
+            self.h = hook_output(self.learn.model[1][1] if not self.layer else self.layer)
+            self.feat = {}
+
+    def after_pred(self):
+        if self.run_projector:
```
jph00, line 54: Let's reduce vert space by putting next line with this one

fastai/callback/tensorboard.py
```diff
@@ -42,8 +43,98 @@ def after_epoch(self):
...
+    def _remove(self):
+        if getattr(self, 'h', None): self.h.remove()
+
+    def __del__(self): self._remove()
+
+# Cell
+class TensorBoardProjectorCallback(Callback):
+    "Saves Embeddings for Tensorboard Projector"
+    def __init__(self, log_dir=None, layer=None):
+        store_attr('log_dir,layer')
```
jph00, line 74: `store_attr()`

fastai/callback/tensorboard.py
```diff
@@ -42,8 +43,98 @@ def after_epoch(self):
...
+
     def after_fit(self): self.writer.close()
 
+    def _remove(self):
+        if getattr(self, 'h', None): self.h.remove()
+
+    def __del__(self): self._remove()
+
+# Cell
+class TensorBoardProjectorCallback(Callback):
```
jph00, line 71: Lots of dupe code with previous class - suggest you have them both inherit from one base class

fastai/callback/tensorboard.py
```diff
@@ -42,8 +43,98 @@ def after_epoch(self):
...
+    def _remove(self):
+        if getattr(self, 'h', None): self.h.remove()
+
+    def __del__(self): self._remove()
+
+# Cell
+def write_projector_embedding(learn, writer, feat):
+    lbls = [learn.dl.vocab[l] for l in feat['lbl']] if getattr(learn.dl, 'vocab', None) else None
+
+    writer.add_embedding(feat['vec'],
```
jph00, line 107: Prefer on one line

fastai/callback/tensorboard.py
```diff
@@ -42,8 +43,98 @@ def after_epoch(self):
...
+
+            write_projector_embedding(self.learn, self.writer, self.feat)
+
+
+    def _remove(self):
+        if getattr(self, 'h', None): self.h.remove()
+
+    def __del__(self): self._remove()
+
+# Cell
```
jph00, line 103: Do this and the next one need to be public? If not, rename them to start with `_`

jph00 (changes requested):
Very nice! Some minor cleanup suggestions

floleuerer:
I've created a BaseCallback and inherited from that. Made the code clearer. Also tried to implement all your suggestions.

jph00:
Great! Thanks so much :)

PR #2828, opened 2020-09-22
Updating test to be more readable, and adding doc description

marii-moe:
Changed the previous test to hopefully be a little bit more readable. I have convinced myself that it is more readable at this point, but would definitely like another opinion. 

For some reason the images were changed as well, which was unintentional. Can correct if there is a way to do that.


jph00:
Much better! :D

PR #2834, opened 2020-09-25
GatherPredsCallback can now have save_preds and save_targs set to false

EtienneT:
GatherPredsCallback can now have save_preds and save_targs set to False to prevent accumulating them in memory and also not saving them on disk.  This can prevent out of memory exception with large output sizes (for example Autoencoder with large tabular data where some categorical variables have a lot of possible values).  The default value of None has the same behavior as before, we accumulate only in memory.

Also added support saving non tensor output using pickle (useful if the model output tuples or dict).  By default it uses save_array like before, but if the data we are saving is not a numpy array, then revert to pickle.


fastai/callback/core.py
```diff
@@ -86,7 +86,10 @@ def before_validate(self):
 if not hasattr(defaults, 'callbacks'): defaults.callbacks = [TrainEvalCallback]
 
 # Cell
-#TODO: save_targs and save_preds only handle preds/targets that have one tensor, not tuples of tensors.
+try: import pickle
```
jph00, line 89: No need for try here. All pythons we run on have pickle

fastai/callback/core.py
```diff
@@ -106,14 +109,21 @@ def after_batch(self):
...
-        else: (self.save_targs/str(self.iter)).save_array(targs[0])
+        elif self.save_targs is not False: self.save_data((self.save_targs/str(self.iter)), targs[0])
         if self.with_loss:
             bs = find_bs(self.yb)
             loss = self.loss if self.loss.numel() == bs else self.loss.view(bs,-1).mean(1)
             self.losses.append(self.learn.to_detach(loss))
 
+    def save_data(self, path, data):
+        if isinstance(data, (np.ndarray, np.generic)):
+            path.save_array(data)
```
jph00, line 122: combine with previous line

fastai/callback/core.py
```diff
@@ -106,14 +109,21 @@ def after_batch(self):
...
+        else:
+            with open(path,'wb') as file:
```
jph00, line 124: combine with previous line

fastai/callback/core.py
```diff
@@ -106,14 +109,21 @@ def after_batch(self):
         if not hasattr(self, 'pred'): return
         preds,targs = self.learn.to_detach(self.pred),self.learn.to_detach(self.yb)
         if self.save_preds is None: self.preds.append(preds)
-        else: (self.save_preds/str(self.iter)).save_array(preds)
+        elif self.save_preds is not False: self.save_data((self.save_preds/str(self.iter)), preds)
         if self.save_targs is None: self.targets.append(targs)
-        else: (self.save_targs/str(self.iter)).save_array(targs[0])
+        elif self.save_targs is not False: self.save_data((self.save_targs/str(self.iter)), targs[0])
```
jph00, line 114: `is not` should be `!=` shouldn't it?

jph00 (changes requested):
Can you please explain more about what the purpose of this is? Maybe best to add some examples, explanation, and reasoning in the notebook, along with some tests.

EtienneT:
I fixed all styling, sorry about that.

My main use case for this change is when using a large tabular dataset in memory (2gb in memory).  The model is a tabular variational autoencoder where I try to reconstruct both continuous and categorical variables.  I want to use the compressed representation just like the [Porto Seguro Winning Solution](https://forums.fast.ai/t/porto-seguro-winning-solution-representation-learning/8499).  Some of the categorial variables in my dataset have large cardinality, so the output of the model is quite large since I have several categorical variables all outputting softmaxed one hot encoded vectors, plus all the continuous variables.

I don't care about the reconstructed values, I only care about the compressed representation.  So I use a HookCallback to get those in get_preds:

```
class GetBottleNeckCallback(HookCallback):
    def __init__(self, path, **kwargs):
        super().__init__(**kwargs)
        self.path = path
        
    def hook(self, m, i, o):
        (self.path/str(self.iter)).save_array(self.learn.to_detach(i[0]))

hook = GetBottleNeckCallback(path=Path('saved/compressed'), modules=[learn.model.decoder[0][0]])
learn.add_cbs(hook)
dl = learn.dls.test_dl(df)
learn.get_preds(dl=dl, save_preds=False, save_targs=False)
```

I don't care about the prediction and the targets because I already have them in my dataframe.  I just want the data produced by the GetBottleNeckCallback.

This is the main reason for this change.  A variational auto-encoder returns a tuple, the reconstructed input, the mu and logvar of the distribution.  So I tried saving that with pickle before, but it resulted in 300+megs files for each batch.  If I let the prediction accumulate in memory I get an out of memory on my 64gb ram in about 10s.

So that's why I added the option not to save anything both in memory and on disk for scenarios where a callback is doing the job in the background.  I hope this clarifies!

Thanks,

EtienneT:
Having talked with @muellerzr about the scenario where this was relevant for the project I was working on, I think this pull request might be irrelevant.  Closing for now!

PR #2835, opened 2020-09-25
Add try/catch to MultiCategorize and Categorize

muellerzr:
Adds a verbose catch when there is an attempt to encode data not present in the training dataset's vocab.


jph00:
Thanks!

Issue #2827, opened 2020-09-22
manually supply categories dict to Tabular


DmitriyG228:
Feature requests should first be proposed on [the forum](https://forums.fast.ai). 

**Link to forum discussion.**

https://forums.fast.ai/t/tabular-categorical-classes-for-out-of-memory-datasets/79366

**Is your feature request related to a problem? Please describe.**

I am training with tabular dataset that is larger than RAM, so I reload the DataLoaders on n-th epoch.
As a matter of fact, some categories vary from one load to an other in my case, so that would be helpful to manually supply with my categories dict. As a workaround, one has to initially create a setup dataset with all the unique categories.

**Describe the solution you'd like**

Be able to manually supply with my categories dict.

**Describe alternatives you've considered**

As a workaround, one has to initially create a setup dataset with all the unique categories.

marii-moe:
Seems this was fixed in https://github.com/fastai/fastai/pull/2837

PR #2837, opened 2020-09-26
Add ability to pass in pre-determined stats to FillMissing, Categorify, and Normalize, and allow Normalize to pass kwargs

muellerzr:
Fulfills https://github.com/fastai/fastai/issues/2827 while also doing the same for the rest of the `procs`. Normalize also was given the ability to store `kwargs`, this is needed due to the slight name changing (`mean` to `means`, `std` to `stds`)


fastai/data/transforms.py
```diff
@@ -345,10 +345,10 @@ def broadcast_vec(dim, ndim, *t, cuda=True):
 class Normalize(DisplayedTransform):
     "Normalize/denorm batch of `TensorImage`"
     parameters,order = L('mean', 'std'),99
-    def __init__(self, mean=None, std=None, axes=(0,2,3)): store_attr()
+    def __init__(self, mean=None, std=None, axes=(0,2,3), **kwargs): store_attr(**kwargs)
```
jph00, line 348: What are the kwargs for? Can we list them explicitly, or use `delegates`?

fastai/tabular/core.py
```diff
@@ -237,8 +237,18 @@ def _decode_cats(voc, c): return c.map(dict(enumerate(voc[c.name].items)))
...
     order = 1
+    def __init__(self, classes=None):
+        store_attr()
+        super().__init__()
     def setups(self, to):
-        store_attr(classes={n:CategoryMap(to.iloc[:,n].items, add_na=(n in to.cat_names)) for n in to.cat_names})
+        if self.classes is None:
+            classes={n:CategoryMap(to.iloc[:,n].items, add_na=(n in to.cat_names)) for n in to.cat_names}
+        else:
+            classes = {}
```
jph00, line 247: Would `defaultdict` allow us to simplify this?

fastai/tabular/data.py
```diff
@@ -27,7 +27,7 @@ def from_csv(cls, csv, skipinitialspace=True, **kwargs):
         return cls.from_df(pd.read_csv(csv, skipinitialspace=skipinitialspace), **kwargs)
 
     @delegates(TabDataLoader.__init__)
-    def test_dl(self, test_items, rm_type_tfms=None, process=True, **kwargs):
```
jph00, line 30: Why is this removed?

jph00 (changes requested):
Thanks! Just some little questions to look at

fastai/data/transforms.py
muellerzr: `kwargs` are for us to be able to pass `means` and `stds` over to `Normalize` explicitly. I couldn't figure out a better way to do this so far, so let me know if you have ideas. Can we use `delegates`? 

Relevant code is `Normalize.from_tab`

fastai/tabular/data.py
muellerzr: We don't ever pass in `rm_type_tfms` for anything and it doesn't serve a purpose for tabular. (Unless you can think of a scenario we can keep it).

jph00:
Just write out the param names and defaults explicitly, I guess

muellerzr:
So you're saying we have `Normalize` be:

```python
def __init__(self, mean=None, std=None, axes=(0,2,3), means=None, stds=None)
```
? 

Because I can explicitly say it in `Normalize.from_tab`, but just want to make sure we want it explicit here as well.

muellerzr:
If we did go that route I would just get rid of `Normalize.from_tab` entirely most likely since it's just redundancy

jph00:
On Mon, Sep 28, 2020, at 9:45 AM, Zachary Mueller wrote:
> If we did go that route I would just get rid of `Normalize.from_tab` entirely most likely since it's just redundancy


Sounds like that would simplify the API, so that seems like an improvement?

jph00:
> We don't ever pass in `rm_type_tfms` for anything and it doesn't serve a purpose for tabular. (Unless you can think of a scenario we can keep it).


OK thanks. In the future, please try to keep separate fixes in separate PRs, with a description of what each is doing. Makes life much easier! :)

muellerzr:
> OK thanks. In the future, please try to keep separate fixes in separate PRs, with a description of what each is doing. Makes life much easier! :)

Can do! Tried to do this for this time around (I have 3-ish tabular PR's including this one) but will keep this in mind better

muellerzr:
@jph00 do you see something inheritably wrong with:
```python
class Categorify(TabularProc):
    "Transform the categorical variables to something similar to `pd.Categorical`"
    order = 1
    def __init__(self, classes=defaultdict(L)): 
        store_attr()
        super().__init__()
    def setups(self, to):
        for n in to.cat_names:
            if n not in self.classes:
                self.classes[n] = CategoryMap(to.iloc[:,n].items, add_na=n)
```

My belief/understanding is this does the same thing however it fails on the fourth test:
```python
AssertionError: ==:
(#4) ['#na#',0,1,2]
['#na#', 'H', 'M', 'L']
```

fastai/tabular/core.py
muellerzr: It will and it won't. There's some weird behavior with if it's a categorical datatype in the `DataFrame` (see next push), but considering we did it always anyways I went ahead and did a check for it to recalculate

muellerzr:
So I can't quite seem to get the `defaultdict` working how it should, my attempt was:

```python
class Categorify(TabularProc):
    "Transform the categorical variables to something similar to `pd.Categorical`"
    order = 1
    def __init__(self, classes=defaultdict(L)): 
        store_attr()
        super().__init__()
    def setups(self, to):
        for n in to.cat_names:
            if n not in self.classes or is_categorical_dtype(to[n]):
                self.classes[n] = CategoryMap(to.iloc[:,n].items, add_na=n)

    def encodes(self, to): to.transform(to.cat_names, partial(_apply_cats, self.classes, 1))
    def decodes(self, to): to.transform(to.cat_names, partial(_decode_cats, self.classes))
    def __getitem__(self,k): return self.classes[k]
```

Do you see anything wrong with this implementation I may have missed? The issue is it passes all the base tests but not the TabularPandas `Pipeline` tests

fastai/tabular/core.py
muellerzr: Wound up not working as it should the further we got in the pipeline (see recent comment)

WillowRainbowe:
I have unsubscribed from this account 3 times, and yet I still continue to
keep receiving emails from many many people in this unforsaken group. I
want ALL OF THESE DARN EMAILS TO STOP NOW!!!!! I don't want to hear any
excuses either. I believe I have been extremely patient, considering all
the aggravating multitude of emails that I have had to delete every single
day.
If this is not taken care of, I will be forced to take more drastic
measures that you and your group/company will really not appreciate. So
please take me off your list and have EVERYONE STOP sending me anymore
emails.

Thank you,

ntr.lvin.wiccan.witch@gmail.com

muellerzr:
@jph00 I think that last commit should have things working how you'd like them with the `defaultdict`, let me know if otherwise.

hamelsmu:
@muellerzr can you resolve these conflicts?

PR #2840, opened 2020-09-28
fix _get_sz_change_idxs

### LINKED ISSUE #2820: _get_sz_change_idxs returns an extra element, 0
**Describe the bug**
The function _get_sz_change_idxs returns an extra 0 if `sizes` changes between index 0 and 1.

**To Reproduce**
1. run
```python
from fastai.vision.models.unet import _get_sz_change_idxs
_get_sz_change_idxs([[3,128,128], [16,64,64], [32,32,32], [16,32,32], [32,32,32], [16,16]])
```
2. It should return [0,1,4]. Instead, it returns [0,0,1,4]

**Expected behavior**
It should pass the following test
```python
test_eq(_get_sz_change_idxs([[3,128,128], [16,64,64], [32,32,32], [16,32,32], [32,32,32], [16,16]]), [0,1,4])
```

**Additional context**

The problem seems to be that the following line in the function definition is not needed:
```python
if feature_szs[0] != feature_szs[1]: sz_chg_idxs = [0] + sz_chg_idxs
```
[This](https://github.com/fastai/fastai/blob/master/nbs/15a_vision.models.unet.ipynb) is the nb where the function is.

This behavior makes the DynamicUnet to use F.interpolate unnecessarily.

If you agree, I can make a PR with this line deleted and with the extra test.

ababino:
Fixes _get_sz_change_idxs function returning an extra 0. Fixes #2820

### DIFF
--- a/fastai/vision/models/unet.py
+++ b/fastai/vision/models/unet.py
@@ -11,7 +11,6 @@ def _get_sz_change_idxs(sizes):
     "Get the indexes of the layers where the size of the activation changes."
     feature_szs = [size[-1] for size in sizes]
     sz_chg_idxs = list(np.where(np.array(feature_szs[:-1]) != np.array(feature_szs[1:]))[0])
-    if feature_szs[0] != feature_szs[1]: sz_chg_idxs = [0] + sz_chg_idxs
     return sz_chg_idxs
 
 # Cell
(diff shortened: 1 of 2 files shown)


jph00:
Many thanks

PR #2841, opened 2020-09-29
Enable TrackerCallback to maintain best over subsequent calls to fit

ohmeow:
Added reset_on_fit paramter to TrackerCallback's constructor.  If set = False, the callback will maintain the "best" value over subsequent calls to "fit."  If set = False (default), the current behavior whereby "best" is reset upon each call to "fit" in kept.


jph00 (changes requested):
This is actually rather neat :) . If you add some docs to explain it, and a little test, I'll merge it! Thanks @ohmeow

Issue #2785, opened 2020-09-10
Passing custom function or array to Rotate in transforms


ecatkins:
According to the docs, you can pass `Rotate` a `callable` that returns a float to customize the angle applied. As i understand this...
```
def my_rotate(x):
    return 10.5

tfm_list = [Rotate(p=1.0, draw=my_rotate)]
my_tfms = setup_aug_tfms(tfm_list)
```

However when I call `show_batch` it returns the following error
![image](https://user-images.githubusercontent.com/8319226/92762814-f6212a00-f360-11ea-993f-ecb1b7316d89.png)

Using the standard `Rotate` without specifying `draw` e.g. ` Rotate(max_deg=10, p=0.5)` works fine.

To experiment, I also tried to pass an array of length `BATCH_SIZE` (16 in this case) to `draw`, which is also specified in the docs and I get a different error
```
thetas = [10.0] * BATCH_SIZE
tfm_list = [Rotate(p=1.0, draw=thetas]
my_tfms = setup_aug_tfms(tfm_list)
```
![image](https://user-images.githubusercontent.com/8319226/92764707-a4799f00-f362-11ea-8ec4-fd19d580191b.png)

The length of my `thetas` is not equal to `1` even though the docs say it must be equal to the batch size.

Strangely if I try to purposely misspecify the array so that it is less than the `BATCH_SIZE`...
```
thetas = [10.0] * (BATCH_SIZE-1)
```
Now the assertion is what I would have expected it to be `len(thetas)==len(BATCH_SIZE)`

![image](https://user-images.githubusercontent.com/8319226/92764487-772cf100-f362-11ea-8872-50b5c2c36614.png)

marii-moe:
I am now actively working on this issue.

marii-moe:
After looking into this a bit, an array will not work for batch_tfms, which I believe is where this code is being used, and where I was able to reproduce the issues listed here. I do think this points out that the documentation here needs to be updated. I will probably try to do that. 

Here are some examples of how this works with batch transforms. 
```
def my_rotate_different(x): 
    return torch.arange(0,x.size(0)*15,15,device=x.device)

tfm_list = [Rotate(p=1.0, draw=my_rotate_different)]
my_tfms_different = setup_aug_tfms(tfm_list)

tfm_list = [Rotate(p=1.0, draw=10.5)]
my_tfms_constant = setup_aug_tfms(tfm_list)

#Example 1
set_seed(42,True)
[...]

PR #2845, opened 2020-09-30
Added ability to have variable length draw

marii-moe:
Further tests for all identified Transforms can be found here. If we want all of these added I can add these: https://gist.github.com/marii-moe/07412da09f32054dca32373d26aa631e

Documentation of new and existing functionality is still being worked on. Wanted to know if you liked this approach for #2785 . If so I will document the functionality here as well as that which is not very clear as identified by this issue as well. Such as how to use draw as a function, as that actually seems to require understanding tensors. 

IF this behavior is not desired, I can also simply implement it so that only batch_size and 1 are legitimate values for the len(draw). I just thought is was a free win to allow variable length here. Batch size 1 is required because of TfmdDL._one_pass,similar to one_batch with bs=1, being used in decode to retain types.

### DIFF
--- a/fastai/vision/augment.py
+++ b/fastai/vision/augment.py
@@ -458,7 +458,7 @@ def _draw_mask(x, def_draw, draw=None, p=0.5, neutral=0., batch=False):
     if draw is None: draw=def_draw
     if callable(draw): res=draw(x)
     elif is_listy(draw):
-        test(len(draw),x.size(0),lambda a,b:a>=b)
+        test(len(draw),x.size(0), ge)
         res = tensor(draw[:x.size(0)], dtype=x.dtype, device=x.device)
     else: res = x.new_zeros(x.size(0)) + draw
     return mask_tensor(res, p=p, neutral=neutral, batch=batch)
--- a/fastai/vision/augment.py
+++ b/fastai/vision/augment.py
@@ -458,8 +458,8 @@ def _draw_mask(x, def_draw, draw=None, p=0.5, neutral=0., batch=False):
     if draw is None: draw=def_draw
     if callable(draw): res=draw(x)
     elif is_listy(draw):
-        test_eq(len(draw), x.size(0))
-        res = tensor(draw, dtype=x.dtype, device=x.device)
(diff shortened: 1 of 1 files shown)


jph00 (commented):
I'm afraid I'm not following the logic of this patch. Can you explain exactly what's going on here? It's been a while since I looked at this code!

marii-moe:
@jph00 
I'll reuse actual test cases if that is okay. All the above assumes p=1 which is the probability that the transform is applied. 
1.
`
test_eq(_draw_mask(x, def_draw, 1, p=1), tensor([1.,1,1,1,1]))`
Since the above's third argument is 1, 1 is the "draw" 1 is the mask for every element of the batch. In Rotate this could be set to 3030, and would mean every image was rotated 30 degrees. 
2.
`test_eq(_draw_mask(x, def_draw, [0,1,2,3,4], p=1), tensor([0.,1,2,3,4]))`
This provides a list of values for "draw", so each of these are applied to an element of the batch. This patch is trying to allow this already existing functionality to be used in a dataloader. If we take the rotate example this would be the same as providing [30,90,120,150,180] degree rotations. 
3.
test_eq(_draw_mask(x[0:3], def_draw, [0,1,2,3,4], p=1), tensor([0.,1,2]))
This is new functionality I am proposing, where you can have a variable batch size as long as it is less than the len(draw).  In my rotate example this would be like providing [30,90,120,150,180], but it would only use the first three if your batch_size=3. 

This functionality in 3 is not needed, but support for bs=1 has to exist for dls._one_pass to work, and therefore decoding of a batch in the dataloader. Alternative way to allow functionality in 2 is to instead have a special exception for batch size of 1, which I think would also be fine. I was planning on documenting the behavior in 3 if you thought it was okay, but if it seems more confusing than useful I can instead add a special exception for batch size of 1.

jph00:
Awesome! Thanks so much for the great explanation :) If you have a chance to document this in the nb, that would be really helpful.

marii-moe:
@jph00 Already working on it!

PR #2849, opened 2020-10-01
add show_install method

tcapelle:
bring back show_install functionalities from V1 to V2


tcapelle:
Hey Jeremy, I am not sur if you want this methods to be on the test_util file, but I did not wanted to create a new file without your blessing.
Maybe a more general `util.py` file and add all the things that are on the `imports,py` and `torch_core.py` that should be elsewhere?

nbs/97_test_utils.ipynb
```diff
@@ -1,23 +1,13 @@
 {
  "cells": [
-  {
```
jph00, line 3: Careful not to delete cells like this in your PRs! :) I'll add it back after I merge

tcapelle: why it happen? 
I never touched the ipynb's without jupyter...

jph00: It means you deleted the cell in Jupyter.

Issue #2781, opened 2020-09-10
SiameseTransform uses validation items during training


ucohen:
**Describe the bug**
in `SiameseTransform` (_24_tutorial.siamese.ipynb_), 
the `_draw()` method picks items from all files and not just the proper split.

**To Reproduce**
This is evident by testing that `tfm.valid` has items from the train split:
```
valids = [v[0] for k,v in tfm.valid.items()]      
assert [v for v in valids if v in files[splits[0]]] == 0
```

**Expected behavior**
Train files should only draw samples from train split, valid from valid split.

My suggestion is to setup both dictionaries for the train and valid splits during `__init__`:
```
class SiameseTransform(Transform):
    def __init__(self, files, splits):
        self.lbl2files = [{l: [f for f in files[splits[i]] if label_func(f) == l] for l in labels}
                          for i in range(2)]
        self.valid = {f: self._draw(f,1) for f in files[splits[1]]}
    def encodes(self, f):
        f2,same = self.valid.get(f, self._draw(f,0))
        img1,img2 = PILImage.create(f),PILImage.create(f2)
        return SiameseImage(img1, img2, same)
    
    def _draw(self, f, split=0):
        same = random.random() < 0.5
        cls = label_func(f)
        if not same: cls = random.choice(L(l for l in labels if l != cls)) 
        return random.choice(self.lbl2files[split][cls]),same
```

Place between these lines with triple backticks:

```
```

**Additional context**

yijinlee:
I think you are right – well spotted. There seem to be two places where this `SiameseTransform` code (near identical) needs fixing in the notebook. Will you be able to make the changes and submit a PR? I think that might be easiest for Jeremy and co to review and merge. Might also be worth adding in some cells with the `assert` test that you used, to highlight the importance of checking and testing for such train/val leakage? (the tutorial notebook is not part of the fastai library code base, and so there is no obvious way to add the `assert` test elsewhere, other than within the notebook itself to show)

yijinlee:
Actually, having thought about it a bit more, does it matter that the second image in the Siamese pair, which is the 'reference' to compare with the first image (from validation set), is in the training set? During inference, you would be feeding the trained model a test-image and a reference-image (known class/category) anyways, right? And so the validation-set is being used in a similar way, i.e. comparing the validation-image with a reference-image (known class/category from training set), right?

Interested to hear from others who have more knowledge and experience in this...

ucohen:
[...]

PR #2851, opened 2020-10-02
fixed issue #2781 with code and test from ucohen

yijinlee:
Used the code from @ucohen. Had to modify the test:
```python
valids = [v[0] for k,v in tfm.valid.items()]      
assert not [v for v in valids if v in files[splits[0]]]
```


jph00:
Great!

PR #2853, opened 2020-10-03
added support for tb projector word embeddings

floleuerer:
Added support for exporting word embeddings for Tensorboard projector (function tensorboard_word_embeddings) and updated tests and documentation.

tensorboard_word_embeddings will automatically get the correct layer and vocab for a fast.ai model - for transformers or other models the embedding layer and vocab can be passed as parameters.


floleuerer:
I've added feat['vec'].squeeze() to support unflattend layers for image models (e.g. pooling layers). Changed docs and tests to show the different options how to pass the callback (fit_one_cycle(5, cbs=cbs), get_preds(dl=dl, cbs=cbs)).

Issue #2755, opened 2020-09-04
Add option to create TabularDataLoaders.test_dl() inplace


plynch-chwy:
Feature requests should first be proposed on [the forum](https://forums.fast.ai). 

**Link to forum discussion.**

https://forums.fast.ai/t/tabulardataloaders-test-dl-cant-be-created-in-place/78172

**Is your feature request related to a problem? Please describe.**
When working with large pandas dataframes, ram can end up being a bottle neck. `learner.dls.test_dl()` doesn’t allow passing `inplace=True` to the new dataset, so it ends up making a new copy.

**Describe the solution you'd like**
I'd like to be able to pass `inplace=True` to `learner.dls.test_dl()`.

**Describe alternatives you've considered**
I manually took the code from `TabularDataLoaders.test_dl()` to get the dataloader using an inplace df, but it doesn't _feel_ very fastai like.
```
to = type(learner.dls.train_ds)(df, do_setup=False, reduce_memory=False, y_block=TransformBlock(), inplace=True, **attrdict(learner.dls.train_ds, 'procs','cat_names','cont_names','y_names', 'device'))
to.process()
dl = learner.dls.valid.new(to, bs=bs, inplace=True, device="cuda")
```

**Additional context**
Currently [`TabularDataLoaders.test_dl()` accepts](https://github.com/fastai/fastai/blob/master/fastai/tabular/data.py#L31:L33) `**kwargs`, but those are passed to `self.valid.new`. My naive approach would be to add `inplace=False` to `Tabular.new()` and `TabularDataLoaders.test_dl()`, passing `inplace` to `self.train_ds.new(test_items, inplace=inplace)`. I see that `Tabular.new()` since that is used in a handful of places though, so I can't be sure of the overall impact.

RMNT:
I have encountered this problem when working with fastai on google colab. Inplace option would be a life saver

muellerzr:
I plan on working on this over the weekend

PR #2858, opened 2020-10-03
Tabular: Add option for inplace test_dl and factory methods

muellerzr:
Adds the ability to preprocess your dataset inplace (rather than making a copy). Includes a test

Originated from this FR: https://github.com/fastai/fastai/issues/2755


muellerzr:
Also I removed the `rm_type_tfms` here as it seemed much more centralized. I'll update the other tabular PR once this one has been merged (since this one and the prior PR are much more minor)

PR #2860, opened 2020-10-03
Added clarification to DataBlock tutorial

stevegoldman:
I noted that the TextBlock tokenization process puts tokenized text into a column called "text", even if the original text was in a column with another name.  I found this behavior confusing when I went through the tutorial.


jph00:
Many thanks.

PR #2863, opened 2020-10-05
PILDicom update

asvcode:
For a majority of DICOM images we usually need to `scale` the images so that the pixels represent the correct tissue densities.  Once the images are scaled you can view the normalized images using the `show` function.  This however does not get translated when you use `PILDicom`.  By default `PILDicom` should display the best visualization.  This update also gets round `ValueError: image has wrong mode` errors.

**Examples and Tests**

**SIIM dataset**
(current `PILDicom`)

scale1

(updated `PILDicom`)

scale2

minor visual changes (checking to see that there are no errors with the update) and this dataset does not have a `RescaleSlope` or `RescaleIntercept` so does not really take into consideration the full scale of the update.

**osic-pulmonary-fibrosis**  [dataset](https://www.kaggle.com/c/osic-pulmonary-fibrosis-progression)

(current `PILDicom`)
This throws a `ValueError: image has wrong mode` because the images are 32-bit signed integer pixels.  However even after converting the mode to 32-bit floating point pixels results in the following images .

scale3

(updated `PILDicom`)

scale4

The updated version no longer throws the wrong mode error and the images look a lot better out of the box.

**Melanoma** [dataset](https://www.kaggle.com/c/siim-isic-melanoma-classification) - Note that the `Photometric Interpretation` in this dataset is `YBR_FULL_422` hence the color of the images.

(current `PILDicom`)

scale5

(updated `PILDicom`)

scale6

**RSNA Pulmonary** [dataset](https://www.kaggle.com/c/rsna-str-pulmonary-embolism-detection)

(current `PILDicom`) - will throw a `ValueError: image has wrong mode`  error and after changing the mode:

scale7

(updated `PILDicom`)

scale8

### DIFF
--- a/fastai/medical/imaging.py
+++ b/fastai/medical/imaging.py
@@ -45,11 +45,11 @@ class PILDicom(PILBase):
     @classmethod
     def create(cls, fn:(Path,str,bytes), mode=None)->None:
         "Open a `DICOM file` from path `fn` or bytes `fn` and load it as a `PIL Image`"
-        if isinstance(fn,bytes): im = Image.fromarray(pydicom.dcmread(pydicom.filebase.DicomBytesIO(fn)).pixel_array)
-        if isinstance(fn,(Path,str)): im = Image.fromarray(dcmread(fn).pixel_array)
-        im.load()
-        im = im._new(im.im)
-        return cls(im.convert(mode) if mode else im)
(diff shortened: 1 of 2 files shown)


jph00:
I don't agree we should create them like this - we have ways to show scaled images, so I think the stored array should be the plain values.

PR #2865, opened 2020-10-05
ShowGraphCallback suppressed when calling learn.predict()

sutt:
Fix for forum item: [https://forums.fast.ai/t/showgraphcallback-and-learn-predict-do-not-work-together/79980](https://forums.fast.ai/t/showgraphcallback-and-learn-predict-do-not-work-together/79980)

This change will suppress ShowGraphCallback from calling its `.before_fit` event when learner uses methods .predict or .get_preds or .validate.

This [notebook demonstrates](https://github.com/sutt/fastai-tutorial/blob/master/bug-showgraphcb-1.ipynb) how to quickly reproduce the bug. It only appears on LMLearner and not on synthetic_learner.


jph00:
Thanks!

PR #2866, opened 2020-10-05
Allow for setting of `timeout` for `untar_data` and `download_data`

muellerzr:


muellerzr:
I am currently partial to this solution, as it solves half the problem. `untar_data` has issues if the URL is not poised exactly with the filename, leading to storage issues. Let me work on that as well

muellerzr:
@jph00 my solution was to change the `.name` being used to save when passing in a custom `dest`. the prior did not quite make sense because if I pass in a `dest` I would expect my data to be saved to `dest` not `dest/mnist` (for example). Let me know any other thoughts in regards to that.

muellerzr:
Currently there is a failing test as a result of that, but I don't want to change anything without your thoughts first of course. The issue with this is the urls used by google drive look something like: `https://drive.google.com/uc?export=download&id=1cI2ZIhY2nbc-azOtY1VXr2xfBh21ry` so it will store in a weirdly named `uc?export=download&id=1cI2ZIhY2nbc-azOtY1VXr2xfBh21ry` folder

muellerzr:
I think I'll just leave that issue for another day, one simple pr function at a time :)

jph00:
Thanks!

PR #2867, opened 2020-10-05
Write CSVLogger file after epoch

floleuerer:
The CSVLogger callback writes the csv-file at the end of training. Writing the content of the csv to disk after each epoch makes it easier to keep track of the training without access to the notebook. Added file.flush() and os.fsync(file.fileno()).

see also CSVLogger in fastai1: https://github.com/fastai/fastai1/commit/606f9d3dda712401b693fa0c142482c9221760e0


jph00:
Good idea!

PR #2870, opened 2020-10-07
Added no_random context manager for rng state

marii-moe:
As discussed previous added the ```no_random``` context manager, along with convenience functions. The plan is to use ```no_random``` for documenting fastai resources that use randomness, in order to control the possibility of edge cases confusing the point being made in documentation. Moved this into its own pull request, I will be using this for vision augmentations if it is accepted. 

The name ```no_random``` was chosen mainly because I plan on using this in documentation, and want the purpose to be painfully clear. 

While creating this I found pytorch has something very similar here: https://pytorch.org/docs/stable/_modules/torch/random.html#fork_rng


sutt:
I'll second this idea...it threw me for a loop a while back as wrote up here: https://forums.fast.ai/t/identical-loaded-model-not-reproducible/68890/9

@marii-moe  are you sure this can affect the RNG state in dataloaders/learners which are designed to purposefully not share state?

jph00:
Super!

marii-moe:
> 
> 
> I'll second this idea...it threw me for a loop a while back as wrote up here: https://forums.fast.ai/t/identical-loaded-model-not-reproducible/68890/9
> 
> @marii-moe are you sure this can affect the RNG state in dataloaders/learners which are designed to purposefully not share state?

I include in the docs that this will not affect those RNGs, it seems to be a semi-common problem people run into. If you did want to control those random number generators you could probably control the initialization seed like so:
```with no_random: ImageDataLoaders(...) ```
but you would need to recreate the dataloader every time. 
The way you did it in your forum thread is how I usually control the RNGs. 

Your problem is difficult for me to solve, without requiring people to have an understanding of how dataloaders work. 

My only idea would be to keep a list of [weakrefs](https://docs.python.org/3/library/weakref.html), and patch the RNG initializers to require them to be added to this list. Simply iterate through this list to store/reload RNG state. This is a harder problem though, and I don't feel confident about solving it now, but I will think on it. 

Here is what I mentioned in the notebook: 
> Note: It is important to remember that classes like Dataloader have internal random number generators, and no_random will have no effect on those random number generators.

PR #2875, opened 2020-10-09
wandb callback: support logging lists of parameters

guyrosin:
When using `WandbCallback` I got an exception:
"WandbCallback could not log config parameters -> Object of type function is not JSON serializable"

This happens because the config dictionary contains an item that's mapped to a list of functions (`Learner.__init__.metrics` is mapped to the list of metrics used by fastai's Learner).
`wandb._format_config()` supports serializing functions, but not lists of functions - which is the type of the fastai metrics object.
This PR updates `_format_config()` with that required support.

### DIFF
--- a/fastai/callback/wandb.py
+++ b/fastai/callback/wandb.py
@@ -136,13 +136,27 @@ def _make_plt(img):
     return fig, ax
 
 # Cell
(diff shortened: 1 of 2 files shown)


jph00:
Many thanks

Issue #2852, opened 2020-10-02
learn.summary() gives error if the learner contains cbs


baidut:
fastai     : 2.0.15
fastcore   : 1.0.9

**Describe the bug**
learn.summary() gives error if the learner contains cbs

**To Reproduce**
```python
from fastai.vision.all import *
path = untar_data(URLs.PETS)/'images'

def is_cat(x): return x[0].isupper()
dls = ImageDataLoaders.from_name_func(
    path, get_image_files(path), valid_pct=0.2, seed=42,
    label_func=is_cat, item_tfms=Resize(224))

# TypeError: TypeError: unsupported operand type(s) for -: 'NoneType' and 'float'
learn = cnn_learner(dls, resnet34, metrics=error_rate, cbs = [SaveModelCallback()])

# IndexError: list index out of range
# learn = cnn_learner(dls, resnet34, metrics=error_rate, cbs = [ShowGraphCallback()])
learn.summary()
```

bdatko:
I am seeing the same issue on the newer version.

fastai : 2.0.16
fastcore : 1.2.3

```python
from fastai.vision.all import *
path = untar_data(URLs.PETS)/'images'


def is_cat(x): return x[0].isupper()
dls = ImageDataLoaders.from_name_func(
    path, get_image_files(path), valid_pct=0.2, seed=42,
    label_func=is_cat, item_tfms=Resize(224))

# TypeError: TypeError: unsupported operand type(s) for -: 'NoneType' and 'float'
learn = cnn_learner(dls, resnet34, metrics=error_rate, cbs = [SaveModelCallback()])

# IndexError: list index out of range
# learn = cnn_learner(dls, resnet34, metrics=error_rate, cbs = [ShowGraphCallback()])
learn.summary()
```

jph00:
Many thanks for the clear reporter and repro!

PR #2881, opened 2020-10-15
fix SaveModelCallback to work with learn.summary()

sutt:
Fixes issue: https://github.com/fastai/fastai/issues/2852

The fix simply temporarily removes `SavedModelCallback` so it can run through the problem code then adds it back on.

**Important explanation:** Why use nested `with` statements?
```
        with learn.removed_cbs(train_only_cbs) as l:
            with l: r = l.get_preds(dl=[batch], inner=True, reorder=False)
```

The original code uses `with learn:` which triggers `learn.__enter__()` and  `learn.__exit__()` before and after the call to `get_preds()`. This could be important for setup/teardown reasons.

When we add the block `with learn.remove_cbs(...):` this does *not* fire `__enter__` and `__exit__` for learn. Instead this with-block is really about temporarily removing cbs and restoring them. In other words, the inner with-block preserves the behavior of the previous code; the outer loop fixes the bug.

**Future Extensions:**
`_train_loop_only` can be attached to other callbacks to prevent them from creating an error

### DIFF
--- a/fastai/callback/hook.py
+++ b/fastai/callback/hook.py
@@ -147,7 +147,9 @@ def layer_info(learn, *xb):
     def _track(m, i, o): return (m.__class__.__name__,)+total_params(m)+(apply(lambda x:x.shape, o),)
     with Hooks(flatten_model(learn.model), _track) as h:
         batch = apply(lambda o:o[:1], xb)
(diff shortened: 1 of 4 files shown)


jph00:
Many thanks.

PR #2888, opened 2020-10-19
RandomCrop when Target is Larger than Image Size

pford221:
Hello,

I hope this PR is found to be useful. This implements a quick fix for when the target size of the image is larger than the original size of the image. `random.randint()` requires `a` <= `b` in the signature. 

The only tricky part is that the upper bound of the `self.tl` when target size > original size must be negative such that the padding occurs. This is why we have `(wd, -1)` and `(hd, -1)` for width and height respectively when that condition is met. 

I hope it's to be expected that the `__version__` updated automatically as that was not a change I made. I'm also uncertain what happened with `nbs/04_data.external.ipynb`. I didn't intend to change anything in that notebook. 

Thanks in advance for your consideration of this PR.


jph00:
Many thanks @pford221 . Can you please resolve the conflicts and at-mention me when ready to merge?

pford221:
@jph00, merge conflicts have been resolved. Thanks!

jph00:
Great :)

Issue #2850, opened 2020-10-02
register_forward_hook in AWD_LSTM not working


arnaujc91:
Why are the forward hooks not working for the `AWD_LSTM` model? 

I try the following:

```
import torch 
import torch.nn as nn
from fastai.text.all import *

def hook_fn(m, i, o):
  print(m._get_name())
  print("     HOOK WORKING!    ")

model =  AWD_LSTM(3, 5, 6, 2)

for m in flatten_model(model):
    if has_params(m):
        m.register_forward_hook(hook_fn)

model(torch.randint(3, (64,72)))
```
and the output is just the result of forwarding the input, no hooks activated. 
Output:
```
tensor([[[ 0.1744,  0.2180, -0.2876,  0.2426, -0.1670],
         [ 0.1820,  0.2045, -0.2988,  0.2179, -0.1540],
         [ 0.1936,  0.2058, -0.2988,  0.2025, -0.1598],
         ...,
         [ 0.1955,  0.2122, -0.3008,  0.1770, -0.1693],
         [ 0.2033,  0.2148, -0.2982,  0.1771, -0.1725],
         [ 0.2020,  0.2136, -0.2995,  0.1757, -0.1727]],

        [[ 0.1591,  0.2153, -0.2708,  0.2311, -0.1720],
         [ 0.1792,  0.2202, -0.2829,  0.2234, -0.1747],
         [ 0.1858,  0.2211, -0.2892,  0.2106, -0.1770],
         ...,
         [ 0.1911,  0.2190, -0.2972,  0.1798, -0.1790],
         [ 0.1910,  0.2182, -0.2972,  0.1784, -0.1794],
         [ 0.1920,  0.2187, -0.2975,  0.1799, -0.1788]],

        [[ 0.1521,  0.2057, -0.2798,  0.2316, -0.1627],
         [ 0.1764,  0.2048, -0.2896,  0.2143, -0.1631],
         [ 0.1866,  0.2077, -0.2941,  0.2010, -0.1676],
         ...,
         [ 0.1960,  0.2209, -0.2964,  0.1804, -0.1803],
         [ 0.1960,  0.2209, -0.2964,  0.1804, -0.1803],
         [ 0.1960,  0.2209, -0.2964,  0.1804, -0.1803]],

        ...,

        [[ 0.1432,  0.2092, -0.2800,  0.2401, -0.1603],
         [ 0.1666,  0.2100, -0.2891,  0.2253, -0.1622],
[...]

PR #2890, opened 2020-10-19
allow ActivationStats cb to work with AWD_LSTM

sutt:
Fix for: https://github.com/fastai/fastai/issues/2850
The ActivationStats callback needs modification to work with the AWD_LSTM model:
 - First, slightly modify the forward method of `awd_lstm.WeightDropout`
 - Second, allow the hook function to work when the layer output is a tuple of tensors, like with a LSTM layer.

A simple example of using this modification can be seen here: https://github.com/sutt/fastai2-dev/blob/master/oct-misc/actstats-2-demo-2.ipynb (inspired by @arnaujc91)

**Details:**
WeightDropout.forward now uses `__call__` on it's module, instead of the module's forward. This allows hooks which are handled in `pytorch.nn.Module._call_impl` to work for this callback (or any HookCallback). Without this, the hook for the LSTM module inside WeightDropout never get's called. You can see a demonstration of this detail here: https://github.com/sutt/fastai2-dev/blob/master/oct-misc/actstats-2-demo.ipynb)

Since an LSTM module returns a nested tuple, but we want to do stats on the values inside the respective tensors, we handle this by flattening any tuple output and computing the mean/std/hist for each of the output tensors.

Stats in Embedding layers are still not collected for in the callback.


jph00:
Many thanks.

PR #2896, opened 2020-10-22
Docs for vision augmentation

marii-moe:
The code was reordered, but mostly to group like functions under different headings. There should be no change to code logic.


marii-moe:
Probably needs to be more documentation as well, but wanted to commit what I have right now because I am kind of stuck on how to head forward.

jph00:
Wow this is fantastic! :D Thanks @marii-moe

PR #2903, opened 2020-10-24
Fix RuntimeError when passing weights to BCEWithLogitsLossFlat

### LINKED ISSUE #2745: Weighted BCEWithLogitsLossFlat Loss Function Shape Mismatch
Related forum post: https://forums.fast.ai/t/customized-loss-function-with-weights-for-multi-label-multi-classification/47517/9

Please confirm you have the latest versions of fastai, fastcore, fastscript, and nbdev prior to reporting a bug: YES

**Describe the bug**
`BCEWithLogitsLossFlat` fails with a `RuntimeError` due to shape mismatch when you pass in weights through the `pos_weight` argument. This happens because the targets/inputs are being flattened per `BaseLoss`'s `flatten` argument, which is not exposed in `BCEWithLogitsLossFlat`'s initialiser.

**To Reproduce**
1. Setup dummy data
```python
#dummy ont-hot encoded data
bs,c = 64,10
yb,preds = torch.ones(bs,c), torch.ones(bs,c)
pos_wt   = torch.ones(c)
```

2. Compare with `nn.BCEWithLogitsLoss()` with and without weighted loss
```python
#unweighted loss function, works fine
lf_torch  = nn.BCEWithLogitsLoss()
lf_fastai = BCEWithLogitsLossFlat()

assert torch.allclose(lf_torch(yb,preds), lf_fastai(yb,preds)) # passes
```

```python
#weighted loss function, throws RuntimeError due to shape mismatch
#caused by flattening
lf_torch  = nn.BCEWithLogitsLoss(pos_weight=pos_wt)
lf_fastai = BCEWithLogitsLossFlat(pos_weight=pos_wt)

assert torch.allclose(lf_torch(yb,preds), lf_fastai(yb,preds)) # -> RuntimeError
```
```python
<ipython-input-186-c4b189ff236f> in <module>
      4 lf_fastai = BCEWithLogitsLossFlat(pos_weight=pos_wt)
      5 
----> 6 assert lf_torch(yb,preds) == lf_fastai(yb,preds)

~/anaconda3/lib/python3.8/site-packages/fastai/layers.py in __call__(self, inp, targ, **kwargs)
    295         if targ.dtype in [torch.int8, torch.int16, torch.int32]: targ = targ.long()
    296         if self.flatten: inp = inp.view(-1,inp.shape[-1]) if self.is_2d else inp.view(-1)
--> 297         return self.func.__call__(inp, targ.view(-1) if self.flatten else targ, **kwargs)
    298 
    299 # Cell

~/anaconda3/lib/python3.8/site-packages/torch/nn/modules/module.py in _call_impl(self, *input, **kwargs)
    720             result = self._slow_forward(*input, **kwargs)
    721         else:
--> 722             result = self.forward(*input, **kwargs)
    723         for hook in itertools.chain(
    724                 _global_forward_hooks.values(),

~/anaconda3/lib/python3.8/site-packages/torch/nn/modules/loss.py in forward(self, input, target)
    626 
    627     def forward(self, input: Tensor, target: Tensor) -> Tensor:
--> 628         return F.binary_cross_entropy_with_logits(input, target,
    629                                                   self.weight,
    630                                                   pos_weight=self.pos_weight,

~/anaconda3/lib/python3.8/site-packages/torch/nn/functional.py in binary_cross_entropy_with_logits(input, target, weight, size_average, reduce, reduction, pos_weight)
   2538         raise ValueError("Target size ({}) must be the same as input size ({})".format(target.size(), input.size()))
   2539 
-> 2540     return torch.binary_cross_entropy_with_logits(input, target, weight, pos_weight, reduction_enum)
   2541 
   2542 

RuntimeError: The size of tensor a (10) must match the size of tensor b (640) at non-singleton dimension 0
```

**Solution**
Passing `flatten=True` to `BaseLoss`'s initialiser fixes the issue.
```python
class BCEWithLogitsLossFlat(BaseLoss):
    "Same as `nn.CrossEntropyLoss`, but flattens input and target."
    @use_kwargs_dict(keep=True, weight=None, reduction='mean', pos_weight=None)
    def __init__(self, *args, axis=-1, floatify=True, thresh=0.5, **kwargs):
        super().__init__(nn.BCEWithLogitsLoss, *args, axis=axis, floatify=floatify, is_2d=False,
                         flatten=False, # this fixes it
                         **kwargs)
        self.thresh = thresh

    def decodes(self, x):    return x>self.thresh
    def activation(self, x): return torch.sigmoid(x)

lf_torch  = nn.BCEWithLogitsLoss(pos_weight=pos_wt)
lf_fastai = BCEWithLogitsLossFlat(pos_weight=pos_wt)

assert torch.allclose(lf_torch(yb,preds), lf_fastai(yb,preds)) # passes
```

What I'm unclear of is how changing the `flatten` parameter will affect other parts of the pipeline. Perhaps the `flatten` parameter should be exposed in `BCEWithLogitsLossFlat`'s init or be changed to `True` when `pos_weight is not None`?

rsomani95:
Fixes #2745


fastai/losses.py
```diff
@@ -51,6 +51,8 @@ class BCEWithLogitsLossFlat(BaseLoss):
     "Same as `nn.BCEWithLogitsLoss`, but flattens input and target."
     @use_kwargs_dict(keep=True, weight=None, reduction='mean', pos_weight=None)
     def __init__(self, *args, axis=-1, floatify=True, thresh=0.5, **kwargs):
+        # `flatten`=True throws a RuntimeError due to shape mismatch
```
rsomani95, line 54: Should we print out a warning here in case the user explicitly passes `flatten=True` when creating the loss function?

nbs/01a_losses.ipynb
```diff
@@ -300,7 +304,21 @@
    "cell_type": "code",
    "execution_count": null,
    "metadata": {},
-   "outputs": [],
+   "outputs": [
+    {
+     "ename": "AssertionError",
+     "evalue": "Torch not compiled with CUDA enabled",
```
rsomani95, line 310: I think this error is being caused automatically through some github workflow action?
I created three different forks and pushed the same code without running this cell hoping it would go away but it didn't. Apologies if I'm missing something obvious here.

fastai/losses.py
```diff
@@ -51,6 +51,8 @@ class BCEWithLogitsLossFlat(BaseLoss):
...
+        if kwargs.get('pos_weight') is not None: kwargs['flatten'] = False
```
jph00, line 55:
```suggestion
        if kwargs.get('pos_weight', None) is not None: kwargs['flatten'] = False
```

nbs/01a_losses.ipynb
jph00: Comment out the cell and run it to remove the outputs, and then uncomment. The cell won't be executed in CI

fastai/losses.py
jph00: If it's an error, then maybe better to throw a clear exception here, rather than just warn?

nbs/01a_losses.ipynb
rsomani95: Done in 8bd7e9f9d4ea53127dbae2beb122daf7549a624e

fastai/losses.py
rsomani95: Definitely. Added this in 732a85cfcff77bcef458a4b32188e9915cad1b1a

fastai/losses.py
rsomani95: Added in 732a85cfcff77bcef458a4b32188e9915cad1b1a

jph00:
Thanks!

PR #2907, opened 2020-10-27
Fixed bug with hue transform. 

marii-moe:
Fixed bug with hue transform not applying when using it in a dataloader. Noticed that most code is a "space transform", and therefore closely mirrors the functionality in lighting transform. Since the code would be almost all duplicate, refactored most of the code in lighting(logit space) transform to a "SpaceTfm."


jph00:
Very cool. Can you please resolve the conflicts and at-mention me so I can merge when ready?

marii-moe:
@jph00 Conflicts resolved! `nbdev_fix_merge` did all the work.

PR #2908, opened 2020-10-27
Updated wording of docs to more closely mirror expected input.

### LINKED ISSUE #2785: Passing custom function or array to Rotate in transforms

marii-moe:
Updated docs to more closely match expected inputs. Next plan is to include examples of usage, as draw is very confusing from the point of view of someone who is not familiar with _draw_mask. 

Fixes #2785


marii-moe:
For some reason my local environment is broken with the newest version. Going to close this for now.

PR #2914, opened 2020-10-28
feat(wandb): allow custom description

borisdayma:
Allow custom description in W&B artifacts as per [Rahul's request](https://forums.fast.ai/t/workflow-to-compare-monitor-models-using-wandbcallback/68673/40?u=boris)


jph00:
@borisdayma just in case you didn't see on Discord yet, I've had to remove `log_args` due to a memory leak. Let's discuss how to fix up any missing functionality as soon as you have a moment.

PR #2921, opened 2020-10-31
Augmentation docs

marii-moe:
Adds more documentation to vision augmentation. Adds support for suptitles,or titles for all of the images, to show_images.


marii-moe:
I noticed on reviewnb the augmentation notebook ended up being formatted weirdly around the warp transform. It looks fine in my local and here: https://github.com/fastai/fastai/blob/bae078f9a18ee392c78d762471b55a240c7ef675/nbs/09_vision.augment.ipynb

jph00:
Thank you!  Very nice :)

Issue #2926, opened 2020-11-02
torch 1.7.0 incompatibilty 


pford221:
Hello,

There is a thread on the forums that reproduces a potential bug with `AWD_LSTM` with torch==1.7.0
https://forums.fast.ai/t/the-size-of-tensor-a-400-must-match-the-size-of-tensor-b-0-at-non-singleton-dimension-2/81277

I've investigated and I believe the source of the problem is line 14 in awdlstm.py. Specifically this line:
```x.new(*sz).bernoulli_(1-p).div_(1-p)```

When `x` is a `TensorText` the `new` method does not appear to be compatible and it returns nothing. 

Here's a simple example:

```
from fastai.text.data import TensorText
tt = TensorText([[1,2,3], [4,5,6]])
tt.new(size=(2,3))
```
You'll notice that what is returned is an empty TensorText object where I think  the desire is to return the data contents of `tt`.

I think using `new_zeros` will work just fine for line 14 of awdlstm.py. I will make a PR to this effect.

Thanks!

pford221:
Hi, 

I wrote this in PR #2927 as it was being merged, so at the risk of duplicating myself, I'll write it again:

-----
I'm afraid that this incompatibility is more pervasive than just this instance (I was afraid of that!). Another case is in line 172 of fastai.text.learner in the LMLearner.predict() method. Again, calling idxs.new([idx]) is returning an empty TensorText.

I'm happy to do what I can to try to spot other instances with a broad find and replace but I worry that I won't always know when the tensor is TensorText and whether this issue extends to other sublasses of torch.Tensor.
------

jph00:
Fixed now.

PR #2927, opened 2020-11-02
Fix/new tensor for dropout

pford221:
Hello,

This is a very simple PR related to [#2926](https://github.com/fastai/fastai/issues/2926). Please let me know if you have concerns.

Thanks!

### DIFF
--- a/fastai/text/models/awdlstm.py
+++ b/fastai/text/models/awdlstm.py
@@ -11,7 +11,7 @@ from ..core import *
 # Cell
 def dropout_mask(x, sz, p):
     "Return a dropout mask of the same type as `x`, size `sz`, with probability `p` to cancel an element."
-    return x.new(*sz).bernoulli_(1-p).div_(1-p)
+    return x.new_zeros(*sz).bernoulli_(1-p).div_(1-p)
 
 # Cell
 class RNNDropout(Module):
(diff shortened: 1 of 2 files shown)


jph00 (changes requested):
Thanks @pford221 ! I think actually `new_empty` would be better. Would you mind making that change?

pford221:
Hi Jeremy,

I'm afraid that this incompatibility is more pervasive than just this instance (I was afraid of that!). Another case is in line 172 of fastai.text.learner in the `LMLearner.predict()` method. Again, calling `idxs.new([idx])` is returning an empty TensorText. 

I'm happy to do what I can to try to spot other instances with a broad find and replace but I worry that I won't always know when the tensor is `TensorText` and whether this issue extends to other sublasses of `torch.Tensor`.

jph00:
My bad - I actually knew about this issue already, but forgot to rigorously check for it. It's actually a change in the PyTorch behavior of `Tensor.new`. We need to replace *all* uses of `new` with `new_empty` for any kind of tensor.

I'll work on fixing this myself tomorrow, time permitting - although a PR in the meantime if you happen to have the time and inclination would be most welcome.

It also sounds like we're missing tests to cover some functionality, since we clearly have bugs which aren't being picked up by tests! :(

pford221:
I tried to update line 172 of fastai.text.learner in the LMLearner.predict() method to:
```
idxs.new_empty([idx])
```
And it threw a CUDA error, "device-side assert triggered". So then I tried to make sure this tensor is on the same device as `idxs` and changed it to:
```
idxs.new_empty([idxs], device=idxs_all.device)
```
Unfortunately that didn't work either.

My point is that it appears - at least in select cases - that simply updating  `new` to `new_empty` might lead to other downstream issues.

jph00:
Yeah sorry what I said was wrong - anyhoo I've patched the PyTorch bug now so `new` works again :)

Issue #2910, opened 2020-10-27
In "Transfer learning in text" tutorial, the "dls.show_batch()" show wrong outputs


dgg32:
I am following the [Beginner tutorial](https://docs.fast.ai/tutorial.text). When I run the code both in my local machine in Colab, several times, the output of cell 5 "dls.show_batch()" are wrong (screenshot below),  from the second item downwards they are all "xxpad xxpad xxpad xxpad xxpad xxpad". These are also different from what the tutorial page showed:

My run:
<img width="1391" alt="Bildschirmfoto 2020-10-27 um 19 52 56" src="https://user-images.githubusercontent.com/1873196/97348172-1748c600-188e-11eb-8d88-778c26505580.png">

Tutorial page:
<img width="779" alt="Bildschirmfoto 2020-10-27 um 19 55 41" src="https://user-images.githubusercontent.com/1873196/97348522-8cb49680-188e-11eb-8302-e8a77bde1ea7.png">

What is the cause of this issue please?

marii-moe:
Starting working on this one.

marii-moe:
I am able to reproduce this issue in my local environment.

PR #2932, opened 2020-11-03
Updated so that padding is removed during decode. Fixes: #2910

marii-moe:
Before pad_tok was introduced we used remove PAD('xxpad') by default, you can find that here: https://github.com/fastai/fastai/blame/adeb4483f43eeff00289874b506b08fbf05d3204/fastai2/text/data.py#L46

Updated 'path', because it seemed to be using regular IMDB when slow commands ARE allowed, but not when slow are not run. This causes errors as the IMDB dataset does not have texts.csv. 

Fixes: #2910


marii-moe:
VisibleDeprecationWarning I am still looking into, but that may be out of scope of this fix.

jph00:
Awesome job! :D

Issue #2911, opened 2020-10-28
Confusing behavior and documentation of tabular test dataloader


byte01:
Feature requests should first be proposed on [the forum](https://forums.fast.ai). 

**Link to forum discussion.**

https://forums.fast.ai/t/kaggle-titanic-and-decisiontreeregressor/79655/5?u=maxim

**Is your feature request related to a problem? Please describe.**
Currently tst_dl = dls.valid.new(to_tst) doesn't work when a test set has some missing values other than the train set.

**Describe the solution you'd like**
~~A thinking machine, capable of AGI.~~
Please provide a clear documented way to predict on a test set with different missing values than training set with tabular data loader. Maybe there is a way to automatically add proper NaN columns to test and train sets, without having to do it manually?

**Describe alternatives you've considered**
Currently

dls = to.dataloaders()
test_dl = dls.test_dl(df_test)
m.predict(test_dl.dataset.xs[to_keep].drop(to_drop, axis=1))

works (on kaggle House Prices: Advanced Regression Techniques) if I add a fake row with NaNs (in places where NaNs appear in train+test dataset) to my train set.

While
to_test = to.valid.new(df_test)
m.predict(to_test.xs[to_keep].drop(to_drop, axis=1))

doesn't work, but in documentation it is shown as one of the ways to get test predictions:
https://docs.fast.ai/tabular.core

**Additional context**
The way I add a fake row with NaNs:

df_all = df.append(df_test)
display_all(df_all.isnull().sum().sort_index()/len(df_all))

dep_var = ‘SalePrice’

nas = df_all.isnull().sum()
nas[dep_var] = 0
nas

df_mode = df.mode(axis=0,dropna=False)
df_mode

fake_nan_row = np.where(nas == 0, df_mode, np.NaN)
pd.DataFrame(fake_nan_row)

df = df.append(pd.DataFrame(fake_nan_row,columns=df_mode.columns,index=[len(df)]))
display_all(df)

fake_idx = len(df)-1
fake_idx

def idx_move_train(splits, idx):
    if idx in splits[0]:
        return splits
    else:
        return (splits[0]+[idx], splits[1].remove(idx))
splits =  RandomSplitter(valid_pct=0.2)(range_of(df))

splits = idx_move_train(splits, fake_idx)

jph00:
@muellerzr would you be open to taking a look at this? Do you already have a notebook that might be suitable?

muellerzr:
[...]

PR #2933, opened 2020-11-03
Improve and clean tabular documentation, add note about the capabilities of ML models

muellerzr:
Helps solve https://github.com/fastai/fastai/issues/2911

I've cleaned and hid most of the tests that are unnecessary to show in the actual documentation and explained a few more aspects of tabular cleaner. To address the issue I added this notice to both the tabular tutorial notebook as well as tabular core:

> Note: Since machine learning models can't magically understand categories it was never trained on, the data should reflect this. If there are different missing values in your test data you should address this before training

Otherwise the other major change is adjustments to `Categorize` and `Normalize` which are done with `@Categorize` I changed to `exporti` to help clean up the documentation


muellerzr:
@jph00 let me know if there are any other improvements you can think of or better ways to rephrase that

jph00:
This looks great! I didn't see where you added `exporti`, but note that iirc this removes from `__all__`, which is possibly not what you meant to do? If not, perhaps undo that with a followup PR? Let me know what cleanup you were trying to achieve with it. Note that you can always put `show_doc` for a symbol somewhere else in an nb, to stop it from displaying in the default place

muellerzr:
@jph00 it's meant to clean up sections like so:
https://docs.fast.ai/tabular.core#setups

Aka places where `@Categorize` and `@Normalize` are. They don't display well in the documentation, and results in a very awkward looking placeholder where we might not necessarily *want* to show `encodes`, `decodes`, etc but rather say `Normalize`

One compromise I can think of is a `show_doc` or manual doc (since `show_doc` won't show this new behavior) to describe that the `proc` is available and modified. I'm open to your ideas though

jph00:
Oh OK great - yeah stuff like `setups` certainly doesn't need to be in the docs or part of `__all__`, so that's fine. As long as the transform classes themselves are exported.

Issue #2874, opened 2020-10-09
get_preds fails when passed a dataloader where the batchsize does not evenly divide the number of items


adamfarquhar:
Please confirm you have the latest versions of fastai, fastcore, fastscript, and nbdev prior to reporting a bug (delete one): YES

**Describe the bug**
See the forum item: https://forums.fast.ai/t/learn-get-preds-batch-size-and-number-of-items-mismatch-error/79649

If get_preds receives a dataloader where batchsize does not evenly divide the number of items, then it fails with an error:
IndexError: index 98 is out of bounds for dimension 0 with size 96

**To Reproduce**

Call get_preds with a dataloader containing a number of items that is not an integer multiple of the batchsize.

**Expected behavior**
Get_preds should return predictions for all if the items in the dataloader.

**Error with full stack trace**
See the previously reported item on the forum. 
```
```

**Additional context**

dreamflasher:
I can reproduce the bug and would appreciate a fix.

dreamflasher:
The solution is to make predictions in the following way:
```
probs, _ = learn.get_preds(dl=learn.dls.test_dl(files, drop_last=False))
```
Somewhere drop_last defaults to True, which should be changed.

muellerzr:
@dreamflasher @adamfarquhar the issue is you're trying to pass in a *set* of `DataLoaders`, where `get_preds` can only get inference on **one** `DataLoader` at a time. @dreamflasher solution is correct because it is the proper way to use the inference API.

jph00:
@adamfarquhar FYI in the future, in the "To reproduce" section please provide complete code to reproduce the problem from scratch, rather than just a general description.

dreamflasher:
@jph00 @muellerzr Why is the drop_last necessary, though? Shouldn't that default to false for a test_dl?

muellerzr:
It always is. The issue had nothing to do with the DataLoader itself moreso
the fact they were passing in both DataLoaders to get_preds. If you can
create a reproducible example of test_dl dropping the last batch please
open an issue however all test_dl’s should not drop the last batch since
they’re generated based on the validation DataLoader.

kevinbird15:
I looked into this and I think the *problem* is how you are creating your test_dl.  I think the preferred method of getting a test_dl to put into `preds, ground_truth = learn.get_preds(dl=test_dls.train)` would be https://docs.fast.ai/data.core.html#DataLoaders.test_dl and here is an example using your example: 

```python
[...]

PR #2936, opened 2020-11-03
Raises error when passing in `DataLoaders` to `get_preds`

muellerzr:
Adds a more explicit error when trying to pass in multiple `DataLoaders` to `Learner.get_preds`:

"`dl` is something other than a single `DataLoader` object"

Addresses https://github.com/fastai/fastai/issues/2874

( cc: @jph00 )


jph00:
Many thanks!

PR #2937, opened 2020-11-03
Clean up vision docs and exports

muellerzr:
Cleans up the vision documentation by showing a standard block for `@patch`'d properties relating to transforms:
> Calls `@patch`'d `function` behaviors for `type1`, `type2`, etc

Also removed duplicate `CropPad` and kept the `ItemTransform` version


muellerzr:
Not 100% sure why it's trying to do the last pr too, let me investigate quickly

muellerzr:
@jph00 should be good to go once tests pass

PR #2938, opened 2020-11-04
Change res_col_name to tok_text_col

muellerzr:
Adjusts the one-liner for `tokenize_df` to explain what happens to the text and changed the parameter name to something that makes more sense

( cc: @jph00 )


jph00:
Thank you

PR #2940, opened 2020-11-04
Update sidebar.json

stestagg:
Currently, the sidebar on https://docs.fast.ai/ has a bad link for callabcks/progress.

I believe this change should fix that


jph00:
Many thanks.

PR #2942, opened 2020-11-04
fixed train valid leakage in siamse tutorial

kessido:


jph00:
@kessido looks like you need to run `nbdev_clean_nbs`.

kessido:
> @kessido looks like you need to run `nbdev_clean_nbs`.

done, is it ok now?

jph00:
Great!

PR #2944, opened 2020-11-06
Fix BnFreeze bug. `before_epoch` -> `before_batch`

rsomani95:
The tests in the notebook failed when running as is. This is because of a bug in `BnFreeze` where it needs to be called `before_batch` rather than `before_epoch`. I don't know why this is the case, but the test demonstrates that this is indeed the case. I'd opened a [similar PR](https://github.com/fastai/fastai2/pull/375) in the dev version but didn't communicate it clearly. Hopefully this is clearer.

Shall I remove the demonstration of why `before_epoch` fails once this PR is reviewed?


marii-moe:
@rsomani95 I was fixing a different issue and ended up debugging this one in the process, I wasn't aware you were working on this one, sorry about that. This callback seems to be inter-playing with the TrainEvalCallback. This is how I was able to get it working for me: 
```class BnFreeze(Callback):
    run_after=TrainEvalCallback
    "Freeze moving average statistics in all non-trainable batchnorm layers."
    def before_train(self):
        set_bn_eval(self.model)
```
I am not very familiar with the `BnFreeze` code myself. This should be the offending code in `TrainEvalCallback` that makes this not work: 
```def before_train(self):
        "Set the model in training mode"
        self.learn.pct_train=self.epoch/self.n_epoch
        self.model.train()
        self.learn.training=True
```
Wanted to go ahead and contribute this as you are the one leading and most familiar with this one.

marii-moe:
@rsomani95 tracking this one in a different pull request, so closing this one. 

https://github.com/fastai/fastai/pull/2944

PR #2956, opened 2020-11-07
Make Random Splitter use Numpy rather than Pytorch

muellerzr:
As the title says when doing RandomSplitter on a large dataset it winds up taking a long time and can lead to a weird spike in memory (when its obscenely large).

For a quick example when trying on 1,000,000 numbers I found that torch.permute took 7.24 seconds while np.random.permutation only took 377ms

(git history is now good @jph00 :) )

### DIFF
--- a/fastai/data/transforms.py
+++ b/fastai/data/transforms.py
@@ -85,8 +85,8 @@ class AttrGetter(ItemTransform):
 def RandomSplitter(valid_pct=0.2, seed=None):
     "Create function that splits `items` between train/val with `valid_pct` randomly."
     def _inner(o):
-        if seed is not None: torch.manual_seed(seed)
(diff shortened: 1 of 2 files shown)


muellerzr:
I decided to use `np.random.choice` because it's faster. (for 10 million items it reduces it from 1.78 seconds -> 1.21 seconds)

jph00:
Nice. Turns out it wasn't the `permute` that's slow, but `iter` and cast tensor->int. Just pushed a different approach to solving the same problem, which takes the same time as `choice`. Note that when timing stuff like this you should run `gc.collect()` before each timing, otherwise gc overhead will make results highly variable.

muellerzr:
Ah interesting! Okay. Thanks!

PR #2958, opened 2020-11-07
Fix RandomSubsetSplitter upstream adjustment to RandomSplitter

muellerzr:
Applies the same logic as https://github.com/fastai/fastai/issues/2957 but to `RandomSubsetSplitter` too (cc: @jph00 )


jph00:
Thanks!

PR #2959, opened 2020-11-07
feat(wandb): use store_attr

borisdayma:
Logs config parameters through ` store_attr`

I'm now running a few tests to check it works as intended.


borisdayma:
@jph00 I'm happy with it!

Note that the subclassing of Pytorch tensors will be handled directly in wandb repo: https://github.com/wandb/client/pull/1474

PR #2964, opened 2020-11-09
Update to show_install method to use fastcore.run

tcapelle:
Quick fix to `show_install` that was not showing `nvidia-smi`.
Split on smaller funcs and use of `fastcore.run` method (instead of `subprocess`)


jph00 (commented):
Good idea! But you need to `#export` both `nvidia_smi` and `nvidia_mem`.

PR #2966, opened 2020-11-09
Deprecate `config` in `create_cnn` and instead pass kwargs directly

borisdayma:
`cnn_learner` and `unet_learner` have now direct access to all config parameters and don't require a separate `config` object anymore.


borisdayma:
Now I'm getting more params logged than before!

The next step will be to figure out how to log params from `learn.fit`, `learn.fit_one_cycle`, etc… (included method name).

borisdayma:
It should now be backwards compatible

fastai/vision/learner.py
```diff
@@ -161,42 +156,76 @@ def _add_norm(dls, meta, pretrained):
...
+                # learner args
+                loss_func=None, opt_func=Adam, lr=defaults.lr, splitter=None, cbs=None, metrics=None, path=None,
+                model_dir='models', wd=None, wd_bn_bias=False, train_bn=True, moms=(0.95,0.85,0.95),
+                # other model args
+                **kwargs):
     "Build a convnet style learner from `dls` and `arch`"
-    if config is None: config = {}
+
+    if config:
+        print('Warning: config is deprecated and will be removed in the future. Pass your args directly to cnn_learner.')
```
jph00, line 169:
```suggestion
        warnings.warn('config param is deprecated. Pass your args directly to cnn_learner.')
```

borisdayma:
The CI failure seems unrelated

jph00:
Thank you!

PR #2972, opened 2020-11-11
Updated gh installation docs for linux

rajaskakodkar:
Changes: Dropped the conda install step for `gh` and added link from gh [docs](https://github.com/cli/cli/blob/trunk/docs/install_linux.md#installing-gh-on-linux) for installation steps on Linux.


jph00:
Thanks for your PR! :)

PR #2976, opened 2020-11-12
Update Colab link in README

Bharat123rox:
Update Colab link to point to fastai notes, instead of general Welcome page of Google Colab


jph00:
The README is auto-generated from `index.ipynb`. I'll fix it there - thanks for the idea!

PR #2984, opened 2020-11-14
Update classify_url in lesson2-download.ipynb to use PILImage.create

mess-lelouch:
While I was upgrading my application for fastai v2, I realized that the starlette application that I had did not work. Then I realized that the code I used for reference in the fastai repo looked like the one in v1.

What I did here was to update the code to create the item with `PILImage.create`. This is what worked when I updated my application to use fast ai V2.


jph00:
Thank you.

PR #3014, opened 2020-11-19
Pretty-up learn.summary(), get rid of the progress bar on top

muellerzr:
This is a pretty bulky PR in regards to how `learn.summary` was changed. Previously summary was fairly dry in how things were cut together. In this new version when the input sizes are similar in the layers they are grouped together. An example can be seen in the image below: 
image

There is one piece of logic I'm not too thrilled about, here: `if not chnged and not prev_sz == sz and j > 0` which checks if we are the first row or not. This will add a line on top of the layers before we go into the rest of the summary. If there is an obvious way I am missing to make that a bit nicer let me know!

I also added in some line comments as to what each if statement points to, as it's not obvious to read and doesn't really show in the documentation well with those private functions. The only other "weird" addition is I had to repeat a line solely due to the param group potentially being there. Let me know if there is something else that can be improved!


muellerzr:
The utils update/change was due to a mismatch that got overlooked it seems (I made a new fork just before doing this < 1hr ago)

muellerzr:
I'll also note that some tests had to be changed as well. These needed to be adjusted as ReLU is no longer perceived as "trainable" (since it has no parameters) and a second value of `same` is returned as well, which dictates whether or not the input and output shapes remained the same

fastai/callback/hook.py
```diff
@@ -144,12 +144,26 @@ def total_params(m):
 # Cell
 def layer_info(learn, *xb):
     "Return layer infos of `model` on `xb` (only support batch first inputs)"
-    def _track(m, i, o): return (m.__class__.__name__,)+total_params(m)+(apply(lambda x:x.shape, o),)
+    def _track(m, i, o):
+        for inp, out in zip(i,o):
+            if inp.shape[1:] == out.shape: same = True
+            else: same = False
+        if hasattr(m, 'weight'): # non activation layer
```
jph00, line 151: There's a lot of duplicate code here. Can you please clean it up? Also, I don't think we need to check for weights or param groups do we? We should only be checking whether the output size changes.

fastai/callback/hook.py
```diff
@@ -144,12 +144,26 @@ def total_params(m):
...
+
+        if same: return (type(m).__name__,)+('','')+('', '', same) # activation layer
+        else: return (type(m).__name__,)+('','')+(apply(lambda x:x.shape, o), '', same) # pooling layer
+
     with Hooks(flatten_model(learn.model), _track) as h:
         batch = apply(lambda o:o[:1], xb)
         train_only_cbs = [cb for cb in learn.cbs if hasattr(cb, '_only_train_loop')]
-        with learn.removed_cbs(train_only_cbs) as l:
-            with l: r = l.get_preds(dl=[batch], inner=True, reorder=False)
+        with learn.removed_cbs(train_only_cbs) and learn.no_logging():
```
jph00, line 165: `and` should be `,`

fastai/callback/hook.py
```diff
@@ -144,12 +144,26 @@ def total_params(m):
...
+            with learn as l: r = l.get_preds(dl=[batch], inner=True, reorder=False)
```
jph00, line 166: You could move this into the previous line by adding `,learn` AFAICT

fastai/callback/hook.py
```diff
@@ -164,20 +178,31 @@ def module_summary(learn, *xb):
...
-        ps += np
-        if trn: trn_ps += np
-        res += f"{typ:<20} {_print_shapes(sz, bs)[:19]:<20} {np:<10,} {str(trn):<10}\n"
-        res += "_" * n + "\n"
+        if j == 0: res += f'\n{"":<20} {_print_shapes(sz, bs)[:19]:<20}' # to avoid a double line at the top
+        if not chnged and not prev_sz == sz and j > 0: res += "\n" + "_" * n + "\n" + f'{"":<20} {_print_shapes(sz, bs)[:19]:<20}'
+        j = 1
+        if np is '':
+            if sz is not '':  # Pooling layer
+                res += f"\n{typ:<20} {'':<20} {'':<10} {'':<10} {'':<10}"
```
jph00, line 197: Refactor the dupe formatting codes to avoid duplication, and put conditionals on 1 line where you can

jph00 (changes requested):
Thanks Zach! Let's try to simplify this code as much as possible. It's too hard to follow at the moment.

fastai/callback/hook.py
muellerzr: This let's us check what freeze group they are in (so it's something new). Would there be a better way to check what param groups a layer is in other than checking by weight? I also checked for weights because we discussed differentiating between activation functions, poolings, and everything else. Looking for the weights helps me do this.

jph00:
Oh I see - sorry I didn't know there were two things in this PR. Could we separate these issues, and tackle the layer groups issue later (if we decide to do it)? Let's keep each step simple. :)

muellerzr:
Sure. I only included it as it was requested a few times in the forums and looked like a thing that should be added in. Can definitely separate into another PR

fastai/callback/hook.py
muellerzr: @jph00 I've simplified this a bit

muellerzr:
@jph00 just did a heavy refactor and threw out the extra stuff for this pr. Let me know if you have any thoughts.

muellerzr:
I also don't know why 09b is disliking me, looks like somehow a nb build mismatch that made it through the testing API? I don't know

jph00:
thanks!

Issue #3012, opened 2020-11-18
Redundant BatchNorm in create_head after Flatten


muellerzr:
`create_head` will utilize `LinBnDrop` when creating it's head. As a result we end up in a situation such as:
```python
BatchNorm2d -> AdaptiveAvgPool2d -> AdaptiveMaxPool2d -> Flatten -> BatchNorm1d
```
(A better visualization is below): 
![image](https://user-images.githubusercontent.com/7831895/99588600-d0477f80-29b8-11eb-97d9-a27a22e9cb34.png)

The fix is just getting rid of our first batch norm layer in `create_head`

muellerzr:
A better solution would be to check for a batch norm layer at the end of the encoder

jph00:
> The fix is just getting rid of our first batch norm layer in create_head

I don't think that works. I'm not entirely sure this is automatable. Perhaps the best we can hope for is a bool param that can be set via `model_meta`?

PR #3015, opened 2020-11-19
Avoid redundant BatchNorm in create_head

renato145:
An attempt to solve the issue #3012 

I added a `first_bn` param in `create_head`.
And in `create_cnn_model`, it checks if the last layer is a `nn.BatchNorm2d` to set the `first_bn` param.

I though that may not be enough since some operations can be done after the batchnorm that are not necessarily a `nn.Module`, so I added a 2nd check to see if the output of the last batchnorm is the same as the output of the model.

### DIFF
--- a/fastai/vision/learner.py
+++ b/fastai/vision/learner.py
@@ -5,6 +5,7 @@ __all__ = ['has_pool_type', 'create_body', 'create_head', 'default_split', 'mode
 
 # Cell
 from ..basics import *
+from ..callback.hook import *
 from .core import *
 from .data import *
 from .augment import *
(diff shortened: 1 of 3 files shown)


jph00:
This is super neat, but I think it's a bit to complex for this foundational function.

I like the changes to `create_head`, so let's keep those. But could you remove the changes in `create_cnn_model` so this becomes just a flag people pass in manually?

renato145:
Ok :)

renato145:
There were some warning I didn't notice before, I just cleaned them.

jph00:
Many thanks!

Issue #3025, opened 2020-11-23
UserWarning when using SaveModelCallback() on after_epoch


muellerzr:
Started from this thread: https://forums.fast.ai/t/userwarning-when-using-savemodelcallback/82445

When `SaveModelCallback` hits `after_epoch` it will try to load in a state without an optimizer (as it's not saved) raising a flag in `learn.load`

PR #3026, opened 2020-11-23
Make Learner.load respect with_opt and fix SaveModelCallback loading issue

muellerzr:
Fixes https://github.com/fastai/fastai/issues/3025

When investigating the solution I noticed we had a sneaky bug in learn.load that was also impacting SaveModelCallback. In SaveModelCallback we don't save the optimizer state, so first thing we had to do was adjust the `after_fit` to pull `with_opt`. This lead to a minor rabbit hole as `Learner.load` doesn't actually use the `with_opt` parameter that's passed with it. This PR fixes this


jph00:
Good job @muellerzr !

PR #3028, opened 2020-11-23
tests(wandb): use a temporary directory

borisdayma:
This will prevent from creating a `wandb` folder when running tests.


jph00:
Perfect :)

PR #3030, opened 2020-11-25
Fix BnFreeze bug. 

marii-moe:
Moved BnFreeze after begin_train in TrainEvalCallback. This now makes the slow test pass, before it did not pass, though we did not notice it because it was a slow test. I am adding this because I am currently working on fp16 related changes, and this was blocking me, those will be in future pull requests.

TrainEvalCallback should truthfully run before everything else, but there is no current way to do this easily, as only `towards_end` exists and no concept of numerical order exists. Issue for this created here: https://github.com/fastai/fastcore/issues/227

### DIFF
--- a/fastai/callback/training.py
+++ b/fastai/callback/training.py
@@ -44,6 +44,7 @@ def set_bn_eval(m:nn.Module, use_eval=True)->None:
         set_bn_eval(l)
 
 class BnFreeze(Callback):
+    run_after=TrainEvalCallback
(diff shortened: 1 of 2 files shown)


marii-moe:
This is the offending code in `TrainEvalCallback` that makes this not work: 
```
def before_train(self):
        "Set the model in training mode"
        self.learn.pct_train=self.epoch/self.n_epoch
        self.model.train()
        self.learn.training=True
```

jph00:
Many thanks!

PR #3037, opened 2020-11-26
[Breaking Imports] Refactor MixUp and CutMix into MixHandler

muellerzr:
# **Breaking Changes**

Refactors MixUp and CutMix into a general MixHandler class, with MixUp and CutMix simply implementing a `before_batch` to perform the data augmentation. To use now do `from fastai.callback.mixup import *` to get access to both


jph00:
What's the breaking change(s) exactly?

jph00:
Let's change the module name back to `mixup` from `mixups` so that's one less breaking change

muellerzr:
Was about to say, the imports would be the major shift/break. Sounds good

muellerzr:
@jph00 should be all good to go now!

jph00:
Are there any breaking changes as far as you know?

Have you trained some models before and after this change and confirmed they're at least as accurate as previously? (If not, please do so! :) )

muellerzr:
Yup, I have been using this in a kaggle competition and they are just as accurate, only breaking changes are the imports (which we've fixed to a degree :) )

muellerzr:
@jph00 can also confirm other Mix variants can be adapted easily, just did `Ricap` a moment ago and it trains as I would expect :) (would Ricap also be a good idea to include in this "mix" module? I can do that as a separate PR later)

jph00:
Yes RICAP would be great.

jph00:
Wonderful! :)

muellerzr:
Great! Will definitely make a PR post this then :)

muellerzr:
I'd like to put this on pause just for a few days, Ignacio Oguiza (from the forums) had a chance to look over the code as well and made some recommendations but I won't be able to reflect them in my code for a little bit.

jph00:
@muellerzr any news on this?

muellerzr:
Not quite yet, sorry! Final projects. I’ll be able to look at this over the
weekend.

muellerzr:
@jph00 okay changes have been pushed. Here's what Ignacio suggested:

Regarding CutMix: 
> I went back through the original CutMix paper and lambda can take any value between 0 and 1, not .5 and 1, so I've changed the logic for that. 
> Also if the input shape is [batch size x channels x height x width], then x (which corresponds to w) needs to be applied to the 3rd dimension, and y (H) to the 2nd dimension. This doesn't make any difference with sqaured images, but it'd break with non-squared images.

Regarding my most recent changes:
In order to not have to worry about too much magic and make this adaptable for Mix that use custom losses, I changed up how `MixHandler` can act. We have a `mixup_criterion` which is for `MixUp` and `CutMix`, and if other loss functions are needed then they can directly pass in their custom ones as `loss_func`

jph00:
Thankyou :)

PR #3040, opened 2020-11-26
Fixed lr too high for gradient accumulation fixed #3023

### LINKED ISSUE #3023: Gradient Accumulation causes lower lr to be same as non-gradient accumulation
I am already working on this, will update this description as I fix it. We do not scale the loss by number of batches to accumulate, so we end up with too large of gradients when doing gradient accumulation. 

More details here: https://github.com/fastai/fastai2/pull/194

marii-moe:
We were not dividing by the number of batches to accumulate, so this effectively increasing the learning rate. Added test to make sure this is fixed. I think this thread was lost when fastai/fastai2 got moved to fastai/fastai: https://github.com/fastai/fastai2/pull/194

fixes #3023

### DIFF
--- a/fastai/callback/training.py
+++ b/fastai/callback/training.py
@@ -20,11 +20,12 @@ class ShortEpochCallback(Callback):
 class GradientAccumulation(Callback):
     "Accumulate gradients before updating weights"
     toward_end,run_before=True,MixedPrecision
-
     def __init__(self, n_acc=32): store_attr('n_acc')
     def before_fit(self): self.count=0
-
+    def after_loss(self):
+        if self.training: self.learn.loss /= self.n_acc/find_bs(self.learn.yb)
     def after_backward(self):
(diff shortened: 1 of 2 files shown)


jph00:
Nice one!

muellerzr:
I know this is fixed and merged now, but I'm hesitant to use it. I was seeing better results with the old implementation of the high LR and worse results here. Do you have any advice @marii-moe as to how I should adapt my old LR to work with this new adjustment?

marii-moe:
> 
> 
> I know this is fixed and merged now, but I'm hesitant to use it. I was seeing better results with the old implementation of the high LR and got worse results here. Do you have any advice @marii-moe as to how I should adapt my old LR to work with this new adjustment?

I am not familiar with your particular example, but you should get approximately the same results if you set your learning rate like so: 
new_lr = old_lr*n_acc/bs

PR #2900, opened 2020-10-23
lr_find() fails with torch.nn.modules.module.ModuleAttributeError: 'ModuleList' object has no attribute 'grad'

dreamflasher:
Code that has been working with fastai1:

```
model = EfficientNet.from_pretrained(base_arch, num_classes=num_classes)
learn = Learner(data, model, metrics=metrics, loss_func=loss_func, path=tempfile.mkdtemp(),
                        splitter=lambda model: [model._conv_stem, model._blocks, model._conv_head])
learn.lr_find()
```
(in fastai1 there was no splitter, but one would call learn.split() with the list directly)

Now fails with fastai2:

```
  File "/home/df/git/mitl/mitlmodels/ml_utils.py", line 251, in find_lr
    learn.lr_find()
  File "/home/df/.local/lib/python3.8/site-packages/fastai/callback/schedule.py", line 228, in lr_find
    with self.no_logging(): self.fit(n_epoch, cbs=cb)
  File "/home/df/.local/lib/python3.8/site-packages/fastcore/logargs.py", line 56, in _f
    return inst if to_return else f(*args, **kwargs)
  File "/home/df/.local/lib/python3.8/site-packages/fastai/learner.py", line 207, in fit
    self._with_events(self._do_fit, 'fit', CancelFitException, self._end_cleanup)
  File "/home/df/.local/lib/python3.8/site-packages/fastai/learner.py", line 157, in _with_events
    finally:   self(f'after_{event_type}')        ;final()
  File "/home/df/.local/lib/python3.8/site-packages/fastai/learner.py", line 133, in __call__
    def __call__(self, event_name): L(event_name).map(self._call_one)
  File "/home/df/.local/lib/python3.8/site-packages/fastcore/foundation.py", line 345, in map
    def map(self, f, *args, gen=False, **kwargs): return self._new(map_ex(self, f, *args, gen=gen, **kwargs))
  File "/home/df/.local/lib/python3.8/site-packages/fastcore/foundation.py", line 203, in map_ex
    return list(res)
  File "/home/df/.local/lib/python3.8/site-packages/fastcore/foundation.py", line 186, in __call__
    return self.fn(*fargs, **kwargs)
  File "/home/df/.local/lib/python3.8/site-packages/fastai/learner.py", line 137, in _call_one
    [cb(event_name) for cb in sort_by_run(self.cbs)]
  File "/home/df/.local/lib/python3.8/site-packages/fastai/learner.py", line 137, in <listcomp>
[...]

PR #3042, opened 2020-11-27
Skip modules without requires_grad and data attributes

dreamflasher:
This fixes two exceptions when migrating code from fastai1 to fastai2:

```
torch.nn.modules.module.ModuleAttributeError: 'ModuleList' object has no attribute 'requires_grad'
torch.nn.modules.module.ModuleAttributeError: 'Conv2dStaticSamePadding' object has no attribute 'data'
```

Comparable to this PR: https://github.com/fastai/fastai/pull/2900


jph00:
Many thanks.

PR #3043, opened 2020-11-27
Fix captum

### LINKED ISSUE #3016: captum module fails after upgrade to captum v0.3.0
Please confirm you have the latest versions of fastai, fastcore, fastscript, and nbdev prior to reporting a bug (delete one): YES.

**Describe the bug**
The [captum nb](https://github.com/fastai/fastai/blob/master/nbs/73_callback.captum.ipynb) fails after upgrading to captum v0.3.0
I found 3 issues:
1. An import problem.

This one can be fixed by changing 
```python
from captum.insights.features import ImageFeature
```
for 
```python
from captum.insights.attr_vis.features import ImageFeature
```

2. A problem with `NoiseTunnel`. It seem to be related to how `.requiresed_grad_()` works inside `torch.no_grad()` on `Tensor` sub classes. For example, if you run the following code:
```python
x = TensorBase(torch.rand(4,3,16,16))
with torch.no_grad():
    print('x.requires_grad', x.requires_grad)
    y = x.requires_grad_()
    z = x
    print('y.requires_grad', y.requires_grad)
    print('z.requires_grad', z.requires_grad)
```
the output is:
```
x.requires_grad False
y.requires_grad False
z.requires_grad True
```

3. A compatibility problem between fastai `Normalize` and how Captum deals with batched inputs. The behavior that is problematic is the following:
```python
x = TensorImage(torch.rand(3,16,16))
x = Normalize.from_stats(*imagenet)(x)
x.shape
```
now, `x.shape` is `(1,3,16,16)`

**Additional context**

Here is the gist of the analysis with a clearer description of the issues. If you think any of these are good enough to make PR, just let me know, and I'll be happy to make them.

https://gist.github.com/ababino/a675d88ca2826c5eb85b16ebd9c8f446

ababino:
partially fixes #3016

The problem that remains is that `NoiseTunnel` metric does not work with Tensor subclasses.

This problem was introduced in [this captum commit](https://github.com/pytorch/captum/commit/ee41c68f3f94d2c41e43b130177f49c49dfec21f) where they put [this line](https://github.com/pytorch/captum/blob/794b55302f16e1e9ed2f10379a6989bc7a5378b4/captum/attr/_core/noise_tunnel.py#L237) within the context `torch.no_grad()`.

For some reason, when an object is a tensor subclass, it does not assign its `requires_grad` property when it is modified in place. For example:
```python
class FooTensor(Tensor): pass
x = FooTensor(torch.rand(4,3,16,16))
with torch.no_grad():
    print('x.requires_grad', x.requires_grad)
    y = x.requires_grad_()
    z = x
    print('y.requires_grad', y.requires_grad)
    print('z.requires_grad', z.requires_grad)
```
returns
```
x.requires_grad False
y.requires_grad False
z.requires_grad True
```

But I expected `y` to have `requires_grad=True`.

A solution to this problem is to subclass a `Tensor` like so:

```python
class FooTensor(Tensor):
    def requires_grad_(self):
        self.requires_grad = True
        return self
```

But that is probably no good solution since I don't know why the original problem happened.


jph00 (changes requested):
Thanks @ababino ! It looks like you accidentally edited a different notebook, which you didn't include in the PR. Could you please provide the correct notebook, then I can merge this? The CI will let you know when everything is synced correctly.

jph00:
Oh also, could you please modify `settings.ini` to update the dev requirements for `captum` to the correct version range?

ababino:
Oh! sorry. sure

ababino:
Now CI passed. 

I also had to install flask, but I have some issues with my conda env, so I am not sure if it was just me. Maybe this would be handled automatically by conda/pip at the moment of installation. It might be necessary to add it to the settings too.

ababino:
I have just checked with a fresh environment, and it worked without installing flask manually. We don't need that change.

jph00:
Thanks so much!

PR #3049, opened 2020-12-02
Fixes loss scale logged as part of training loss fixes #3048

### LINKED ISSUE #3048: Gradient Accumulation + Mixed Precision shows artificially high training loss
Example of issue on forum: https://forums.fast.ai/t/what-is-the-v2-equivalent-of-accumulatescheduler/66409/13

Please confirm you have the latest versions of fastai, fastcore, fastscript, and nbdev prior to reporting a bug (delete one): YES

**Describe the bug**
The bug occurs when Gradient Accumulation and the MixedPrecision Callback are both used. Gradient Accumulation run before Mixed Precision and causes the `after_backwards` to not be run, meaning that the loss is not unscaled before it is logged. This means that very large losses such as 6000000+ to be logged. 

**To Reproduce**
Run this code: 
```
seed=random.randint(0,2**32-1)

with no_random(seed): 
    db=synth_dbunch(bs=8,n_train=1,n_valid=1,cuda=True)
    learn = synth_learner(data=db)
    learn.fit(1, lr=0.01)
#start without gradient overflow
max_loss_scale=2048.0
with no_random(seed): 
    db=synth_dbunch(bs=1,n_train=8,n_valid=8,cuda=True)
    learn = synth_learner(data=db,cbs=[GradientAccumulation(n_acc=8)])
    learn.to_fp16(max_loss_scale=max_loss_scale)
    learn.fit(1, lr=0.01)
```
The training loss will be very high, 5000+ for fp16. fp32 will be reasonable. 

**Expected behavior**
Similar training loss between fp32 and fp16 versions. <2 difference in loss. 

**Additional context**
I have already gotten a fix, bill be submitting it in pull request soon.

marii-moe:
More info here: #3048

Gradient Accumulation was canceling the batch, but MixedPrecision was leaving loss scaled, so that the training loss logged for the user was VERY high. Moved the unscaling of the batch to `after_batch`

Fixes #3048


jph00:
Very cool!

Issue #2814, opened 2020-09-17
Having to override the pickle default behavior when learner has custom classes & functions


ecatkins:
**Describe the bug**

FYI this is an issue I have observed across `fastai1` & `fastai2` when trying to deploy a model, see this [forum post](https://forums.fast.ai/t/error-loading-saved-model-with-custom-loss-function/37627). If you train your model using custom methods & classes, you have to define these classes again when deploying your model. However the problematic behavior I have experienced is that you then must call `load_learner` in your top level `__main__` module, otherwise you will still receive the error:

```
AttributeError: Can't get attribute 'custom_function' on <module '__main__' from 'my_script.py'>
```

This initially proved inconvenient (but not insurmountable) given the way some of projects were structured. However the use case I found where it causes a real issue, was trying to deploy one of my applications using `celery` - in this case, by definition you cannot call `load_learner` in the `__main__` module. It appears other people might have run into [similar issues](https://forums.fast.ai/t/loading-databunch-and-running-the-file-fails-with-celery/71357).

I found a workaround by overriding the default `pickle` behavior that gets passed to `load_learner`

```
import imp, sys
pickle2 = imp.load_module('pickle2', *imp.find_module('pickle'))

class CustomUnpickler(pickle2.Unpickler):

    def find_class(self, module, name):
        try:
            return super().find_class(__name__, name)
        except AttributeError:
            return super().find_class(module, name)

pickle2.Unpickler = CustomUnpickler

def load_learner(fname, cpu=True):
    "Load a `Learner` object in `fname`, optionally putting it on the `cpu`"
    distrib_barrier()
    res = torch.load(fname, map_location='cpu' if cpu else None, pickle_module=pickle2)
    if hasattr(res, 'to_fp32'): res = res.to_fp32()
    if cpu: res.dls.cpu()
    return res

```

Is there an easier solution to this problem I am missing? If not, is this something that could be built into `fastai`?

PR #3052, opened 2020-12-03
expose pickle_module in learn.export

dreamflasher:
Expose pickle_module in learn.export in order to allow, eg. dill for exporting models, because pickle fails when lambdas are used in models. 

More details: https://stackoverflow.com/questions/57057022/learner-export-problem-can-t-pickle-function-self-defined-function-at-0x

Fixes: https://github.com/fastai/fastai/issues/2814


jph00:
Good idea!

dreamflasher:
Thank you for the quick merges!

Issue #3045, opened 2020-11-29
Feature Request: Update the Albumentations tutorial to show training and validation pipelines


muellerzr:
I'm leaving this open to myself so I remember to get to it (won't have the time to really sit down and do it for a bit). I've refactored the code in the tutorial to make use of separate training and validation pipelines, and I also want to simplify the tutorial in some places where possible so beginners can learn how to do it without much struggle. @jph00 you can assign this to me (if possible?) but this is something on the TODO I want to make sure doesn't get swept under the rug with finals

PR #3058, opened 2020-12-07
Improve albumentations tutorial

muellerzr:
Improves the albumentations tutorial by showing integrations with `DisplayedTransform` and `RandTransform`, along with showing individual training pipelines for the validation and training transforms. These transforms can be summized below:

```python
class AlbumentationsTransform(DisplayedTransform):
    split_idx,order=0,2
    def __init__(self, train_aug): store_attr()
    
    def encodes(self, img: PILImage):
        aug_img = self.train_aug(image=np.array(img))['image']
        return PILImage.create(aug_img)

class AlbumentationsTransform(RandTransform):
    "A transform handler for multiple `Albumentation` transforms"
    split_idx,order=None,2
    def __init__(self, train_aug, valid_aug): store_attr()
    
    def before_call(self, b, split_idx):
        self.idx = split_idx
    
    def encodes(self, img: PILImage):
        if self.idx == 0:
            aug_img = self.train_aug(image=np.array(img))['image']
        else:
            aug_img = self.valid_aug(image=np.array(img))['image']
        return PILImage.create(aug_img)
```

Fulfills feature request https://github.com/fastai/fastai/issues/3045


jph00:
This is great!

PR #3065, opened 2020-12-09
Update conda env file to torch 1.7

tcapelle:
Maybe we should make this file automatic after the `settings.ini`?


jph00:
Thanks. I'm not even sure why we have an env file actually!...

tcapelle:
why not, it is not the preferred way of installing fastai?

jph00:
The docs show the preferred way.

Issue #3066, opened 2020-12-09
wandb integration failing with latest wandb library


dreamflasher:
**Describe the bug**
wandb introduced a breaking change with their library, when calling `wandb.log()` – previously this would not increase the step, but now it does. Currently the integration increases this counter on the first call, and consecutive calls fail because the provided step (not increased) does not match with the increased step counter.

**To Reproduce**
Use the WandbCallback

**Error with full stack trace**

Place between these lines with triple backticks:

```
wandb: WARNING Step must only increase in log calls.  Step 6422 < 6423; dropping {'epoch': 20}.
```

**Additional context**
I recommend to remove the internal step counter and let wandb take care of that. I'll propose a PR asap.

borisdayma:
Hi @dreamflasher ,

Do you have any reproducible code you could share so that we can see the issue?
Providing the step explicitly ensures we associate validation and training data to the correct step.

With the proposed PR, `wandb.step` would increase after the end of each training batch. It means that validation metrics would be associated with the next step.

If you log custom data, the recommended way is to create `wandb_process` with the new `@typedispatch` as shown in the notebook (maybe we can better document it).

Otherwise you would just use `commit=False` in your own calls to `wandb.log`.

Let me know if it works for you.

dreamflasher:
@borisdayma The problem is that the current integration fails to report half of the result, because of: https://github.com/fastai/fastai/issues/3066

dreamflasher:
Thanks a lot! This is happening in a bigger integrated project, I'll try to distill a simpler piece of code for reproduction. In my code I always call wandb.log with `commit=False`.

dreamflasher:
@borisdayma I wasn't able to create a concise example, but I believe I found the root cause – actually what I already wrote in the initial message – that you changed the wandb.log api (always increase), but at four places in the code it shouldn't be increased.

Here's the PR: https://github.com/fastai/fastai/pull/3080

Does this make sense to you and is this the expected behaviour?

borisdayma:
I'm still surprised it would solve your problem:
* `commit=True` submits metrics as well as previously uncommitted metrics (for example when using `commit=False` or `step`), then increments the wandb step. When `False` it just adds the metrics to dict of uncommitted values for future commit.
* `step` gives direct access to wandb step and prevents from logging at a previous step. When logging at a future step, it will commit all previous steps, and not commit only the current step.

I'm thinking that there is a mismatch between the logging from the callback and your custom logging.

You could:
[...]

PR #3067, opened 2020-12-10
Leave _wandb_step count to wandb

dreamflasher:
Instead of counting by ourselves, leave it to wandb, in order to avoid us counting incorrectly.
Needed, because wandb changed their API/log behaviour.
Fixes https://github.com/fastai/fastai/issues/3066


borisdayma:
I don't think current behavior should be changed but let's discuss it in the issue.

dreamflasher:
Let's close this PR for now.

PR #3073, opened 2020-12-14
Add `from_epoch` on schedulers

renato145:
Sometimes I need to run long trains and can't do that on a single session. When I resume the training I would like to keep the values of schedulers like `fit_one_cycle` instead of restarting them.

To handle this issue I added the parameter `from_epoch` on `ParamScheduler`, with this is possible to resume a training process:
```python
learn.fit_one_cycle(10, from_epoch=7)
```


jph00:
Thanks for the PR. This is touching on too many files for my liking, so I'd rather not go with this solution.

PR #3075, opened 2020-12-14
Update captum.py

genime:
change dls to self.dls


jph00:
We use nbdev, so this needs to be updated in the notebook. Feel free open reopen this PR or open a new one if you have a chance to do this.

Issue #3082, opened 2020-12-18
Transformers Notebook Tutorial: CUDA out of memory Error


c1505:
Please confirm you have the latest versions of fastai, fastcore, fastscript, and nbdev prior to reporting a bug (delete one): YES

**Describe the bug**
Running this notebook https://github.com/fastai/fastai/blob/master/nbs/39_tutorial.transformers.ipynb in google colab, I get an out of memory error when code is executed `learn.lr_find()`
`RuntimeError: CUDA out of memory. Tried to allocate 786.00 MiB (GPU 0; 15.90 GiB total capacity; 14.30 GiB already allocated; 599.06 MiB free; 14.40 GiB reserved in total by PyTorch)`

**To Reproduce**
1. Open notebook in google colab pro with a GPU runtime
2. Run the notebook

**Expected behavior**
The notebook to run successfully on a typical deep learning machine.  It seems unlikely that most people will be working wit a GPU that has more than 16 GB of memory.

**Error with full stack trace**

Place between these lines with triple backticks:

```
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-31-d81c6bd29d71> in <module>()
----> 1 learn.lr_find()

18 frames
/usr/local/lib/python3.6/dist-packages/torch/nn/functional.py in linear(input, weight, bias)
   1690         ret = torch.addmm(bias, input, weight.t())
   1691     else:
-> 1692         output = input.matmul(weight.t())
   1693         if bias is not None:
   1694             output += bias

RuntimeError: CUDA out of memory. Tried to allocate 786.00 MiB (GPU 0; 15.90 GiB total capacity; 14.30 GiB already allocated; 599.06 MiB free; 14.40 GiB reserved in total by PyTorch)
```

**Additional context**

c1505:
It does run if I set the batch size to 4.  Perplexity was lower with that setting and fit_one_cycle than with a batch size of 8.  New to this so don't really understand that metric that well. 
Would a batch size of 4 be a better default so it runs on GPUs that don't have an immense amount of memory ?

c1505:
PR merged.  Closing.

PR #3083, opened 2020-12-18
Decrease default batch size for Transformers Notebook Tutorial 

c1505:
Running the notebook in google colab as is with 16GB results in an out of memory error. Changing batch size from 8 to 4 fixes it. 
See https://github.com/fastai/fastai/issues/3082


jph00:
Thanks.

PR #3098, opened 2020-12-24
adds std argument to Embedding layer init

arampacha:
`torch.nn.init` has `trunc_normal_` which is slightly different from one currently used in fastai. Here is a [gist with details](https://gist.github.com/arampacha/1689fd1bd5dd7f29bc76d4ba76226f9c).


jph00:
Many thanks!

PR #3100, opened 2020-12-24
Clean up MixUp docs

muellerzr:


jph00:
FYI @muellerzr if you look at the commit here something is adding a bunch of empty tags, which is going to make conflicts more likely, so you might want to remove whatever that is.

Issue #3111, opened 2020-12-28
unet_learner will break if any model params are passed in 


muellerzr:
As the title states, `unet_learner` will break if anything is passed in such as `act_cls` or `sa`, valid parameters for the model. Minimal example:

```python
from fastai.vision.all import *

path = untar_data(URLs.CAMVID_TINY)
dls = SegmentationDataLoaders.from_label_func(
    path, bs=8, fnames = get_image_files(path/"images"),
    label_func = lambda o: path/'labels'/f'{o.stem}_P{o.suffix}',
    codes = np.loadtxt(path/'codes.txt', dtype=str)
)
learn = unet_learner(dls, resnet34, y_range=(0,1))
test_eq(learn.cbs, None) # this should be failing!
```

muellerzr:
Issue was caused by a breaking change in `unet_learner` that was fixed in `cnn_learner` in regards to `store_attr`

PR #3112, opened 2020-12-29
Fix unet_learner store_attr bug

muellerzr:
Fixes https://github.com/fastai/fastai/issues/3111

Adjusted `unet_learner` to properly use `store_attr` similar to `cnn_learner`. Also included a test.

(cc @jph00)


jph00:
Thank you :)

Issue #3113, opened 2020-12-29
create_head does not double the input size when `concat_pool=True`


Ankur-singh:
Please confirm you have the latest versions of fastai, fastcore, and nbdev prior to reporting a bug : YES (version 2.1.10) 
- fastai - 2.1.10
- fastcore -  1.3.13

**Describe the bug**
![create_head](https://user-images.githubusercontent.com/14893062/103296117-6f7d7f00-4a1b-11eb-9691-58624cdf046a.png)

The `create_head` function in *fastai/vision/learner.py* does not double the input size (i.e. `nf` ) when `concat_pool=True`

Whenever `concat_pool` is set to `True`, ` AdaptiveConcatPool2d` layer is used. ` AdaptiveConcatPool2d` layer returns an output tensor which is doubles the size of the input tensor. The next `LinBnDrop` layer uses the output tensor as input. But the sizes don't align. 

**To Reproduce**
Steps to reproduce the behavior, run the following code:
```python
from fastai.vision.all import *
head = create_head(2048, 11)
x = torch.randn(4, 2048, 8, 8)
head(x)
```

**Expected behavior**
`nf` value should be multiplied by 2 whenever `concat_pool=True`.  Adding a single line at the beginning of the function should do the job.

The final function will look something like this.

```python
def create_head(nf, n_out, lin_ftrs=None, ps=0.5, concat_pool=True, bn_final=False, lin_first=False, y_range=None):
    "Model head that takes `nf` features, runs through `lin_ftrs`, and out `n_out` classes."
    nf = 2*nf if concat_pool else nf  # update
    lin_ftrs = [nf, 512, n_out] if lin_ftrs is None else [nf] + lin_ftrs + [n_out]
    ps = L(ps)
    if len(ps) == 1: ps = [ps[0]/2] * (len(lin_ftrs)-2) + ps
    actns = [nn.ReLU(inplace=True)] * (len(lin_ftrs)-2) + [None]
    pool = AdaptiveConcatPool2d() if concat_pool else nn.AdaptiveAvgPool2d(1)
    layers = [pool, Flatten()]
    if lin_first: layers.append(nn.Dropout(ps.pop(0)))
    for ni,no,p,actn in zip(lin_ftrs[:-1], lin_ftrs[1:], ps, actns):
        layers += LinBnDrop(ni, no, bn=True, p=p, act=actn, lin_first=lin_first)
    if lin_first: layers.append(nn.Linear(lin_ftrs[-2], n_out))
[...]

PR #3115, opened 2020-12-30
Make create_head more intuitive with concat_pool + tests

muellerzr:
Fulfills https://github.com/fastai/fastai/issues/3113

This PR makes `create_head` a bit more intuitive on the layers being used, and adds a few tests to ensure that `concat_pool` is still being respected/behaving properly


jph00:
Thanks @muellerzr !

PR #3117, opened 2020-12-30
Fix #3114

### LINKED ISSUE #3114: `cont_cat_split` function doesn't work for int64 pandas columns
`cont_cat_split` function doesn't work for int64 pandas columns

**To Reproduce**
1.  Create a column with int64 values
2. Pass it to `cont_cat_split` function
3. Get the column as continuous no matter what `max_card` is

I think the problem comes from the comparison to `int` which is different from `dtype('int64')`
```python
if df[label].dtype == int and df[label].unique().shape[0] > max_card or df[label].dtype == float:
            cont_names.append(label)
```

gkuwanto:
In #3114 it is said that `cont_cat_split` does not work for a certain data types. Changed The code to use numpy's `issubdtype` to check for integers and floats

**To Reproduce**
```python
from fastai.tabular.core import *
import pandas as pd
df = pd.DataFrame({'f32': pd.Series(range(21),dtype='float32'), 'f64': pd.Series(range(21),dtype='float64')})
print(cont_cat_split(df))
```
**Expected Behaviour**
Both of the column names should be categorized as continuous


jph00:
Thanks!

PR #3118, opened 2020-12-31
added wd argument

tmabraham:
As requested by @jph00 (in Discord), I have created a quick PR to add weight decay as an argument for the Imagenette script. Let me know if there are any problems!


jph00:
Thanks!

PR #3122, opened 2021-01-04
Fixes issue with tabular procs storing extra dataframe in memory

muellerzr:
Fixes https://github.com/fastai/fastai/issues/3120


fastai/tabular/core.py
```diff
@@ -270,7 +270,8 @@ def decodes(self, to:Tabular):
 @Normalize
 def setups(self, to:Tabular):
     store_attr(means=dict(getattr(to, 'train', to).conts.mean()),
-               stds=dict(getattr(to, 'train', to).conts.std(ddof=0)+1e-7))
+               stds=dict(getattr(to, 'train', to).conts.std(ddof=0)+1e-7),
+               but='to')
```
jph00, line 274: Move the `but=`s to the first line of the function - no need to use xtra vertical space for that little thing! :)

muellerzr:
@jph00 this is good to go now. It seems to also have picked up some other notebooks needing to be cleaned with the \r from nbdev

PR #3124, opened 2021-01-04
Timedistributed layer and Image Sequence Tutorial

tcapelle:
This module is equivalent to [Keras TimeDistributed Layer](https://keras.io/api/layers/recurrent_layers/time_distributed/). This wrapper allows to apply a layer to every temporal slice of an input. By default it is assumed the time axis (`tdim`) is the 1st one (the one after the batch size). A typical usage would be to encode a sequence of images using an image encoder.

This PR also adds a nice tutorial on adapting fastai to work on image squences, it is inspired from the siamese tutorial.
- UCR101 dataset setup
- Simple model baseline
- Image/LSTM model
- Using a Transformer based arch


tcapelle:
I am currently making a tutorial notebook.

jph00:
@tcapelle I'll wait until you've done the tutorial notebook so I can see which bits of that we might want to merge into this nb.

tcapelle:
what do you think about moving all tutorials to one folder? So users can submit more tutorials.

muellerzr:
I believe that's what examples/ is for

tcapelle:
My english is not excellent but it is a first version. It may be better to put this notebooks elsewhere (examples??). Maybe move all tutorials there also?

hamelsmu:
@tcapelle is this PR ready to review?  I see that it is still marked as draft

tcapelle:
Not yet, let me add one example of the new transformer archs that have been recently invented (TimesFormer, ViVit, etc)

jph00:
Sorry it took a while to merge this @tcapelle - it looks really great! :D

tcapelle:
Thanks Jeremy, It may be a good idea to move the notebooks out of nbs to not make it run on the tests.

muellerzr:
@tcapelle tutorials with all_slow flag won't run on github pushes (as it only runs `nbdev_test_nbs`. I'd still say we want them running with the --slow flag, as this way we know if a tutorial broke :)

PR #3129, opened 2021-01-07
fix wandb off-by-one

dreamflasher:
fixes both issues in https://github.com/fastai/fastai/issues/3066

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 <!-- do not remove -->
 
+## 2.2.2
+
+### Bugs Squashed
+
+- tensorboard and wandb can not access `smooth_loss` ([#3131](https://github.com/fastai/fastai/issues/3131))
+
+
 ## 2.2.0
 ### Breaking Changes
 


jph00:
I fixed the smooth_loss issue this morning by setting the callback order. Feel free to reopen this or open a new PR if there's still an unresolved issue.

dreamflasher:
@jph00 I can't reopen, would you be so kind? This fixes two issues – thanks for fixing one, I'd like to adjust this PR so that it fixes the other issue I linked to, thanks!

jph00:
Is this ready now @dreamflasher ?

borisdayma:
Let me do a test.
I'm scared of a side effect which will be that last step is uncommitted in colab until people call `wandb.finish()` (which is typically never done). This is not needed in scripts as it will be automatically called when it finishes.

fastai/callback/wandb.py
```diff
@@ -120,7 +120,7 @@ def after_fit(self):
                 log_model(self.save_model.last_saved_path, metadata=metadata)
         self.run = True
         if self.log_preds: self.remove_cb(FetchPredsCallback)
-        wandb.log({}) # ensure sync of last step
+        wandb.log({}, step=self._wandb_step) # ensure sync of last step
```
borisdayma, line 123: I believe this would be equivalent to just removing that line.

borisdayma: Maybe the solution should be to do:
```
# ensure sync of last step
wandb.log({})
self._wandb_step += 1
```

dreamflasher: @borisdayma can you run all three variants in your test suite?

dreamflasher: @borisdayma Looks like a reset to upstream, closes an open PR. In any case: I just tested the your last suggestion and this works!

PR #3136, opened 2021-01-07
from_dicoms update

asvcode:
Currently `from_dicoms` does not allow you to choose the window width and level.  This change allows you to specify the window parameters when calling `from_dicoms`

Currently this is used to create the dataframe:
`pd.DataFrame.from_dicoms(items)`

You cannot specify a window without updating the code in `as_dict`.  With this update you can now call the window parameters directly.

`pd.DataFrame.from_dicoms(items, window=dicom_windows.brain)`


jph00:
Can you please run `nbdev_build_lib` with the latest nbdev version?

asvcode:
Updated to the latest version, hopefully all good now. Thanks

fastai/medical/imaging.py
```diff
@@ -381,7 +381,7 @@ def as_dict(self:DcmDataset, px_summ=True, window=dicom_windows.brain):
     return res
 
 # Cell
-def _dcm2dict(fn, **kwargs): return fn.dcmread().as_dict(**kwargs)
+def _dcm2dict(fn,window,**kwargs): return fn.dcmread().as_dict(window=window, **kwargs)
```
jph00, line 384: should `window` have a default argument?

asvcode: I used `window` because `as_dict` uses window so thought to keep it consistent.

jph00: So you'd need to have the same thing here, right? Otherwise `_dcm2dict` can't be called without a `window` param

asvcode: To be honest that is what I thought but this works.  Out of the box calling:

`dicom_dataframe = pd.DataFrame.from_dicoms(items)`  - works.

`dicom_dataframe = pd.DataFrame.from_dicoms(items, window=dicom_windows.mediastinum)` - raises an error that window is not defined.

Updating to:


`@patch
def _dcm2dict(fn, window, **kwargs): return fn.dcmread().as_dict(window=window, **kwargs)`

Now this seems to work:
`dicom_dataframe = pd.DataFrame.from_dicoms(items, window=dicom_windows.mediastinum)`

Ill look more into it but I made sure that the `img_pct_window` values were updating depending on the window.

dict



dict2

jph00: Is there a reason you don't want to add the default param to this function? It would save some time for both of us, unless there's a reason you think it's a bad idea. I'd love get this PR merged

asvcode: Just trying to minimize the code :), will fix asap thank you

asvcode:
I made the change and I also changed

`dicom_dataframe.head(2).T.head(8)`

to

`dicom_dataframe.head(2).T.tail(5)`

so you can see how the window choice affects the `img_pct_window`

jph00:
Thanks!

PR #3138, opened 2021-01-08
Make TabularProcs consistant, clean up Tabular code

muellerzr:
I had some issues with the other repo, so I just redo'd it. 

Cleans up the tabular code to make it more consistent with the rest of the module. Changes include:

`_TabIloc` now uses `store_attr()` rather than `self.to = to`

`Categorify` now uses typedispatching for the `encodes` and `decodes`

`FillMissing` now uses `to` rather than `dsets` in the `setups` to be consistent with the rest of the `procs`. Along with this it also now uses type-dispatch for the setups

This _could_ be considered a breaking change with the `FillMissing`? 

@jph00 otherwise good to go for this minor adjustment


muellerzr:
@jph00 this one is fine when you are ready to get to it

muellerzr:
@jph00 this one looks ready to review without any conflicts :)

fastai/tabular/core.py
```diff
@@ -130,7 +130,7 @@ def df_shrink(df, skip=[], obj2cat=True, int2uint=False):
 # Cell
 class _TabIloc:
     "Get/set rows by iloc and cols by name"
-    def __init__(self,to): self.to = to
+    def __init__(self,to): store_attr()
```
jph00, line 133: not sure the point of this change?

fastai/tabular/core.py
```diff
@@ -238,11 +238,11 @@ def _decode_cats(voc, c): return c.map(dict(enumerate(voc[c.name].items)))
 class Categorify(TabularProc):
     "Transform the categorical variables to something similar to `pd.Categorical`"
     order = 1
-    def setups(self, to):
+    def setups(self, to:Tabular):
```
jph00, line 241: What's the purpose of the typing here? Do we need typedispatch for these classes?

jph00 (changes requested):
Not sure if I understand the benefits of these changes - can you provide more explanation please?

fastai/tabular/core.py
muellerzr: They're not, we don't need type dispatching but I felt like it helps people understand that tabular's get thrown in there (I can of course remove them though!)

fastai/tabular/core.py
muellerzr: Yeah thinking back on this, `store_attr()` is really only useful when it's 2+ params. Will revert back

PR #3139, opened 2021-01-08
add `_eq_` support to TensorImageBase (fix #3130)

### LINKED ISSUE #3130: Issues with fastai Tensor subclasses & equality operation
Hi,

I think this issue is known based on [Zachary's response on this thread](https://forums.fast.ai/t/chapter-13-learner-error/83114), but I didn't see a github issue for it explicitly

Versions:
fastai : 2.2.1
torch: 1.7.1
fastcore: 1.3.16

Reproducible example:
```
import numpy as np
from fastai.torch_core import TensorImage
from fastai.torch_core import TensorMask

size = (3, 10, 10)
ti = TensorImage(np.random.uniform(size=size))
tm = TensorMask(np.random.choice([0,1], size=size))

ti == tm
```

Error:
```
<ipython-input-27-f607939df9b1> in <module>
      6 tm = TensorMask(np.random.choice([0,1], size=size))
      7 
----> 8 ti == tm

~/.local/share/virtualenvs/fastai_modeling-wXHhCDlU/lib64/python3.7/site-packages/torch/tensor.py in wrapped(*args, **kwargs)
     23         from torch.overrides import has_torch_function, handle_torch_function
     24         if not all(type(t) is Tensor for t in args) and has_torch_function(args):
---> 25             return handle_torch_function(wrapped, args, *args, **kwargs)
     26         try:
     27             return f(*args, **kwargs)

~/.local/share/virtualenvs/fastai_modeling-wXHhCDlU/lib64/python3.7/site-packages/torch/overrides.py in handle_torch_function(public_api, relevant_args, *args, **kwargs)
   1066     raise TypeError("no implementation found for '{}' on types that implement "
   1067                     '__torch_function__: {}'
-> 1068                     .format(func_name, list(map(type, overloaded_args))))
   1069 
   1070 def has_torch_function(relevant_args: Iterable[Any]) -> bool:

TypeError: no implementation found for 'torch.tensor.eq' on types that implement __torch_function__: [<class 'fastai.torch_core.TensorImage'>, <class 'fastai.torch_core.TensorMask'>]

tcapelle:
This solution is not complete, but fixes the issue reported.
I don't get why `__array_eq__` still does not work.
just for ref: https://pytorch.org/docs/stable/notes/extending.html


jph00:
Thanks!

PR #3144, opened 2021-01-10
Add Focal Loss to avaialable losses

pford221:
Hi,

This PR adds Focal Loss to the codebase. I'm struggling with the diff on  nbs/01a_losses.ipynb because `log_args` doesn't seem to be part of fastcore anymore? Therefore, I can't run the notebook to minimize the diffs.


jph00:
Why do you need `log_args`? Do you have the latest versions of nbdev and fastcore?

Could you please try to reduce the vertical space your code is using, e.g. by using the ternary operator for the return, and using a wider line width?

I haven't looked closely, but is there an option to include focal loss in the flattened cross entropy loss directly?

pford221:
I do not need `log_args` but the notebook that I was updating it has `log_args` in many cells and it is not in my environment which has fastcore 1.3.19. So the notebook is not runnable in the environment which I thought might be needed for the diffs. But I was able to handle the diffs using the github.com UI (VSC doesn't seem to allow me to edit jupyter notebook as json anymore).

I can elongate lines and use ternary operator but it seems excessively long in my IDE:
```
return focal_loss.mean() if self.reduction == 'mean' else focal_loss.sum() if self.reduction == 'mean' else focal_loss
```
I suppose I could use an abbreviation for `focal_loss` to shorten if you'd prefer.

`CrossEntropyLossFlat` doesn't currently support focal loss. Here's what it looks like. 
```
class CrossEntropyLossFlat(BaseLoss):
    "Same as `nn.CrossEntropyLoss`, but flattens input and target."
    y_int = True
    @use_kwargs_dict(keep=True, weight=None, ignore_index=-100, reduction='mean')
    def __init__(self, *args, axis=-1, **kwargs): super().__init__(nn.CrossEntropyLoss, *args, axis=axis, **kwargs)
    def decodes(self, x):    return x.argmax(dim=self.axis)
    def activation(self, x): return F.softmax(x, dim=self.axis)
```

jph00:
Sounds like you might be working from an old version of fastai? log_args isn't used in the current version.

The line you showed doesn't look too long to me - feel free to rename things if you like though.

I know the current flatten CE loss function doesn't have focal loss - I'm wondering, however, if it should be added as a parameter, rather than creating a new loss function (since cross entropy loss is a special case of focal loss).

pford221:
Yes, thank you! I was using a lagged forked version unbeknownst. 

I've fiddled around with trying to combine focal loss and cross entropy loss and I think this will work, but it's not aesthetically pleasing. The challenge is with the `reduction` argument. If focal loss is desired, I believe it has to be cached in another parameter which I've called `self.user_rdc` because `nn.CrossEntropyLoss` needs to be called with `reduction = 'none'` first so that we get back a tensor that is returned can be operated on element-wise.

```
@delegates()
class CrossEntropyLossFlat(BaseLoss):
    "Same as `nn.CrossEntropyLoss`, but flattens input and target."
    y_int = True
    @use_kwargs_dict(keep=True, weight=None, ignore_index=-100, reduction='mean')
    def __init__(self, *args, gamma=None, axis=-1, **kwargs):
        self.gamma = gamma
        if self.gamma:
            if 'reduction' in kwargs: self.user_rdc = kwargs.pop('reduction')
            super().__init__(nn.CrossEntropyLoss, *args, reduction='none', axis=axis, **kwargs)
        else:
            super().__init__(nn.CrossEntropyLoss, *args, axis=axis, **kwargs)
    def __call__(self, inp, targ, **kwargs):
        if self.gamma:
            ce_loss = super().__call__(inp, targ, **kwargs)
            pt = torch.exp(-ce_loss)
            fl_loss = (1-pt)**self.gamma * ce_loss
            return fl_loss.mean() if self.user_rdc == 'mean' else fl_loss.sum() if self.user_rdc == 'sum' else fl_loss
        else:
            return super().__call__(inp, targ, **kwargs)
    def decodes(self, x):    return x.argmax(dim=self.axis)
    def activation(self, x): return F.softmax(x, dim=self.axis)
```

If you could opine on whether this is preferred to having a separate class, or if you think there's a better way, that would be most appreciated.

tcapelle:
I would separate them with a partial or a sub class.
I would also add tests for different cases. (multi label

pford221:
Subclassing is certainly possible. The best way I can think to include is like so:

```
class FocalLossFlat(CrossEntropyLossFlat):
    """
    Same as CrossEntropyLossFlat but with focal paramter, `gamma`. From  Lin et al. in this paper
    Focal loss is introduced by Lin et al. in https://arxiv.org/pdf/1708.02002.pdf. Note the class
    weighting factor in that paper, alpha, can be implemented through pytorch `weight` argument.
    """
    y_int = True
    @use_kwargs_dict(keep=True, weight=None, ignore_index=-100, reduction='mean')
    def __init__(self, *args, gamma=2, axis=-1, **kwargs):
        self.gamma = gamma
        if 'reduction' in kwargs: self.reduce = kwargs.pop('reduction')
        super().__init__(*args, reduction='none', axis=axis, **kwargs)
    def __call__(self, inp, targ, **kwargs):
        ce_loss = super().__call__(inp, targ, **kwargs)
        pt = torch.exp(-ce_loss)
        fl_loss = (1-pt)**self.gamma * ce_loss
        return fl_loss.mean() if self.reduce == 'mean' else fl_loss.sum() if self.reduce == 'sum' else fl_loss
```

jph00:
That looks pretty good! What do you think Patrick?


On Mon, Jan 11, 2021, at 6:24 PM, Patrick Ford wrote:
> 

> Subclassing is certainly possible. The best way I can think to include is like so:

> `class FocalLossFlat(CrossEntropyLossFlat):
    """
    Same as CrossEntropyLossFlat but with focal paramter, `gamma`. From  Lin et al. in this paper
    Focal loss is introduced by Lin et al. in https://arxiv.org/pdf/1708.02002.pdf. Note the class
    weighting factor in that paper, alpha, can be implemented through pytorch `weight` argument.
    """
    y_int = True
    @use_kwargs_dict(keep=True, weight=None, ignore_index=-100, reduction='mean')
    def __init__(self, *args, gamma=2, axis=-1, **kwargs):
        self.gamma = gamma
        if 'reduction' in kwargs: self.reduce = kwargs.pop('reduction')
        super().__init__(*args, reduction='none', axis=axis, **kwargs)
    def __call__(self, inp, targ, **kwargs):
        ce_loss = super().__call__(inp, targ, **kwargs)
        pt = torch.exp(-ce_loss)
        fl_loss = (1-pt)**self.gamma * ce_loss
        return fl_loss.mean() if self.reduce == 'mean' else fl_loss.sum() if self.reduce == 'sum' else fl_loss
`

tcapelle:
Awesome, focal loss is so useful, didn't we had them at some point on V1?

pford221:
Sure, let's go with the subclassing option. At this stage, It's probably easiest to just start a new PR fresh. I will close this one and re-open tonight. Let me know if any concerns with that strategy.

PR #3146, opened 2021-01-11
Fixed MixHandler to use own loss definition when stack_y

MikeWklm:
Hi, 
I think the current implementation of MixHandler has a tiny bug. It does not change the loss function to the one defined by the MixHandler if stack_y is true.


jph00:
Thanks!

PR #3149, opened 2021-01-12
--wks param to train_imagenette.py to show DDP bug w/ n_workers=0

antorsae:


jph00:
Closing this since it seems to be incorporated in #3151

PR #3150, opened 2021-01-12
distributed 0 workers fix

antorsae:
- --wks param to train_imagenette.py to show DDP bug w/ n_workers=0
- fixes distributed training when using 0 workers


jph00:
Closing this since it seems to be incorporated in #3151

PR #3151, opened 2021-01-12
typo to to non native fp16

antorsae:
- --wks param to train_imagenette.py to show DDP bug w/ n_workers=0
- fixes distributed training when using 0 workers
- fix typo:to_to_non_native_fp16 -> to_non_native_fp16


jph00:
Thanks. I'm going to rename `wks` back to what it was, since there's no reason for this breaking change AFAICT.

PR #3152, opened 2021-01-13
FocalLossFlat loss class added as an option

pford221:
This is a simplified PR that replaces #3144 using a subclass approach.


tcapelle:
Nice!
Probably make the dosctring one liner, and put the rest on a markdown cell above.

pford221:
I'm happy to make any changes, documentation or otherwise, but I'll wait for Jeremy to opine so I only have to make them once.

jph00:
Very nice :)

Issue #3164, opened 2021-01-15
'cat' attribute with pandas dataframe: AttributeError: Can only use .cat accessor with a 'category' dtype


dreamflasher:
Please confirm you have the latest versions of fastai, fastcore, and nbdev prior to reporting a bug (delete one): YES 
```python3 -c "import fastai; import torch; import fastcore; print(torch.__version__, fastai.__version__, fastcore.__version__)"```: 1.7.0+cu110 2.2.4 1.3.17

**Describe the bug**
When having a column with the name `cat` and selecting that column with `label_col` results in ```AttributeError: Can only use .cat accessor with a 'category' dtype```

**To Reproduce**
1. Create a CSV with column name cat
2. Point dataloader to this csv and specify label_col="cat"

**Expected behavior**
no exception

**Error with full stack trace**

Place between these lines with triple backticks:

```
Traceback (most recent call last):
  File "/app/main.py", line 178, in run
    training_dataset_path, wandb_conf=wandb_conf, seed=0
  File "/mitl-model/mitlmodels/ml_algos.py", line 45, in train
    export_location=tempfile.mkdtemp(), k=1, seed=seed)
  File "/mitl-model/mitlmodels/model.py", line 339, in train
    valid_pct)
  File "/mitl-model/mitlmodels/model.py", line 134, in load_data
    image_size / 0.875)), y_block=y_block, valid_pct=valid_pct, num_workers=num_workers)
  File "/opt/conda/lib/python3.7/site-packages/fastai/vision/data.py", line 140, in from_df
    return cls.from_dblock(dblock, df, path=path, **kwargs)
  File "/opt/conda/lib/python3.7/site-packages/fastai/data/core.py", line 180, in from_dblock
    return dblock.dataloaders(source, path=path, bs=bs, val_bs=val_bs, shuffle_train=shuffle_train, device=device, **kwargs)
  File "/opt/conda/lib/python3.7/site-packages/fastai/data/block.py", line 113, in dataloaders
    dsets = self.datasets(source, verbose=verbose)
  File "/opt/conda/lib/python3.7/site-packages/fastai/data/block.py", line 110, in datasets
    return Datasets(items, tfms=self._combine_type_tfms(), splits=splits, dl_type=self.dl_type, n_inp=self.n_inp, verbose=verbose)
  File "/opt/conda/lib/python3.7/site-packages/fastai/data/core.py", line 310, in __init__
[...]

PR #3165, opened 2021-01-15
Fix 'cat' attribute with pandas dataframe: AttributeError: Can only use .cat accessor with a 'category' dtype

dreamflasher:
Fixes https://github.com/fastai/fastai/issues/3164


jph00:
Thank you

Issue #3125, opened 2021-01-05
[Beginner] Bring back the Performance Tips and Update for Latest Versions/Ideas


muellerzr:
If anyone would like to tackle this, in the old version we had this documentation:

https://fastai1.fast.ai/performance.html

Which discussed bringing in libjpeg turbo among other libraries. It would be nice to have this back in the documentation and updated where it needs be in case anyone has discovered any performance tricks with the external libs, etc needed with fastai.

It's Markdown is available here for you to look at: 
https://github.com/fastai/fastai1/blob/master/docs/performance.md

In total what this PR should do is update that markdown, move it to `fastai/docs_src/dev/performance.md`, and update the sidebar at `fastai/docs_src/sidebar.json` to include it in with the "Getting started" category

nglillywhite:
Happy to tackle this, @muellerzr thanks for the suggestion and guidance. I'll put together a PR and post back here if that's the best way forward. Let me know if there's anything else.

nglillywhite:
I noticed that check_perf() wasn't migrated into the new repo and so I removed that from the previous performance doc, was this intentional or this function didn't make the cut to the new version of the library?

muellerzr:
@jph00 you may be better at suggesting this. Ideally if you wanted to bring it over, it would reside in the same place that we already have the `show_install`? 

The actual func in v1 is here: https://github.com/fastai/fastai1/blob/master/fastai/utils/collect_env.py#L156

nglillywhite:
Hey @muellerzr , I think this can be closed now. Thanks for the guidance and help with my first PR and contribution to fastai, the experience was really positive (thanks jeremy as well for feedback).

jph00:
Great!

PR #3169, opened 2021-01-17
migrate-performance-tips

nglillywhite:
Migrated the performance tips and tricks from fastai as per https://github.com/fastai/fastai/issues/3125
- Migrated performance.md
- Removed reference to check_perf() which hasn't been migrated but existed in fastai1
- Checked links & Legibility

### DIFF
--- a/docs_src/sidebar.json
+++ b/docs_src/sidebar.json
@@ -1,8 +1,7 @@
 {
     "Getting started": {
         "Overview": "/",
-        "Quick Start": "/quick_start.html",
-        "Performance Tips and Tricks": "/dev/performance.html"
+        "Quick Start": "/quick_start.html"
     },
     "Tutorials": {
         "Overview": "/tutorial.html",
@@ -27,7 +26,8 @@
         "empty2": {
             "Advanced": {
                 "Albumentations Tutorial": "/tutorial.albumentations.html",
-                "Siamese Tutorial": "/tutorial.siamese.html"
+                "Siamese Tutorial": "/tutorial.siamese.html",
+                "Performance Tips and Tricks": "/dev/performance.html"
             }
         },
         "empty3": {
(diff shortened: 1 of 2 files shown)


muellerzr:
@nglillywhite have you verified that all of it works? `fastai` no longer has a `pillow_version` function. You could do something like:

```python
import PIL
print(PIL.__version__)
```

(for instance right now, it will either say 7.0.0 or 7.0.0.post3, with the latter being simd)

nglillywhite:
Thats a good point, let me verify all the commands and work through the doc. I read it for legibility and checked hyperlinks but didn't run the commands myself. I'll check them all and update accordingly, thanks!

nglillywhite:
@muellerzr , you're right on the pillow version syntax. I updated the example to the latest available versions.

I'm on windows so I couldn't verify the path examples already provided in the other snippets but the rest of the installation commands worked pending having the prerequisites installed. Happy to spin up a VM and work through them if thats needed to get the PR through.

muellerzr:
@jph00 I'd say this is ready for your review :) This would close https://github.com/fastai/fastai/issues/3125

docs_src/sidebar.json
```diff
@@ -1,7 +1,8 @@
 {
     "Getting started": {
         "Overview": "/",
-        "Quick Start": "/quick_start.html"
+        "Quick Start": "/quick_start.html",
+        "Performance Tips and Tricks": "/dev/performance.html"
```
jph00, line 5: Please move this to the advanced tutorials sections.

jph00 (changes requested):
Many thanks! Just one request :)

docs_src/sidebar.json
nglillywhite: I've updated the branch but it still mentions changes required so I'm going to resolve this conversation in the hope that cleans it up. Apologies if I've just added more notifications unnecessarily.

PR #3171, opened 2021-01-18
change set_seed default to reproducible=True

tmabraham:
For some reason, in set_seed, the default is that `reproducible=False`. For this reason, training with a set seed will still result in different results. I believe most people would expect the opposite. Now I always use `reproducible=True`, but I figure other people might be confused so I figured it might be best to update fastai's default.

cc: @jph00


jph00:
I think people would be surprised that setting the seed would make it slower, hence it's False by default.

PR #3173, opened 2021-01-18
Fix iterable datasets

jcaw:
`IterableDataset` objects currently don't work when `_one_pass` is called, because this method attempts to index item `0` but `None` would be required for an `IterableDataset`. However, passing `None` is incompatible with indexed datasets. This PR aims to fix the issues. 

The following changes are made:

- Fix `create_item` to be compatible with both iterable and indexed datasets.
- If a sample item is required, index `do_item` with `None`, not `0`. For iterable datasets, this will get the next item, for indexible datasets it will get the first item.
- Subclasses of `DataLoader` are updated to be compatible with this indexing scheme.
- Native PyTorch `IterableDataset`s have a stubbed `__getitem__` method (that just raises an error), so don't just rely on the presence of that method when establishing the `indexed` property - check for `IterableDataset` classes/subclasses too.

While I've tried to familiarise myself with how `_one_pass` is used, I'm a little unsure whether it could be called such that it consumed the first item of the iterator, then the iterator continued to be used by e.g. the training loop (effectively skipping the first item). Is this a problem, or is the iterator always reset by calling `create_batches`?

### DIFF
--- a/fastai/data/core.py
+++ b/fastai/data/core.py
@@ -48,7 +48,7 @@ class TfmdDL(DataLoader):
                 kwargs[nm].setup(self)
 
     def _one_pass(self):
-        b = self.do_batch([self.do_item(0)])
+        b = self.do_batch([self.do_item(None)])
         if self.device is not None: b = to_device(b, self.device)
         its = self.after_batch(b)
         self._n_inp = 1 if not isinstance(its, (list,tuple)) or len(its)==1 else len(its)-1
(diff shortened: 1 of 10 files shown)


jcaw:
Seems like I've pushed some cruft in the notebooks - let me remove that and force-push before this is reviewed.

jph00:
Looks good! - Many thanks :)

PR #3178, opened 2021-01-20
Changed dataloaders arguments to have consistent overrides

### LINKED ISSUE #3133: DataBlock.dataloaders does not support the advertised "shuffle" argument
fastai 2.2.2, fastcore 1.3.16, nbdev 1.1.6

**Describe the bug**
The DataBlock documentation describes the [https://docs.fast.ai/data.block.html#DataBlock.dataloaders](`dataloaders`) function and explicitly says that there is an optional kwarg `shuffle`.  However, using this results in the error:
```
TypeError: fastai.data.core.TfmdDL() got multiple values for keyword argument 'shuffle'
```

**To Reproduce**
Take the notebook for Chapter 2 of the fastai book, and change the cell which reads `dls = bears.dataloaders(path)` to read `dls = bears.dataloaders(path, shuffle=True)` instead.

**Expected behavior**
It should perform shuffling as requested.

```python
<ipython-input-6-cb46ba7faa34> in <module>
----> 1 dls = bears.dataloaders(path, shuffle=True)

~/fast.ai.course/fastai-venv/lib/python3.9/site-packages/fastai/data/block.py in dataloaders(self, source, path, verbose, **kwargs)
    113         dsets = self.datasets(source, verbose=verbose)
    114         kwargs = {**self.dls_kwargs, **kwargs, 'verbose': verbose}
--> 115         return dsets.dataloaders(path=path, after_item=self.item_tfms, after_batch=self.batch_tfms, **kwargs)
    116 
    117     _docs = dict(new="Create a new `DataBlock` with other `item_tfms` and `batch_tfms`",

~/fast.ai.course/fastai-venv/lib/python3.9/site-packages/fastai/data/core.py in dataloaders(self, bs, val_bs, shuffle_train, n, path, dl_type, dl_kwargs, device, **kwargs)
    211         if dl_type is None: dl_type = self._dl_type
    212         drop_last = kwargs.pop('drop_last', shuffle_train)
--> 213         dl = dl_type(self.subset(0), bs=bs, shuffle=shuffle_train, drop_last=drop_last, n=n, device=device,
    214                      **merge(kwargs, dl_kwargs[0]))
    215         dls = [dl] + [dl.new(self.subset(i), bs=(bs if val_bs is None else val_bs), shuffle=False, drop_last=False,

TypeError: fastai.data.core.TfmdDL() got multiple values for keyword argument 'shuffle'
```

**Additional context**
The cause of the error is clear, though the best way to solve it is not as obvious.  The `DataBlock.dataloaders()` method:
https://github.com/fastai/fastai/blob/373050bcbf5d31712340c06316326c8d0403ed76/fastai/data/block.py#L112-L115
just passes most of the keyword arguments through to its `Datasets` object, which in turn is a subclass of `FilteredBase`, so inherits the `dataloaders()` method from there.  This method is as follows:
https://github.com/fastai/fastai/blob/d1cf81c2aacaad4497d1920c468f3c73a3a6fc71/fastai/data/core.py#L207-L217
This has the keyword argument `shuffle_train` (but not shuffle), but on lines 213-214, calls the `dl_type` function with an explicit `shuffle` keyword and then the `kwargs` passed through (merged with `dl_kwargs[0]`).  So if `kwargs` includes `shuffle`, there will be a repeated keyword argument.

I haven't checked whether the same would apply to the other keyword arguments listed on line 213.

One simple way would be to rewrite lines 213-214 as:
```python
        dl = dl_type(self.subset(0),
                     **merge({'bs': bs, 'shuffle': shuffle_train, 'drop_last': drop_last, 'n': n, 'device': device},
                     kwargs, dl_kwargs[0]))
```

marii-moe:
Fixes #3133

### DIFF
--- a/fastai/data/core.py
+++ b/fastai/data/core.py
@@ -176,8 +176,8 @@ class DataLoaders(GetAttr):
         return cls(*[dl_type(d, bs=bs, **k) for d,k in zip(ds, kwargs)], path=path, device=device)
 
     @classmethod
-    def from_dblock(cls, dblock, source, path='.',  bs=64, val_bs=None, shuffle_train=True, device=None, **kwargs):
-        return dblock.dataloaders(source, path=path, bs=bs, val_bs=val_bs, shuffle_train=shuffle_train, device=device, **kwargs)
+    def from_dblock(cls, dblock, source, path='.',  bs=64, val_bs=None, shuffle=True, device=None, **kwargs):
+        return dblock.dataloaders(source, path=path, bs=bs, val_bs=val_bs, shuffle=shuffle, device=device, **kwargs)
 
     _docs=dict(__getitem__="Retrieve `DataLoader` at `i` (`0` is training, `1` is validation)",
                train="Training `DataLoader`",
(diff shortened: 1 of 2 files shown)


jph00:
Thank you! :)

PR #3179, opened 2021-01-21
SegmentationInterpretation Class with plot_top_losses() method

pford221:
Hello,

This is an unsolicited feature PR. I found the `plot_top_losses()` method quite handy for `ClassificationInterpretation` so I had a go at adding it in for a new, but very simple, `SegmentationInterpretation` class.

I'm happy to provide a notebook that shows how it works with CamVid or another dataset if you'd like. Just need guidance on where to commit that notebook.


jph00 (changes requested):
Hey this is nice! :) Could you please add a basic working example? Just take a look through the NBs to see anywhere that we train a segmentation model, and add an example there. If we don't have any examples, please create a new tutorial notebook. There's a CAMVID_TINY or something like that for doing quick examples.

pford221:
I added a couple examples to the tutorials. I went out of my way to make the diffs as small as possible, but the base64 encoded imagery from `.show_batch()` is a diff. I also tried to rebase my commits and squash them to 1 but because I pulled from origin in between the first commit from the other day and today, I couldn't rebase as I'd like so I apologize for the multiple commits.

jph00:
Super!

PR #3181, opened 2021-01-21
Update plot_lr_find in schedule.py to show correct x-axis scale

maxrylander:
Changed so the plot shows correct scale.


jph00:
We want a log scale here.

Issue #3182, opened 2021-01-22
Tensorboard Callback TypeError: TypeError: requires_grad_() takes 1 positional argument but 2 were given


FabianHertwig:
Please confirm you have the latest versions of fastai, fastcore, and nbdev prior to reporting a bug (delete one): YES 
Version is 2.2.5 according to `fastai.__version__`

**Describe the bug**
When adding the Tensorboard Callback to the learner, there is an error during training:

```text
TypeError: requires_grad_() takes 1 positional argument but 2 were given
```

When removing the callback, there is no error.

**To Reproduce**

https://colab.research.google.com/drive/1hN6-dyJVXOYj8r9V88QLaMlcJRAeTDbx?usp=sharing 

```python
#hide
#skip
! [ -e /content ] && pip install -Uqq fastai  # upgrade fastai on colab

from fastai.vision.all import *
from fastai.callback.tensorboard import TensorBoardCallback

path = untar_data(URLs.PETS)

files = get_image_files(path/"images")
len(files)

def label_func(f): return f[0].isupper()

dls = ImageDataLoaders.from_name_func(path, files, label_func, item_tfms=Resize(224))

learn = cnn_learner(dls, resnet34, metrics=error_rate)
learn.add_cb(TensorBoardCallback(log_dir="logs/tensorboard"))

learn.fine_tune(1)

```

**Expected behavior**
Training completes without errors and tensorboard logs are written to disk.

**Error with full stack trace**

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-9-6952b5de7700> in <module>()
      3 learn.add_cb(TensorBoardCallback(log_dir="logs/tensorboard"))
      4 
----> 5 learn.fine_tune(1)

19 frames
/usr/local/lib/python3.6/dist-packages/fastai/callback/schedule.py in fine_tune(self, epochs, base_lr, freeze_epochs, lr_mult, pct_start, div, **kwargs)
    155     "Fine tune with `freeze` for `freeze_epochs` then with `unfreeze` from `epochs` using discriminative LR"
    156     self.freeze()
--> 157     self.fit_one_cycle(freeze_epochs, slice(base_lr), pct_start=0.99, **kwargs)
    158     base_lr /= 2
    159     self.unfreeze()

[...]

PR #3184, opened 2021-01-22
Make requires_grad_ consistent with PyTorch

muellerzr:
Closes https://github.com/fastai/fastai/issues/3182

PyTorch's `requires_grad_` passes in a `requires_grad` param. This is needed when exporting to jit (and potentially other libraries/frameworks) as they want to adjust that param. The default remains `True`, so no code currently in the library is affected (and should be continued to be used as `TensorBase.requires_grad_()`

<s>@jph00 let me know if you would like me to make an adjustment to instances of `requires_grad_()` in the framework to expose the `requires_grad` param inside of it or not. </s>

cc: @Isaac-Flath as it's relevant to jit

### DIFF
--- a/fastai/torch_core.py
+++ b/fastai/torch_core.py
@@ -340,9 +340,9 @@ class TensorBase(Tensor):
         res = self.as_subclass(Tensor).new() if x is None else self.as_subclass(Tensor).new(x)
         return res.as_subclass(cls)
 
-    def requires_grad_(self):
+    def requires_grad_(self, requires_grad=True):
         # Workaround https://github.com/pytorch/pytorch/issues/50219
-        self.requires_grad = True
+        self.requires_grad = requires_grad
         return self
 
 # Cell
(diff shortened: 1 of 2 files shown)


jph00:
> @jph00 <https://github.com/jph00> let me know if you would like me to make an adjustment to instances of `requires_grad_()` in the framework to expose the `requires_grad` param inside of it or not.


What places did you find that might require that?

muellerzr:
> @jph00 <https://github.com/jph00> let me know if you would like me to make an adjustment to instances of `requires_grad_()` in the framework to expose the `requires_grad` param inside of it or not.
> What places did you find that might require that?

jit's `clone_input` (which is called during `trace`) will call a tensor like so:
```python
    148                 a.detach()
    149                 .clone(memory_format=torch.preserve_format)
--> 150                 .requires_grad_(a.requires_grad)
    151             )
```
(this is from https://github.com/fastai/fastai/issues/3182)

Issue #3180, opened 2021-01-21
bug: error on predict() or using interp with resnet and MixUp in FastAI v2.2.5


cogsci2:
Please confirm you have the latest versions of fastai, fastcore, and nbdev prior to reporting a bug (delete one): YES 

**Describe the bug**
Using resnet18 along with the MixUp callback to categorize images on the PETS data results in a post-training error. When using `learn.predict()`,  `pred` is given as a tensor of n category labels with repeats.  `pred_idx`is a tensor of dimension n number of categories whose contents are floating points.  Actually, `pred_idx` is a direct copy of `prob`

**Current Workarounds**
1. Regressing to FastAI v 2.2.2 or 
2. Remove MixUp 

**To Reproduce**
As requested by Zach Mueller, I have made a minimal reproducer to debug with available here: [https://github.com/cogsci2/Pets-Minimal/blob/main/MinimalPets_debug_mixup.ipynb](https://github.com/cogsci2/Pets-Minimal/blob/main/MinimalPets_debug_mixup.ipynb)

1. Use any image categorization model that uses MixUp and Resnet.  
2. After training, use show_results() and plot a confusion matrix.  

![image](https://user-images.githubusercontent.com/66141096/105308715-a6096e00-5b71-11eb-9756-5a56333ad2c6.png)

**Expected behavior**
No error message is given, kernel execution is not halted.

**Error with full stack trace**

Place between these lines with triple backticks:

Error when using plot_confusion_matrix():
```
---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
<ipython-input-10-4eca0711123a> in <module>
      1 interp = ClassificationInterpretation.from_learner(learn)
----> 2 interp.plot_confusion_matrix(figsize=(8,8))

~/miniconda3/envs/learn/lib/python3.8/site-packages/fastai/interpret.py in plot_confusion_matrix(self, normalize, title, cmap, norm_dec, plot_txt, **kwargs)
     68         "Plot the confusion matrix, with `title` and using `cmap`."
     69         # This function is mainly copied from the sklearn docs
---> 70         cm = self.confusion_matrix()
     71         if normalize: cm = cm.astype('float') / cm.sum(axis=1)[:, np.newaxis]
     72         fig = plt.figure(**kwargs)

~/miniconda3/envs/learn/lib/python3.8/site-packages/fastai/interpret.py in confusion_matrix(self)
     60         "Confusion matrix as an `np.ndarray`."
[...]

PR #3186, opened 2021-01-23
Fix MixHandler event triggers

muellerzr:
Closes https://github.com/fastai/fastai/issues/3180

`before_fit` and `after_fit` are always called, even during just inference, so `learn.loss_func` was always being changed. As a result suddenly our predictions and our ground truths were getting incorrectly changed to different types (as the `torch.lerp` was being applied). We should switch those triggers to `before_train` and `after_train`. This did not affect training, only on inference if you passed in anything related to `MixHandler` into `learn.cbs`

cc: @jph00


muellerzr:
Looks as though I missed a recent commit/fix. Will adjust accordingly

muellerzr:
@jph00 good to go now

jph00:
Thank you

Issue #3187, opened 2021-01-25
Neptune callback can not access 'smooth_loss' of the learner.


pcnudde:
The Neptune callback fails on 2.2.5 when trying to access the leaners smooth_loss on this line: https://github.com/fastai/fastai/blob/master/fastai/callback/neptune.py#L52 

The error is “ModuleAttributeError: ‘Sequential’ object has no attribute ‘smooth_loss’", and when inspecting the learner object after the error indeed the smooth_loss is not present. 

I assume this bug is similar to #3131 and caused by the change in the order of the callbacks

PR #3190, opened 2021-01-25
Add Azure ML Callback for experiment tracking

FabianHertwig:
This adds a Callback to track experiments with Azure ML

I basically copied and adapted the Neptune Callback, therefore I ran into the same issue as described in #3187. This should be fixed once I know what the solution is.


fastai/callback/azureml.py
```diff
@@ -0,0 +1,46 @@
...
+from ..basics import *
+from ..learner import Callback
+
+# Cell
+from azureml.core.run import Run
+
+# Cell
+class AzureMLCallback(Callback):
+    "Log losses, metrics, model architecture summary to AzureML"
+
```
jph00, line 16: Add this line here to fix your smooth loss problem:

```python
order = Recorder.order+1
```

jph00 (changes requested):
Many thanks - just one suggestion to fix your bug

fastai/callback/azureml.py
FabianHertwig: Do you mean `self.order`? Just setting order did not work.

FabianHertwig: Ah, I got that wrong. It is set just inside the class and not in a `__init__`. 

Will add the commit in the next minutes.

jph00:
Many thanks.

PR #3192, opened 2021-01-26
[Don't Merge] force num_workers=0 on Windows

mszhanyi:
It's a draft PR.
1. To enable fastai dataloader on windows, force num_workers=0 on Windows 

Pros:
The fastai dataload can work on Windows
No changes on API.

Cons:
1.  Performance.
2. if users set num_workers > 0, the data isn't shuffled.


jph00:
No I don't think this is a solution - we need to be able to get >0 workers running on Windows. It's OK if it doesn't work in notebooks, but it should at least be made to work in scripts. I'll close this PR since I don't think this is the right approach. With 0 workers, performance isn't acceptable for nearly any practical vision task, for instance.

PR #3198, opened 2021-02-02
Make all other notebooks work on Windows

mszhanyi:
image

Depends on https://github.com/fastai/nbdev/pull/412 and https://github.com/fastai/fastcore/pull/289
1. It takes much more time, I've changed timeout to 30 minutes
2. I'm not very confident with change on
https://github.com/mszhanyi/fastai/blob/e1977e88ab5d01bc94fd41dc393b78a262770084/fastai/data/transforms.py#L180-L184
It's hard to handle backslash on Windows, it is escape character and path separator. 
And posixpath is recognized on Windows.
If you have better suggestion, I'll update it tomorrow.


fastai/data/transforms.py
```diff
@@ -178,8 +178,9 @@ def __init__(self, pat, match=False):
         self.matcher = self.pat.match if match else self.pat.search
 
     def __call__(self, o):
-        res = self.matcher(str(o))
-        assert res,f'Failed to find "{self.pat}" in "{o}"'
+        posix_o = Path(o).as_posix()
```
jph00, line 181: `Path` is really slow, and this function needs to be fast. So I'd suggest something like

```
import os,posixpath
o = str(o).replace(os.sep, posixpath.sep)
```

nbs/08_vision.data.ipynb
```diff
@@ -470,7 +479,16 @@
    "cell_type": "code",
    "execution_count": null,
    "metadata": {},
-   "outputs": [],
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
```
jph00, line 486: Let's remove this warning because it creates a big diff depending on which OS is used to run the NBs.

nbs/31_text.data.ipynb
```diff
@@ -867,7 +877,29 @@
    "cell_type": "code",
    "execution_count": null,
    "metadata": {},
-   "outputs": [],
+   "outputs": [
```
jph00, line 880: You've got an error here - remember that NBs become docs, so we need them to be clean.

nbs/40_tabular.core.ipynb
```diff
@@ -4,7 +4,16 @@
    "cell_type": "code",
    "execution_count": null,
    "metadata": {},
-   "outputs": [],
+   "outputs": [
```
jph00, line 7: Error here needs to be cleaned up

nbs/40_tabular.core.ipynb
```diff
@@ -1800,7 +1809,10 @@
     "\n",
     "test_series(to.cat_names, ['a', 'b_na'])\n",
     "test_series(to['a'], [1,2,3,2,2,3,1])\n",
-    "test_eq(df.a.dtype,int)\n",
+    "if sys.platform == \"win32\":\n",
```
jph00, line 1812: How about:

```
test_eq(df.a.dtype, np.int64 if sys.platform == "win32" else int)
```

Or maybe define a module variable somewhere like

```
default_int = np.int64 if sys.platform == "win32" else int
```

...and then use that?

nbs/tutorial.ipynb
```diff
@@ -4,7 +4,16 @@
    "cell_type": "code",
    "execution_count": null,
    "metadata": {},
-   "outputs": [],
+   "outputs": [
```
jph00, line 7: Looks like there are a few of these to clean up.

jph00 (commented):
Exciting progress!

Have you tried timing the notebooks? Use the `--timing true` switch to `nbdev_test_nbs` to get times for each, to see which are slow.

nbs/40_tabular.core.ipynb
mszhanyi: clear all outputs manully

nbs/31_text.data.ipynb
mszhanyi: Thanks reminder.
Cleared all outputs manully.

nbs/08_vision.data.ipynb
mszhanyi: Thanks reminder.
Cleared all outputs manully.

mszhanyi:
@jph00 
image
image

nbs/tutorial.ipynb
mszhanyi: @jph00 , 
Since most users are using Linux, to avoid potential confusion on the docs.
I reruned all updated notebooks on Linux and push the changes to the repo.

jph00:
> @jph00
> image
> image

Did you try running this notebook manually to see where/why it's so slow? Is that the only one causing the speed issue?

mszhanyi:
> > @jph00

> Did you try running this notebook manually to see where/why it's so slow? Is that the only one causing the speed issue?

@jph00 I did.  sorry, I added the snapshot in https://github.com/fastai/nbdev/pull/412, not in this PR

jph00:
> @jph00 <https://github.com/jph00> I did. sorry, I added the snapshot in fastai/nbdev#412 <https://github.com/fastai/nbdev/pull/412>, not in this PR


Ah OK - looks like you didn't at-mention me, so I missed those. Sorry about that.

PR #3200, opened 2021-02-03
Error when self.pred is a tuple of tensors

joomyjoo:
When self.pred is a tuple of tensors, calling .dtype on self.pred raises a AttributeError.
When self.pred is a tuple of tensors, the component tensors are (usually) of the same type.
Thus, we can check the dtype of the zeroth element of the listify(self.pred).


jph00:
Thanks for the PR. This is an nbdev project, so edits need to be made in the appropriate notebook, along with a test to exercise the new behavior, and prose to explain it.

I'll close this PR - please open a new one, or reopen this one, if you're interested in moving ahead.

PR #3201, opened 2021-02-03
Memory reduction for ImageClassifierCleaner

shusheer:
ImageClassifierCleaner seems to only use learn.get_preds(...with_input=True...) in order to decode the vocab of classes. This uses a lot of memory. Therefore apply the vocab directly to save memory


jph00:
Thanks.

PR #3203, opened 2021-02-04
Not change system default locale

mszhanyi:
remove the line of chaning system default locale.
set encoding="UTF-8" in tok.read_text()


jph00:
Thanks.

PR #3205, opened 2021-02-04
train test optimization on Windows

mszhanyi:
1. We can only use single process to load data, so we random select 1/8 image files as input.
2. Test data for 14_callback.schedule.ipynb.
    step 1
image
    step 2
image

    2.1 For windows CPU, in notebook.
         if we use 1/8 image data, step1  takes about 1minutes,  11 minutes for all image data.
         it will take 8 minutes for step2 for both 1/8 data and all data.
   2.2 Windows GPU, in notebook
         step1 takes 13s for 1/8 data, 1min47s for all data.
         step 2 takes one more minutes for both 1/8 data and all data.
         
         But it only takes 13 seconds for all data. (Does nbdev_test_nbs ignore tests marked #slow)
        image

   2.3 Run scripts with all data on Windows GPU. (the script attached)
         2min41s

3. If we run imagedataloader in scripts with label_func as lambda, there's a exception as below
image


mszhanyi:
[traintest.txt](https://github.com/fastai/fastai/files/5927135/traintest.txt)

jph00:
If you just mark slow cells with `#slow` then they will be skipped. For details, see:
https://nbdev.fast.ai/test

mszhanyi:
@jph00 , Thank your help, no need to extend tests timeout.

PR #3208, opened 2021-02-05
Enhancement: Let TextDataLoaders take in a custom tok_text_col

muellerzr:
Fixes bug introduced in https://github.com/fastai/fastai/pull/3204 and properly implements wanted feature


jph00 (changes requested):
Thanks for this. Can you please add a little documentation and a test?

jph00:
Sorry @muellerzr looks like a conflict has appeared - any chance you could resolve that so I can merge this?

muellerzr:
Sure. Will also add a brief test as well

muellerzr:
@jph00 made adjustments. The test will fail based on the previous commit we're fixing (since IMDB_Sample names it's text column as `text`, it's an easy situation to miss)

jph00:
Great!

PR #3211, opened 2021-02-07
add export for preds callbacks

nightlifelover:
I noticed that the MCDropoutCallback is not exported. This is my attempt to fix it but I am not sure whether I edited the correct file since I am not (yet) familiar with nbdev.

### DIFF
--- a/fastai/callback/all.py
+++ b/fastai/callback/all.py
@@ -8,3 +8,4 @@ from .schedule import *
 from .tracker import *
 from .rnn import *
 from .training import *
+from .preds import *


jph00:
It needs a `#export` at the top of its cell, if it doesn't already have one. Also there's some info here
https://docs.fast.ai/dev-setup.html 

(And the nbdev tutorial of course.)

nightlifelover:
`#export` is already present at the top of the cell where the class `MCDropoutCallback` is defined ([link to source](https://github.com/fastai/fastai/blob/master/nbs/18b_callback.preds.ipynb)). The only thing missing is the entry in the `fastai/callback/all.py` file for the generated `preds.py` file.

jph00:
Great thanks.

Issue #3202, opened 2021-02-04
AttributeError for multiple output models when using mixed precision


joomyjoo:
In the callback/fp16.py, .dtype is called on self.preds
However, when the model has two or more outputs, self.preds is a tuple of tensors not tensor
This results in an assertion error when self.preds.dtype is called.

To reproduce, train a model with two or more outputs with mixed precision.
"""
/usr/local/lib/python3.6/dist-packages/fastai/callback/fp16.py in after_pred(self)
     20     def before_batch(self): self.autocast.__enter__()
     21     def after_pred(self):
---> 22         if self.pred.dtype==torch.float16: self.learn.pred = to_float(self.pred)
     23     def after_loss(self): self.autocast.__exit__()
     24     def before_backward(self): self.learn.loss_grad = self.scaler.scale(self.loss_grad)

AttributeError: 'tuple' object has no attribute 'dtype'
"""

Expected Behavior:
The dtype check should be done on one of the outputs, regardless the number of outputs of a model.
Checking all outputs for dtype should not be necessary, because outputs should be of the same dtype.

Fix:
Using listify(self.preds)[0].dtype instead of self.preds[0] in callback/fp16.py
should resolve this issue.

--> I'm new to github and nbdev; I've managed to push the changes to joomyjoo/fastai;
but im not sure to the request a merge to the master. Hoping someone could help!

tmabraham:
I have a similar issue with my package [UPIT](https://github.com/tmabraham/UPIT), great to see you already have a fix.

Have you tried following [this guide](https://docs.fast.ai/dev-setup.html) to make your changes and submit a PR?

joomyjoo:
> I have a similar issue with my package [UPIT](https://github.com/tmabraham/UPIT), great to see you already have a fix.
> 
> Have you tried following [this guide](https://docs.fast.ai/dev-setup.html) to make your changes and submit a PR?

Thanks, I followed those steps to request a merge [Here](https://github.com/fastai/fastai/pull/3212)

tmabraham:
This issue was solved by #3217 (cc: @jph00)

jph00:
Thanks!

PR #3212, opened 2021-02-08
AssertionError when checking dtype on multiple output model

joomyjoo:
Fixed using listify()[0] on self.pred before calling .dtype. Made changes to the ipynb notebooks for testeq-ing self.x and self.pred dtypes for multiple input and multiple output models (the changes to the jupyter notebook does not seem necessary if the example used in the ipynb is going to be one with a single input and single output.

### DIFF
--- a/nbs/18_callback.fp16.ipynb
+++ b/nbs/18_callback.fp16.ipynb
@@ -197,7 +197,7 @@
     "    def before_fit(self): self.learn.scaler,self.scales = GradScaler(**self.kwargs),L()\n",
     "    def before_batch(self): self.autocast.__enter__()\n",
     "    def after_pred(self):\n",
-    "        if self.pred.dtype==torch.float16: self.learn.pred = to_float(self.pred)\n",
+    "        if listify(self.pred)[0].dtype==torch.float16: self.learn.pred = to_float(self.pred)\n",
     "    def after_loss(self): self.autocast.__exit__()\n",
     "    def before_backward(self): self.learn.loss_grad = self.scaler.scale(self.loss_grad)\n",
     "    def before_step(self):\n",
@@ -222,7 +222,7 @@
     "class FP16TestCallback(Callback):\n",
     "    \"Asserts that predictions are `float16` values\"\n",
     "    order = 9\n",
-    "    def after_pred(self): assert self.pred.dtype==torch.float16"
+    "    def after_pred(self): assert listify(self.pred)[0].dtype==torch.float16"
    ]
   },
   {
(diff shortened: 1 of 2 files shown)


joomyjoo:
Sry for the needlessly long title.
A merge request for fixing [https://github.com/fastai/fastai/issues/3202](url)

joomyjoo:
This is a nbdev-successful pull request for https://github.com/fastai/fastai/pull/3200

joomyjoo:
In the callback/fp16.py, .dtype is called on self.preds
However, when the model has two or more outputs, self.preds is a tuple of tensors not tensor
This results in an assertion error when self.preds.dtype is called.

To reproduce, train a model with two or more outputs with mixed precision.
"""
/usr/local/lib/python3.6/dist-packages/fastai/callback/fp16.py in after_pred(self)
20 def before_batch(self): self.autocast.enter()
21 def after_pred(self):
---> 22 if self.pred.dtype==torch.float16: self.learn.pred = to_float(self.pred)
23 def after_loss(self): self.autocast.exit()
24 def before_backward(self): self.learn.loss_grad = self.scaler.scale(self.loss_grad)

AttributeError: 'tuple' object has no attribute 'dtype'
"""

Expected Behavior:
The dtype check should be done on one of the outputs, regardless the number of outputs of a model.
Checking all outputs for dtype should not be necessary, because outputs should be of the same dtype.

Fix:
Using listify(self.preds)[0].dtype instead of self.preds[0] in callback/fp16.py
should resolve this issue.

nbs/18_callback.fp16.ipynb
```diff
@@ -770,8 +770,8 @@
     "class TestBeforeMixedPrecision(Callback):\n",
     "    order=-55\n",
     "    def before_fit(self): test_eq(first(self.model.parameters()).dtype, torch.float32)\n",
-    "    def before_batch(self): test_eq(self.x.dtype, torch.float32)\n",
-    "    def after_pred(self): test_eq(self.pred.dtype, torch.float16)\n",
+    "    def before_batch(self): test_eq(listify(self.x)[0].dtype, torch.float32)\n",
```
jph00, line 773: Not sure we need to change the test do we?

jph00 (changes requested):
Thanks for the PR. Can you please add a small simple test to the notebook to show the behavior this change is exercising?

nbs/18_callback.fp16.ipynb
joomyjoo: Yup, we do not need to change the tests.

joomyjoo:
A clean edit of the jupyter notebook with a multioutput example is done [here](https://github.com/fastai/fastai/pull/3217).

Will close this pull request

PR #3214, opened 2021-02-08
Fix test in text.data and add ninja to dev requirements

muellerzr:
This PR solves two current issues regarding tests:

First in `text.data` we downloaded the data again but never used it, so now `df = pd.read_csv(path/'texts.csv')` was added. In regards to QRNN failing, it seemed (on my machine) to stem from ninja not being installed to properly handle the c++, so I added this to the dev requirements.

(cc @jph00 )


jph00:
Thank you ! :)

PR #3216, opened 2021-02-08
Fix get_grid() when only ncols is provided but not nrows

nightlifelover:
I noticed a small bug in `get_grid()`: If you only provide `ncols` but not `nrows` it can return less axes than the `n` specified. For example I ran into this:

`len(get_grid_buggy(n=32, ncols=4))`

Returned 20 instead of 32. This PR fixes this behavior.


jph00:
This seems to ignore rows if cols is passed?

nightlifelover:
Right, sorry I was a bit too quick with this one. Let me open an new PR.

PR #3217, opened 2021-02-09
Assertion Error for fp16() in multioutput models (with example)

joomyjoo:
As requested in [here](https://github.com/fastai/fastai/pull/3212) for an example.
Fixes assertionerror for mixedprecision callback using pytorch's native mixed precision function,
when using a multioutput model.

Added an example of a trivial multioutput model.
I used #cuda and #hide in the jupyter notebook cell,
Its my first time using github/contributing with nbdev, but wow is this super hard and complicated.
Im sure itll get better with the learning curve (steep).


jph00:
Good job - thanks for taking the time to figure it out! :)

PR #3218, opened 2021-02-09
Fix get_grid() when only ncols is provided but not nrows

nightlifelover:
I noticed a small bug in get_grid(): If you only provide ncols but not nrows it can return less axes than the n specified. For example I ran into this:

len(get_grid_buggy(n=32, ncols=4))

Returned 20 instead of 32. This PR (2nd attempt) fixes this behavior.


nightlifelover:
Not sure why "nbs are not synched". Is it because of this?

```
-# AUTOGENERATED! DO NOT EDIT! File to edit: nbs/nbs/08_vision.data.ipynb (unless otherwise specified).
+# AUTOGENERATED! DO NOT EDIT! File to edit: nbs/08_vision.data.ipynb (unless otherwise specified).
```

I rand  `nbdev_build_lib --fname nbs/08_vision.data.ipynb ` to create the Python file.

jph00:
Thanks!

PR #3223, opened 2021-02-14
px_summary_update

asvcode:
Allow to toggle on or off the pixel summary within `from_dicoms`.   The main reason for this is that for large datasets the process can take a while so this allows for greater control when creating a dataframe of the metadata.


jph00:
Thanks!

Issue #3225, opened 2021-02-14
OptimWrapper no longer works with Pytorch Optimizers OOTB


muellerzr:
We used to be able to perform something like so to utilize a raw torch optimizer:
```python
opt1 = torch.optim.AdamW(model.parameters(),
                                      betas=(0.9,0.99),
                                      eps=1e-5,
                                      weight_decay=1e-2)
opt_func = compose(opt1, OptimWrapper)
```
And then pass this into `Learner`. However now this is no longer possible with the following error:
```python
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-15-8587f3539821> in <module>()
----> 1 learn.fit(1)

1 frames
/usr/local/lib/python3.6/dist-packages/fastai/learner.py in fit(self, n_epoch, lr, wd, cbs, reset_opt)
    204     def fit(self, n_epoch, lr=None, wd=None, cbs=None, reset_opt=False):
    205         with self.added_cbs(cbs):
--> 206             if reset_opt or not self.opt: self.create_opt()
    207             if wd is None: wd = self.wd
    208             if wd is not None: self.opt.set_hypers(wd=wd)

/usr/local/lib/python3.6/dist-packages/fastai/learner.py in create_opt(self)
    147     def _bn_bias_state(self, with_bias): return norm_bias_params(self.model, with_bias).map(self.opt.state)
    148     def create_opt(self):
--> 149         self.opt = self.opt_func(self.splitter(self.model), lr=self.lr)
    150         if not self.wd_bn_bias:
    151             for p in self._bn_bias_state(True ): p['do_wd'] = False

TypeError: 'AdamW' object is not callable
```

cc @jph00

muellerzr:
It does indeed, however it needs to be defined as such:
```python
def opt_func(params, **kwargs): return OptimWrapper(optim.SGD(params, lr=0.001))
```
The docs should be updated (and this closed once they are)

tyoc213:
I think you can also pass directly to the fit method IIRC just as parameter... barely remember something like `fit(optim= OptimWrapper(optim.SGD(params, lr=0.001)))`

muellerzr:
@tyoc213 that used to be able to work, but now we need to actually write a basic `opt_func` to wrap it in

muellerzr:
Closed with https://github.com/fastai/fastai/commit/89b05f546f86c047428538b9436fa5c2a9421f6e

PR #3226, opened 2021-02-16
optimizer wrapper improve docs

tmabraham:
There is often confusion about how to use `OptimWrapper` (so we can use PyTorch optimizers). Even @muellerzr has experienced some confusion in this regard (see https://github.com/fastai/fastai/issues/3225). This PR is to improve some of the documentation to clarify this. I don't know if this is the best approach so I'm open to moving the text I wrote to another location, or adding some text somewhere else, etc...


jph00:
Do we know what's changed that's caused the `compose()`/inline approach to not work any more? cc @muellerzr

muellerzr:
@jph00 I sadly don't particularly know what exactly is causing the issue. (I also don't know what version fixed this, as I didn't really use this functionality much). 

For some reason it looks like it's trying to do something like `AdamW()()` (IE call the class instance) rather then `.step` I believe? But I also don't know what the old behavior exactly did.

@tmabraham used it much more, so he may have an idea?

tmabraham:
There are two problems when it comes to the compose approach. This is what I am trying:

```
opt_func = compose(partial(torch.optim.Adam, eps=1e-6), OptimWrapper)
```

(I had also tried something similar earlier with AdaBelief)

If I try passing this `opt_func` into `Learner` like this:
```
learn_ = Learner(dls, resnet34(), opt_func=opt_func, metrics=error_rate)
learn_.fit(1)
```
I get this error:
```
TypeError: __init__() got an unexpected keyword argument 'lr'
```
I believe this is a limitation of fastcore's `compose` function. `compose` passes the same `kwargs` into all of the composed functions, but `OptimWrapper` doesn't expect `lr` and other related arguments, hence the error.

If I try with `cnn_learner` like this:
```
path = untar_data(URLs.PETS)/'images'

def is_cat(x): return x[0].isupper()
dls = ImageDataLoaders.from_name_func(
    path, get_image_files(path), valid_pct=0.2, seed=42,
    label_func=is_cat, item_tfms=Resize(224))

learn = cnn_learner(dls, resnet34, opt_func=opt_func, metrics=error_rate)
learn.fit(1)
```

I get the following error:
```
TypeError: optimizer can only optimize Tensors, but one of the params is list
```
I am not sure why I get this error, but it may have something to do with the `param_lists`/the splitting of the model.

For these reasons, the `compose` approach does not currently work...

tyoc213:
It will be nice to know the pytorch commit that caused this, I guess pytorch is 1.7.0 but has been since some time?. So I guess the different behaviour was introduced from 1.6 to current 1.7 isn't? or we can't be sure about that?


As a side note from what I know from the TPU thing, there are some tiny diffs very internal to optimizers from fastai and pytorch so perhaps could be a good time to test in deep compatibility between the two?

tmabraham:
closing and bringing proposed changes to #3241

PR #3230, opened 2021-02-19
Regression fix: Ensure add_datepart adds elapsed as numeric column

aberres:
After the merge of #2640 the `elapsed` column added by `add_date_part` is always of type `object`. 

This causes problems like the one described [here](https://forums.fast.ai/t/lesson-7-official-topic/69896/282?u=aberres).

Before #2640 the added column was always of type int. With my change it will always be of type float. This will also cause `cont_cat_split` to consider the new column as continuous. While this would be a behavioral change, isn't this what we want after all?

The other option would be to cast to int if there are no `nan` input values.

@kevinbird15  @jph00


jph00:
Many thanks.

PR #3242, opened 2021-03-06
Increase testing granularity for speedup

ddobrinskiy:
Using 9 batches instead of 3:

Current setup of 3 batches means that 20+ notebooks fall into the [5-7] batch and it takes 9minutes to test all the notebooks, example: https://github.com/fastai/fastai/pull/3235/checks?check_run_id=1969103831

With minimal code changes, we can split notebooks into 9 batches instead of 3, which should speed-up testing

**upd**: with new configuration, actual notebook testing [takes at most 1m20s](https://github.com/fastai/fastai/pull/3242/checks?check_run_id=2046070822) (excluding docker setup, cache download, et cetera)

PR #3243, opened 2021-03-06
Increase testing granularity for speedup (v2)

ddobrinskiy:
Using 9 batches instead of 3:

Current setup of 3 batches means that 20+ notebooks fall into the [5-7] batch and it takes 9minutes to test all the notebooks, example: https://github.com/fastai/fastai/pull/3235/checks?check_run_id=1969103831

With minimal code changes, we can split notebooks into 9 batches instead of 3, which should speed-up testing


ddobrinskiy:
Tried to fix problem when github is waiting for testing stages that no longer exist. 

The problem was still present though https://github.com/fastai/fastai/pull/3242

PR #3255, opened 2021-03-11
Fix #3246

### LINKED ISSUE #3246: Lighting patch overwrites subclassed `TensorImage` type to `TensorImage`
First asked on the forum [here](https://forums.fast.ai/t/bug-lightning-patch-overwrites-type/86352).

**Describe the bug**
The lightning transform is patched to all `TensorImage` types but converts the input to a `TensorImage` and breaks sub-classed `TensorImage` code. 
See:

https://github.com/fastai/fastai/blob/master/fastai/vision/augment.py#L668

**To Reproduce**
To reproduce and see my proposed fix take a look at the following Colab notebook:

https://colab.research.google.com/drive/1Mexm84QcKNEag1CBmrI70SJ95pyz6YTU?usp=sharing

TL;DR:
Change

```python
@patch
def lighting(x: TensorImage, func): return TensorImage(torch.sigmoid(func(logit(x))))
```

to

```python
@patch
def lighting(x: TensorImage, func): return torch.sigmoid(func(logit(x))
```

But I may be missing why the conversion to TensorImage is necessary.
For me at least, it is unexpected behavior and breaks my subclass.

Would be more than happy to create a tiny PR and hear some comments :slight_smile:

kai-tub:
Do not convert possible `TensorImage` subclass to `TensorImage` in `.lighting` function.
Add hidden test to ensure that the original subclass is indeed untouched.

### DIFF
--- a/fastai/vision/augment.py
+++ b/fastai/vision/augment.py
@@ -665,7 +665,7 @@ class Warp(AffineCoordTfm):
 
 # Cell
 @patch
-def lighting(x: TensorImage, func): return TensorImage(torch.sigmoid(func(logit(x))))
+def lighting(x: TensorImage, func): return torch.sigmoid(func(logit(x)))
 
 # Cell
 class SpaceTfm(RandTransform):


jph00:
Many thanks. I suspect this cast is no longer required due to the auto-casting we added to TensorBase.

PR #3256, opened 2021-03-11
custom tokenizer

iskode:
- fix error setting 'tok_tfm' parameter in TextDataloaders.from_df
- fix error setting  'tok_tfm' parameter in TextDataloaders.from_folder


iskode:
This PR fixes some errors while using a custom tokenizer with the `TextDataloaders` class via 
the parameter `tok_tfm`. See this[ short post ](https://forums.fast.ai/t/changing-the-tokenizer-default-language-in-text-data/86316) for details.

jph00 (changes requested):
Many thanks!

I see that some of the output cells are now missing - could you please re-run the full notebook so that the outputs are included? Otherwise the documentation will be missing the outputs.

iskode:
> Many thanks!
> 
> I see that some of the output cells are now missing - could you please re-run the full notebook so that the outputs are included? Otherwise the documentation will be missing the outputs.

Sorry for my late reply.. and sorry again as  didn't pay attention to cell outputs.
I've made a commit after running all cells in the notebook.

hamelsmu:
@jph00 this is ready for review

jph00:
Many thanks :)

Issue #3191, opened 2021-01-25
Broken link in callbacks docs.


KBodolai:
Hi!

I was going through the callbacks docs, where I stumbled upon a broken link [in the mixed precision section](https://docs.fast.ai/callback.fp16.html). The broken link in question is the one to the video from NVIDIA with some background.

I think the correct link might be [this one](https://on-demand.gputechconf.com/gtc/2019/video/_/S9143/).

K.

muellerzr:
@hamelsmu this can be closed

PR #3263, opened 2021-03-16
Fix dead link #3191

ReinierKoops:
Fixed a link that wasn't working anymore in the docs.


ReinierKoops:
It is related to the [fp16 callbacks](https://docs.fast.ai/callback.fp16.html) doc.

jph00:
Thank you

PR #3267, opened 2021-03-17
Remove empty block

hinryd:
Refer to https://docs.fast.ai/tutorial.collab.html


jph00:
That comment is needed actually.

PR #3292, opened 2021-04-02
Improve inplace params in Tabular's new and allow for new and test_dl to be in place

muellerzr:
Adjusts the params in `Tabular`'s new to include an `inplace` param for those that want to only use one copy in memory. Also included a test for this based on counting `pd.DataFrame` objects in memory. 

This was then also brought up through to `test_dl` as well to properly pass in the `inplace` param. 

Originally based on: https://github.com/fastai/fastai/pull/2858

Closes https://github.com/fastai/fastai/issues/2755

cc: @jph00


hamelsmu:
@jph00 this is ready for review

jph00:
Many thanks :)

PR #3296, opened 2021-04-03
add `at_end` feature to `SaveModelCallback`

tmabraham:
This PR adds a new argument to `SaveModelCallback` that allows us to automatically save the model at the end of fitting.


jph00:
Great idea!

Issue #3294, opened 2021-04-02
Learner.one_batch doesn't move the whole batch (xb,yb) to device when xb is a tuple


oguiza:
PR #3253 modified Learner.one_batch method. 

This is the current version: 

```
def one_batch(self, i, b):
    self.iter = i
    b_on_device = tuple(e.to(device=self.dls.device) for e in b if hasattr(e, "to")) if self.dls.device is not None else b
    self._split(b_on_device)
    self._with_events(self._do_one_batch, 'batch', CancelBatchException)
```
This works well when the xb is a single tensor. Both xb and yb are moved to the device. 
```
# This works well
xb = torch.randn(2, 3, 4)
yb = torch.randint(0, 5, (16,))
b = (xb, yb)
device = default_device()
tuple(e.to(device=device) for e in b if hasattr(e, "to"))
```

However, when the xs is made of a couple of tensors (for example when using mixed dataloaders), it doesn't move the whole batch as it's not applied recursively. 
```
# This doesn't produce the expected output
xb1 = torch.randn(2, 3, 4)
xb2 = torch.randn(2, 3, 4)
yb = torch.randint(0, 5, (16,))
b = ((xb1, xb2), yb)
device = default_device()
tuple(e.to(device=device) for e in b if hasattr(e, "to")) if device is not None else b
```
The code above just returns a tuple with yb.

The following code does produce the expected outcome: 
```
xb1 = torch.randn(2, 3, 4)
xb2 = torch.randn(2, 3, 4)
yb = torch.randint(0, 5, (16,))
b = ((xb1, xb2), yb)
device = default_device()
to_device(b, device=device) if device is not None else b
```

The proposed solution would be to modify this line in Learner.one_batch: 
```
def one_batch(self, i, b):
    self.iter = i
    b_on_device = tuple(e.to(device=self.dls.device) for e in b if hasattr(e, "to")) if self.dls.device is not None else b
    self._split(b_on_device)
    self._with_events(self._do_one_batch, 'batch', CancelBatchException)
```
by this one: 
```
def one_batch(self, i, b):
    self.iter = i
    b_on_device = to_device(b, device=self.dls.device) if self.dls.device is not None else b
    self._split(b_on_device)
    self._with_events(self._do_one_batch, 'batch', CancelBatchException)
```
so that the whole batch is moved to the device.

ohmeow:
Yah, this has broken blurr as well :(

In the case of blurr, `xb` is a dictionary that looks like this:
```
[...]

PR #3298, opened 2021-04-06
Fix issue #3294 (tuples not moving to gpu)

tcapelle:
Fix issue with tuples not moving to gpu from @oguiza 
```python
# This doesn't produce the expected output
xb1 = torch.randn(2, 3, 4)
xb2 = torch.randn(2, 3, 4)
yb = torch.randint(0, 5, (16,))
b = ((xb1, xb2), yb)
device = default_device()
tuple(e.to(device=device) for e in b if hasattr(e, "to")) if device is not None else b
```
but this works:
```python
xb1 = torch.randn(2, 3, 4)
xb2 = torch.randn(2, 3, 4)
yb = torch.randint(0, 5, (16,))
b = ((xb1, xb2), yb)
device = default_device()
to_device(b, device=device) if device is not None else b
```


tcapelle:
I did not even touch this nb :/

jph00:
Many thanks

PR #3299, opened 2021-04-07
upgrade template to use remote theme

hamelsmu:
cc: @jph00


hamelsmu:
reverting this PR.  for some reason didn't render correctly on GitHub pages.   Will investigate.

PR #3305, opened 2021-04-07
Default show mask

tcapelle:
Set default `vmin` param to 0, even for the tutorial it is needed as the zero code is for animal.
Probably was set to 0 for a dataset with 0 as the background.
- Also removes test after export on  `data.load`, this test does not pass on a Ryzen CPU without MKL support.
- Also added a file that was generated automatically with the hashs of the S3 buckets containing imagenette (probably it changed at some point)

### DIFF
--- a/fastai/data/checks.txt
+++ b/fastai/data/checks.txt
@@ -176,12 +176,12 @@
     "2c774ecb40005b35d7937d50f5d42336"
   ],
   "https://s3.amazonaws.com/fast-ai-imageclas/imagenette2-160.tgz": [
-    99003388,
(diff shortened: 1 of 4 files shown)


tcapelle:
dont'get what this error is...

hamelsmu:
I reset the CI in this PR to reflect fixes to nbdev.  @jph00 you can review now

tcapelle:
Thanks @hamelsmu for the reply, I was pretty sure the CI was down. It is crazy these attacks to mine crypto!

jph00:
Many thanks.

PR #3316, opened 2021-04-12
Fix 3312

tcapelle:
Typo on shuffled attr on Learner

Don't know if there is a way to test that non-existant kwargs raise an exception.


muellerzr:
cc @jph00 this looks fine to me. The reason this never popped up any tests is `shuffle` is set to `False` by default, so it was still generating a non-shuffled `DataLoader`. Just instead the wrong param was exposed.

tcapelle:
My question was about exposing a non-existant param. Shouldn't calling a method/func with a non-existant kwargs raise an error?
Here we are calling `DataLoaders.new` with a kwargs that does not exists, how can we make this fail.

jph00:
Many thanks!

PR #3318, opened 2021-04-13
Add windows support in README

mszhanyi:
declaration of windows support
@jph00


jph00:
@mszhanyi I've merged this PR, but it won't actually stick since the README is auto-generated from `nbs/index.ipynb`. So you need to edit that notebook, then run `nbdev_build_docs`. That will automatically update the docs too.

I also suggest you don't add a section saying "windows is now supported" since that is a message that is specific to a point in time. Instead, simply add a section called "Windows support" which says something like "fastai runs on Windows, but please keep in mind the following issues:"... and then list the things people need to know.

PR #3319, opened 2021-04-13
Add order to TrackerCallback derived classes

lluissalord:
There are some cases where you want to specify that the callback is between the callbacks derived from TrackerCallback. Currently all of them have set `order=60` from its parent, if we make that each callback has its own `order` then it can be specified the desired order.

Proposed order is set according to the current behaviour which is `SaveModelCallback`, `ReduceLROnPlateau` and finally `EarlyStoppingCallback` which is logic to have `EarlyStoppingCallback` as the last one.

### DIFF
--- a/fastai/callback/tracker.py
+++ b/fastai/callback/tracker.py
@@ -43,6 +43,7 @@ class TrackerCallback(Callback):
 # Cell
 class EarlyStoppingCallback(TrackerCallback):
     "A `TrackerCallback` that terminates training when monitored quantity stops improving."
+    order=TrackerCallback.order+3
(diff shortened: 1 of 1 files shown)


Leeeooo19888 (approved):

jph00 (changes requested):
Many thanks @lluissalord ! Can you please modify the source notebooks, instead of the .py file directly, and then run make or nbdev_build_lib so that the code is synced with the notebooks?

lluissalord:
Currently I do not have the fork prepared for contributing and not so much time. If at some point I have time to make the changes I will do as you mentioned. 

However, if someone wants to proceed with this small change then I would appreciate.

jph00:
Got it - thanks @lluissalord . I'll close this PR then

PR #3326, opened 2021-04-20
sort Perplexity with NLP metrics

KChalk:
This also puts it immediately adjacent to, but not in, Loss Metrics. Why Perplexity is a subclass of AvgLoss while LossMetric is a subclass of AvgMetric may be worth documenting here. 

Apologies if you think this is trivial, but I almost missed it, as I was not looking for classification metrics.


jph00:
Many thanks @KChalk ! Could you please run `make` or `nbdev_build_lib` so that the code is synced with the notebooks?

KChalk:
Hi @jph00, I hoped to get around the complex notebook set up by editing the html directly, but I see that I did not manage that. If it's quick for you, I don't need credit for the PR and you're welcome to reject it and redo. Otherwise, I can plan to set up an environment and follow the prescribed steps to make the same change in the not-so-distant future.

jph00:
No worries @KChalk - thanks for letting us know.

hamelsmu:
Thanks, @KChalk looks great 👍🏽

PR #3330, opened 2021-04-22
Enhancement: Automatically have fastai determine the right device, even with torch DataLoaders

muellerzr:
# Summary of Changes

## Callback.Core:
The original code looked like so:
```python
if hasattr(self.dls, "device"): self.model.to(self.dls.device)
```
And the purpose was to move the model to the GPU or CPU depending on if we're using fastai `DataLoaders`. This isn't particularly flexible though if we're using raw torch loaders. So instead I've opted for the following:

```python
device = getattr(self.dls, 'device', default_device())
self.model.to(device)
```
This way if we're using raw torch loaders in a CUDA environment, fastai can pick up on that. (More later on how we can still let the user train on the CPU)

## Callback.Data:
Completely removed the `CudaCallback`. With these added changes, it's become irrelevant. 

## Learner:
I've added a `_set_device` function to `Learner`. Its goal is to set the batch to the proper device based on the model.
Basic functionality can be summized as if the model and the dataloader's devices are the same, return the DataLoader's device. If not, return the models' device. 

This still maintains how fastai determined the model device (as it was all dependent on what the model or DataLoader was set to), but also now allows for raw torch loaders to be set accordingly to the model's device, as this is the only thing they *can* set.

Along with this `one_batch` now calls `_set_device` before `_split`

I've added slight changes to `synth_dbunch` and `synth_learner`. Both now support `tfmdDL=True`. If not, then we will use a pytorch `DataLoader` instead, letting us test and ensure that pytorch `DataLoaders` will *always* be supported within the library, and we can know if anything broke this functionality. 

# What This Does to the End Users
Adds a much needed ease of use when using torch `DataLoaders` with fastai's `Learner`, as users would *often* forget (or even know!) that the `CudaCallback` existed, and instead let's them keep the general fastai flow without having to step away from what everyone else using the library is doing.

# Added Tests
As all related `preds` tests are in `Learner`, the added tests for these functionalities also live in there.

## Training with Raw Torch on CPU
The first test is one to ensure we can train with raw pytorch, simply running the same as the training test but also with `tfmdDL=False`:
```python
learn = synth_learner(lr=0.1, tfmdDL=False)
learn(_before_epoch)
learn.model = learn.model.cpu()
xb,yb = next(iter(learn.dls[0]))
init_loss = learn.loss_func(learn.model(xb), yb)
learn.fit(10)
xb,yb = next(iter(learn.dls[0]))
learn.model = learn.model.cpu() # Ensure we're still on CPU even in a CUDA environment
final_loss = learn.loss_func(learn.model(xb), yb)
assert final_loss < init_loss, (final_loss,init_loss)
```
It should be noted here that to ensure that our model returns as a CPU even in a CUDA environment, we need to set it to CPU again
## Training with Raw Torch on CUDA
To make sure that our device setter is actually working, we run the same small CUDA test but with `tfmdDL=False`:
```python
learn = synth_learner(cbs=TestTrainEvalCallback, tfmdDL=False)
learn.fit(1)
```
Otherwise all tests were run with both the `cuda` and `slow` flags, and they passed. 

cc: @hamelsmu and @jph00


hamelsmu:
These changes break `notebooks/nbs/examples/migrating_pytorch.ipynb`

hamelsmu:
ok I fixed that issue

Issue #3051, opened 2020-12-03
ClassificationInterpretation.from_learner not working on training set


lluissalord:
First of all, thank you for this amazing framework!!

Using last version of fastai, when setting ds_idx=0 and training DataLoader is using `drop_last=True` raise IndexError.

My intuition is that as it is taking directly the DataLoader from training I suspect can generate incorrect behaviours if training is set with `drop_last=True`.

I suggest to change this part of the code:
https://github.com/fastai/fastai/blob/a07f271ac6a03cd14ff7f8c031c38527e5b238ed/fastai/interpret.py#L26-L29

to:

```python
def from_learner(cls, learn, ds_idx=1, dl=None, act=None):
        "Construct interpretation object from a learner"
        if dl is None: dl = learn.dls[ds_idx].new(shuffled=False, drop_last=False)
        return cls(dl, *learn.get_preds(dl=dl, with_input=True, with_loss=True, with_decoded=True, act=None))
```

Doing this it is replicated the behaviour seen on `get_preds`:
https://github.com/fastai/fastai/blob/cb01af7dbf7ae16fda68a7eec19a87c75423a8fe/fastai/learner.py#L223-L225

muellerzr:
@hamelsmu can be closed due to my fix earlier today

lluissalord:
Hi @muellerzr,

I have seen that the same issue happen on line 579:
https://github.com/fastai/fastai/blob/7fe5fcaab3c0b692a0be54ffe651f04f4cdd4a39/fastai/learner.py#L575-L579

Could you please also add this fix?
Thanks!

muellerzr:
@lluissalord thanks! Fixed with https://github.com/fastai/fastai/pull/3347

Issue #3248, opened 2021-03-09
Interpretation.from_learner fails with non-default ds_idx argument


d-miketa:
(Getting this on `fastai==2.2.5`)
You can't run Interpretation.from_learner with the training dataset. Something similar happens when you directly pass a dataloader. The method works well when all optional arguments are left as their defaults.

Minimal working example:
```
from fastai.vision.all import *
path = untar_data(URLs.PETS)/'images'

def is_cat(x): return x[0].isupper()
dls = ImageDataLoaders.from_name_func(
    path, get_image_files(path), valid_pct=0.2, seed=42,
    label_func=is_cat, item_tfms=Resize(224))

learn = cnn_learner(dls, resnet34, metrics=error_rate)

interpret = Interpretation.from_learner(learn, ds_idx=0)
interpret.plot_top_losses(5)
```

The code above results in the following error:
```
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
<ipython-input-1-6913508f0193> in <module>
      9 learn = cnn_learner(dls, resnet34, metrics=error_rate)
     10 
---> 11 interpret = Interpretation.from_learner(learn, ds_idx=0)
     12 interpret.plot_top_losses(5)

/usr/local/lib/python3.8/dist-packages/fastai/interpret.py in from_learner(cls, learn, ds_idx, dl, act)
     27         "Construct interpretation object from a learner"
     28         if dl is None: dl = learn.dls[ds_idx]
---> 29         return cls(dl, *learn.get_preds(dl=dl, with_input=True, with_loss=True, with_decoded=True, act=None))
     30 
     31     def top_losses(self, k=None, largest=True):

/usr/local/lib/python3.8/dist-packages/fastai/learner.py in get_preds(self, ds_idx, dl, with_input, with_decoded, with_loss, act, inner, reorder, cbs, **kwargs)
    248                 res[pred_i] = act(res[pred_i])
    249                 if with_decoded: res.insert(pred_i+2, getattr(self.loss_func, 'decodes', noop)(res[pred_i]))
--> 250             if reorder and hasattr(dl, 'get_idxs'): res = nested_reorder(res, tensor(idxs).argsort())
    251             return tuple(res)
    252         self._end_cleanup()

/usr/local/lib/python3.8/dist-packages/fastai/torch_core.py in nested_reorder(t, idxs)
    695     "Reorder all tensors in `t` using `idxs`"
    696     if isinstance(t, (Tensor,L)): return t[idxs]
[...]

PR #3332, opened 2021-04-22
Have interp safely use ds_idx, add tests

muellerzr:
# Summary of Changes
The `Interpretation` class now calls `.new(shuffled=False, drop_last=False)` when generating the `DataLoader` if none were found. This is similar to what is done during `get_preds`, and is needed as shuffling and dropping last will cause `Interpretation` to complain about missing indexs (see issue here: https://github.com/fastai/fastai/issues/3248)

Closes: https://github.com/fastai/fastai/issues/3051 and https://github.com/fastai/fastai/issues/3248

# What This Does to the End Users
Allows them to specify `ds_idx=0` without raising any issues. 

# Added Tests

I've changed the tests in `Interpretation` as a result of the current tests not being extensive enough. 

Before the tests were:
```python
learn = synth_learner()
interp = Interpretation.from_learner(learn)
x,y = learn.dls.valid_ds.tensors
test_eq(interp.inputs, x)
test_eq(interp.targs, y)
out = learn.model.a * x + learn.model.b
test_eq(interp.preds, out)
test_eq(interp.losses, (out-y)[:,0]**2)
```
And while they do get the job done, they're not flexible enough for us to check if the training or validation dataset is doing okay, and that decoding went well (along with the rest of the pipeline).

As a result I've instead added in using the `MNIST_TINY` dataset, and a subsample of it:
```python
#hide
mnist = DataBlock(blocks=(ImageBlock(cls=PILImageBW), CategoryBlock), 
                  get_items=get_image_files, 
                  splitter=RandomSubsetSplitter(.1,.1, seed=42),
                  get_y=parent_label)
test_dls = mnist.dataloaders(untar_data(URLs.MNIST_SAMPLE), bs=8)
test_learner = cnn_learner(test_dls, resnet18)
```

Now here are the associated tests:
```python
#hide
interp = Interpretation.from_learner(test_learner)
x, y, out = [], [], []
for batch in test_learner.dls.valid:
    x += batch[0]
    y += batch[1]
    out += test_learner.model(batch[0])
x,y,out = torch.stack(x), torch.stack(y, dim=0), torch.stack(out, dim=0)
test_eq(interp.inputs, x)
test_eq(interp.targs, y)
losses = torch.stack([test_learner.loss_func(p,t) for p,t in zip(out,y)], dim=0)
test_close(interp.losses, losses)
```
```python
#hide
#dummy test to ensure we can run on the training set
interp = Interpretation.from_learner(test_learner, ds_idx=0)
x, y, out = [], [], []
for batch in test_learner.dls.train.new(drop_last=False, shuffle=False):
    x += batch[0]
    y += batch[1]
    out += test_learner.model(batch[0])
x,y,out = torch.stack(x), torch.stack(y, dim=0), torch.stack(out, dim=0)
test_eq(interp.inputs, x)
test_eq(interp.targs, y)
losses = torch.stack([test_learner.loss_func(p,t) for p,t in zip(out,y)], dim=0)
test_close(interp.losses, losses)
```
It's a bit longer, but it does the same thing that the previous test did, but also takes into account us using the true `DataBlock` api, and adds a test for the `ds_idx=0`.

Let me know if there are any suggestions, such as keeping that first test back in with addition to these, or having these in with `#slow`.

cc @jph00 and @hamelsmu

### DIFF
--- a/fastai/interpret.py
+++ b/fastai/interpret.py
@@ -25,7 +25,7 @@ class Interpretation():
     @classmethod
     def from_learner(cls, learn, ds_idx=1, dl=None, act=None):
         "Construct interpretation object from a learner"
-        if dl is None: dl = learn.dls[ds_idx]
+        if dl is None: dl = learn.dls[ds_idx].new(shuffled=False, drop_last=False)
         return cls(dl, *learn.get_preds(dl=dl, with_input=True, with_loss=True, with_decoded=True, act=None))
 
     def top_losses(self, k=None, largest=True):
--- a/nbs/20_interpret.ipynb
+++ b/nbs/20_interpret.ipynb
@@ -58,6 +58,23 @@
    "metadata": {},
    "outputs": [],
    "source": [
+    "#hide\n",
+    "def filter_files(o): return o[:100]\n",
+    "from fastai.vision.all import *\n",
+    "mnist = DataBlock(blocks=(ImageBlock(cls=PILImageBW), CategoryBlock), \n",
+    "                  get_items=Pipeline([get_image_files, filter_files]), \n",
+    "                  splitter=RandomSubsetSplitter(.1,.1, seed=42),\n",
+    "                  get_y=parent_label)\n",
+    "test_dls = mnist.dataloaders(untar_data(URLs.MNIST_SAMPLE), bs=8)\n",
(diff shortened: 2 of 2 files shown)


hamelsmu:
I'll let Jeremy review this one

jph00:
Many thanks @muellerzr !

Issue #3222, opened 2021-02-13
GAN Training


Isaac-Flath:
Be sure you've searched [the forums](https://forums.fast.ai) for the error message you received. Also, unless you're an experienced fastai developer, first ask on the forums to see if someone else has seen a similar issue already and knows how to solve it. **Only file a bug report here when you're quite confident it's not an issue with your local setup.**

**Please see [this model example](https://github.com/fastai/fastai2/issues/487) of how to fill out an issue correctly.  Please try to emulate that example as appropriate when opening an issue.**

Please confirm you have the latest versions of fastai, fastcore, and nbdev prior to reporting a bug (delete one): YES 

**Describe the bug**

There is a bug in GANLoss.  It errors on current version.  When installed with dev version, it runs but gives warnings.  They do not appear to me to be related, but adding both in case they are.

**To Reproduce**
1. Open `24_vision.gan.ipynb` in colab
2. Run All 
3. Fit at the end of the notebook is the issue

**Error with full stack trace**

**Error (normal install)**
```
TypeError                                 Traceback (most recent call last)
<ipython-input-39-d8be2b5f57fe> in <module>()
----> 1 learn.fit(1, 2e-4, wd=0.)

13 frames
<ipython-input-15-422ae153f30b> in critic(self, real_pred, input)
     14     def critic(self, real_pred, input):
     15         "Create some `fake_pred` with the generator from `input` and compare them to `real_pred` in `self.crit_loss_func`."
---> 16         fake = self.gan_model.generator(input).requires_grad_(False)
     17         fake_pred = self.gan_model.critic(fake)
     18         self.crit_loss = self.crit_loss_func(real_pred, fake_pred)

TypeError: requires_grad_() takes 1 positional argument but 2 were given
```
**Warning (editable install)**
```
/usr/local/lib/python3.6/dist-packages/fastai/callback/core.py:50: UserWarning: You are shadowing an attribute (generator) that exists in the learner. Use `self.learn.generator` to avoid this
  warn(f"You are shadowing an attribute ({name}) that exists in the learner. Use `self.learn.{name}` to avoid this")

/usr/local/lib/python3.6/dist-packages/fastai/callback/core.py:50: UserWarning: You are shadowing an attribute (critic) that exists in the learner. Use `self.learn.critic` to avoid this
  warn(f"You are shadowing an attribute ({name}) that exists in the learner. Use `self.learn.{name}` to avoid this")

[...]

PR #3334, opened 2021-04-22
Fix warnings during GAN training

AndreyYashkin:
This issue is described in https://github.com/fastai/fastai/issues/3222

According to _Callback_ documentation "Callbacks provide a shortcut to avoid having to write self.learn.bla for any bla attribute we seek; instead, just write self.bla. **This only works for getting attributes, not for setting them.**"


fastai/vision/gan.py
```diff
@@ -148,8 +148,8 @@ def _set_trainable(self):
 
     def before_fit(self):
         "Initialize smootheners."
-        self.generator,self.critic = self.model.generator,self.model.critic
-        self.gen_mode = self.gen_first
+        self.learn.generator,self.learn.critic = self.model.generator,self.model.critic
```
jph00, line 151:
```suggestion
        self.gan_generator,self.gan_critic = self.model.generator,self.model.critic
```

fastai/vision/gan.py
```diff
@@ -148,8 +148,8 @@ def _set_trainable(self):
...
+        self.learn.gen_mode = self.gen_first
```
jph00, line 152:
```suggestion
        self.gan_gen_mode = self.gen_first
```

jph00 (changes requested):
Instead of changing things to store in the Learner, let's instead rename to avoid name conflicts. I suggest adding a `gan_` prefix as shown. You'll also need to update the usages of these names throughout the class to add the prefix. Many thanks for your PR!

AndreyYashkin:
My apologies for the late reply. I was too busy last weeks and could not find free time to update and check my patch.

PR #3341, opened 2021-04-29
Updates to AzureMLCallback

yijinlee:
* Add interactive run logging
* Bugfix value extraction for learn.loss.item()
* Convert epoch elapsed time from mm:ss string to seconds (int)
* Update notebook instruction, recommend passing callback in during fitting, instead of directly into learner"


jph00:
Many thanks!

PR #3362, opened 2021-05-06
Add pct feature to get_files

PalaashAgrawal:
Added `pct` argument to get_files and its derivatives (get_image_files, get_text_files, etc). This returns only a certain percent of files from the data when the dataset is too large to experiment with. By default, pct=1., meaning all the files will be extracted (as expected). Using pct, one can get a smaller, more manageable subset of data.

### DIFF
--- a/nbs/05_data.transforms.ipynb
+++ b/nbs/05_data.transforms.ipynb
@@ -4,7 +4,17 @@
    "cell_type": "code",
    "execution_count": null,
    "metadata": {},
-   "outputs": [],
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "\u001b[K     |████████████████████████████████| 204kB 3.8MB/s \n",
+      "\u001b[K     |████████████████████████████████| 61kB 5.4MB/s \n",
+      "\u001b[?25h"
+     ]
+    }
+   ],
    "source": [
     "#hide\n",
     "#skip\n",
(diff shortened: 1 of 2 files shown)


tcapelle:
I am not a big fan of this, it complexify a very used and simple function. `get_files` returns a list, so one can easily call `sample` on it and obtain a smaller one.

nbs/05_data.transforms.ipynb
```diff
@@ -4,7 +4,17 @@
    "cell_type": "code",
```
tcapelle, line 4: <p>Not a fan of adding torch here... just use <code>random.sample</code></p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/fastai/fastai/pull/3362/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

PalaashAgrawal:
> `get_files` returns a list, so one can easily call `sample` on it and obtain a smaller one.

Yes, but I also think that in that case, the relative distribution of categories would be uncertain. For example, if you sample a subset from a dataset that initially had an equal number of files among all categories, now it may or may not be the case. And to ensure that the distribution remains the same, you would have to write a whole block of code including if/else statements and for loops for each category. This would be a seriously long process, especially in the case of extremely large size/cardinal datasets. 

One thing that we can do is write a separate function that's meant just to get a subset of the dataset. Something like `get_subset_files`. And in turn leave this function as is.

tcapelle:
The sampling you are proposing does not solve the distribution issue. It just flips a coin for each file it gets.
The distribution is something that it is no meant to be solved by this type of functions anyway. I think that this type of functionality should be available on the different `DataLoaders` that you want to construct afterwards.

jph00:
Many thanks for the PR. Overall I agree with @tcapelle's view however - I don't think this is the right approach.

PR #3363, opened 2021-05-06
fix 3361

tcapelle:
- Removes casting on segmentation metrics. Adds tests to check it does not come back.
- Adds `ToTensorBase` layer to remove cast.
- Adds this layer at the end of the UNet.
This issue arises from the` __getitem__` that is not implemented on subclasses of `TensorBase`, this is a quick fix but should be probably addressed on `torch_core`.

From the docs: 

> If multiple subclasses exist, the lowest one in the hierarchy will be chosen by default. If there is no unique way to determine such a case, then a TypeError is raised:

So when you index a `TensorImage` with a `TensorMask` it can't resolve the final type.


jph00:
Thanks this makes sense. Possibly it would be even better to have a param where the user passes the type they want to cast to, and have that default to `TensorBase`?

PR #3371, opened 2021-05-12
Check if normalize exists on `_add_norm`

### LINKED ISSUE #3370: Normalize is added twice when using `cnn_learner`
As the title says, when including `Normalize.from_stats` in the `DataLoader` and then trying to use `cnn_learner`, we *double* `Normalize`.

Minimal repro:
```python

def label_func(o): return o[0].isupper()
set_seed(99, True)
path = untar_data(URLs.PETS)/'images'
dls = ImageDataLoaders.from_name_func(
    path, get_image_files(path), valid_pct=0.2,
    label_func=label_func, item_tfms=Resize(224), batch_tfms=[Normalize.from_stats(*imagenet_stats)])

learn = cnn_learner(dls, resnet18, metrics=accuracy)
learn.dls.after_batch
```
You'll see that `learn.dls.after_batch` looks something like this:
```python
Pipeline: IntToFloatTensor -- {'div': 255.0, 'div_mask': 1} -> Normalize -- {'mean': tensor([[[[0.4850]],

         [[0.4560]],

         [[0.4060]]]], device='cuda:0'), 'std': tensor([[[[0.2290]],

         [[0.2240]],

         [[0.2250]]]], device='cuda:0'), 'axes': (0, 2, 3)} -> Normalize -- {'mean': tensor([[[[0.4850]],

         [[0.4560]],

         [[0.4060]]]], device='cuda:0'), 'std': tensor([[[[0.2290]],

         [[0.2240]],

         [[0.2250]]]], device='cuda:0'), 'axes': (0, 2, 3)}
```

And the reason for this is `cnn_learner` always adds in `Normalize` rather than checking if it exists first here: https://github.com/fastai/fastai/blob/master/fastai/vision/learner.py#L157

renato145:
Fixes #3370


muellerzr:
Very nice fix and test! cc: @jph00

fastai/vision/learner.py
```diff
@@ -154,6 +154,7 @@ def _add_norm(dls, meta, pretrained):
     if not pretrained: return
     stats = meta.get('stats')
     if stats is None: return
+    if dls.after_batch.fs.filter(lambda tfm: isinstance(tfm, Normalize)): return
```
jph00, line 157:
```suggestion
    if dls.after_batch.fs.filter(risinstance(Normalize)):
```

https://fastcore.fast.ai/basics.html#risinstance

fastai/vision/learner.py
```diff
@@ -154,6 +154,7 @@ def _add_norm(dls, meta, pretrained):
...
     dls.add_tfms([Normalize.from_stats(*stats)],'after_batch')
```
jph00, line 158: if you change the previous line to `not` then you could move this to the previous line, and get rid of the `return` there

jph00 (changes requested):
Looks great - just a couple of minor suggestions. Please at-mention me when ready to go

renato145:
Ready to go @jph00,  updated with your suggestions.

muellerzr:
I’ve been made aware of that this only adds Normalize to the train, not the
validation. Can you confirm this Renato? And perhaps maybe we can fix that
here?

renato145:
@muellerzr I just tried and it was working fine. Can you share and example of that happening?

muellerzr:
@renato145 thanks for the verification, I believe this may have also been fixed in this latest version (the issue opened up on the forums).

jph00:
Many thanks :)

PR #3374, opened 2021-05-16
remove residual FetchPredsCallback when WandbCallback gets into exception

riven314:
This fix aims to remove `FetchPredsCallback` from Learner when `WandbCallback` get into exception. 
`WandbCallback` adds `FetchPredsCallback` at the beginning of fit and then remove it at the end of fit. However, `WandbCallback` is left after fit when `WandbCallback` get into exception. In my experience, `learner.get_preds` crashed when `FetchPredsCallback` remained. 

A note on notebook cleaning, I tried to run the following before the commit
```
$ nbdev_install_git_hooks
$ nbdev_update_lib
$ nbdev_clean_nbs
```
But after that I found it induced changes in several notebooks, some of them seems not related to my change (as attached):
Screenshot from 2021-05-16 22-32-41
I am not sure if they are the right changes, so I didnt commit those notebook changes. If they are necessary, I can commit those changes as follows.

### DIFF
--- a/fastai/callback/wandb.py
+++ b/fastai/callback/wandb.py
@@ -108,6 +108,7 @@ class WandbCallback(Callback):
                 self.log_predictions(self.learn.fetch_preds.preds)
             except Exception as e:
                 self.log_preds = False
+                self.remove_cb(FetchPredsCallback)
                 print(f'WandbCallback was not able to get prediction samples -> {e}')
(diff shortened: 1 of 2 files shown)


riven314:
updated the PR to sync nb change and lib change

jph00:
Many thanks

PR #3376, opened 2021-05-18
Add Open in Kaggle Badge 

pizzaz93:
This is an example of what adding a Kaggle Badge (that allows for opening the .ipynb as a Kaggle Notebook as seen here https://www.kaggle.com/product-feedback/152480) would look like.

If it makes sense I can submit a further PR adding it to other notebooks in the repo.


muellerzr:
I think this might be better off in nbdev, where all the badges are handled.

cc @hamelsmu

pizzaz93:
Thank you @muellerzr.
Would you be referring to the notebooks in https://github.com/fastai/nbdev/tree/master/nbs?

PR #3383, opened 2021-05-21
changes __reduce_ex__ so that metadata is preserved when serializing

### LINKED ISSUE #3382: Multiprocessing is not maintaining any metadata
Prime example of this behavior:

`TensorPoint`'s keep track of the `img_size` of some tensor When `num_workers` is set to zero, said image size can be seen on the output tensor, but when it is 1+, this is gone.

Attached are two minimal gists, one is from myself and DomMK on the fastai discord:

TensorPoint example: https://gist.github.com/muellerzr/cde7943738848eb3e435dbbb0391be9d

Generic example (by Dom): https://gist.github.com/muellerzr/cde7943738848eb3e435dbbb0391be9d

As well as minimal code to get the TensorPoint issue is below:

```python

path = untar_data(URLs.BIWI_HEAD_POSE)
img_files = get_image_files(path)
def img2pose(x): return Path(f'{str(x)[:-7]}pose.txt')
cal = np.genfromtxt(path/'01'/'rgb.cal', skip_footer=6)
def get_ctr(f):
    ctr = np.genfromtxt(img2pose(f), skip_header=3)
    c1 = ctr[0] * cal[0][0]/ctr[2] + cal[0][2]
    c2 = ctr[1] * cal[1][1]/ctr[2] + cal[1][2]
    return tensor([c1,c2])

biwi = DataBlock(
    blocks=(ImageBlock, PointBlock),
    get_items=get_image_files,
    get_y=get_ctr,
    splitter=FuncSplitter(lambda o: o.parent.name=='13'),
    batch_tfms=[*aug_transforms(size=(240,320)), 
                Normalize.from_stats(*imagenet_stats)])

dls_a = biwi.dataloaders(path, num_workers=0)
dls_b = biwi.dataloaders(path, num_workers=1)

o = next(iter(dls_a[0]))
print(o[1].img_size)

o = next(iter(dls_b[0]))
print(o[1].img_size)

### PR DESCRIPTION
**Pull request by arampacha**
Closes #3382 

Note: when requirements are changed to pytorch>1.8 the customization of `__reduce_ex__` probably can be removed as native `Tensor.__reduce_ex__` loads `__dict__` for Tensor subclasses properly

### CODE REVIEW

**Comment by riven314** [comment]
I think the subclassing in this line is redundant:
```
ret = func(*args).as_subclass(type)
```

unlike pytorch, the `func` in fastai is either `_fa_rebuild_tensor` or `_fa_rebuild_qtensor`, both of which returns subclass instead of torch.Tensor. 

So would it be better to make the change like this?
```
def _rebuild_from_type(func, args, dict):
    ret = func(*args)
    ret.__dict__ = dict
    return ret

# TensorBase
  def __reduce_ex__(self,proto):
      torch.utils.hooks.warn_if_has_hooks(self)
      args = (type(self), self.storage(), self.storage_offset(), tuple(self.size()), self.stride())
      if self.is_quantized: args = args + (self.q_scale(), self.q_zero_point())
      args = args + (self.requires_grad, OrderedDict())
      f = _fa_rebuild_qtensor if self.is_quantized else  _fa_rebuild_tensor
      return (_rebuild_from_type, (f, args, self.__dict__))
```

arampacha:
Makes sense. Another option would be to remove `_fa_rebuild_*` functions to keep it consistent with what's done in pytorch if it's ok?

jph00:
Many thanks!

PR #3386, opened 2021-05-27
add DiceLoss

tcapelle:
Add DiceLoss for segmentation tasks.
It is very useful to combine this type of loss with CE or FocalLoss, used in various kaggle comps.


fastai/losses.py
```diff
@@ -122,4 +122,24 @@ class LabelSmoothingCrossEntropyFlat(BaseLoss):
...
+    "Target mask to one hot"
+    return torch.stack([torch.where(x==c, 1,0) for c in range(classes)], axis=axis)
+
+# Cell
+class DiceLoss:
+    "Dice loss for segmentation"
+    def __init__(self, axis=1, smooth=1):
+        store_attr()
+    def __call__(self, pred, targ):
+        targ = _one_hot(targ, pred.shape[self.axis])
```
jph00, line 138: Why not use the `one_hot` from `torch_core`?

nbs/01a_losses.ipynb
```diff
@@ -414,6 +414,113 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "dl = DiceLoss()\n",
+    "output = torch.randn(32, 4, 5, 10)\n",
+    "target = torch.randint(0,2,(32, 5, 10))\n",
+    "_ = dl(output, target)"
```
jph00, line 467: Could you add some simple example tests that confirm the loss is correct? Currently your tests aren't actually testing anything.

jph00 (changes requested):

fastai/losses.py
tcapelle: I know we have one, but is does not work the same. I would need to modify the other one to be able to create the one-hot segmentation masks.

nbs/01a_losses.ipynb
tcapelle: Would you mind helping me with this? I added some test for the `one_hot` and that the decodes work, but how would you unit test this? failing mostly happen with `flatten_check` so I would just be testing that.

fastai/losses.py
tcapelle: the `DiceMulti` metric already does this:
```python
        for c in range(learn.pred.shape[self.axis]):
            p = torch.where(pred == c, 1, 0)
```

nbs/01a_losses.ipynb
jph00: I'd test stuff like two masks that were identical, two which were entirely non-overlapping, and two which overlapped 50% exactly.

fastai/losses.py
jph00: What's the difference needed in functionality exactly? Is there some different name you could use for yours that makes it clear? And/or some comment you could add?

tcapelle: I could make it a class method, I just put it outside to test it =P

nbs/01a_losses.ipynb
tcapelle: good idea

tcapelle: done.

tcapelle:
It is ready to review @jph00 (I initially solved the CI that was not passing by cleaning tutorial.text, but maybe was not a good idea).

jph00:
Many thanks! :)

PR #3392, opened 2021-06-01
Fixes GANLearner default alternation

### LINKED ISSUE #3390: Change default behavior GANLearner switcher
**Describe the bug**
By default the GANLearner switcher is the FixedGANSwitcher with n_crit=5 & n_gen=1 but the original GAN paper (https://arxiv.org/pdf/1406.2661.pdf) recommends an alternation for the switcher with k=n_crit=n_gen=1

**To Reproduce**
Default FixedGANSwitcher is:
https://github.com/fastai/fastai/blob/301016c5d3de2bdb5269121bd0716538d85f7409/fastai/vision/gan.py#L305

in

https://github.com/fastai/fastai/blob/301016c5d3de2bdb5269121bd0716538d85f7409/fastai/vision/gan.py#L299-L309

**Expected behavior**
I suggest that the default behavior should be k=n_crit=n_gen=1 to be consistend with the original work of Ian Goodfellow et al.
`if switcher is None: switcher = FixedGANSwitcher()`

StFroese:
fixes #3390 
Changed default iterations for GANLearner to n_crit=n_gen=1 and use n_crit=5 case for GANLearner.wgan


StFroese:
Thanks! :)

Issue #3396, opened 2021-06-02
DataLoaders.from_dsets: Transforms specified in kwargs are missing in validation's TfmdDL


riven314:
An elaboration of the issue https://github.com/fastai/fastai/issues/3373

**Versions**
- fastai: 2.3.1
- fastcore: 1.3.20
- torch: 1.8.1+cu101

**Describe the bug**
When using `DataLoaders.from_dsets`, we can add transforms into different stages (e.g. `after_item`, `after_batch`) by specifying them as keyword arguments. Those transforms should be casted to both train's `TfmdDL` and valid's `TfmdDL`. However, we found the transforms only got applied in train's `TfmdDL`. 

**To Reproduce**
```
from fastai.vision.all import *
mnist = DataBlock(blocks=(ImageBlock(cls=PILImageBW), CategoryBlock), 
                  get_items=get_image_files, 
                  splitter=GrandparentSplitter(),
                  get_y=parent_label)
train_ds = mnist.datasets(untar_data(URLs.MNIST_SAMPLE))
valid_ds = mnist.datasets(untar_data(URLs.MNIST_SAMPLE))
dls = DataLoaders.from_dsets(train_ds, valid_ds, 
                             after_item=[ToTensor], 
                             after_batch=[IntToFloatTensor, Normalize.from_stats(*imagenet_stats)])
```

**Expected behavior**
1. `dls.train.after_item` and `dls.valid.after_item` should have the same pipeline of transforms, but we see `dls.valid.after_item` is an empty `Pipeline`
2. same issue for `dls.train.after_batch` and `dls.valid.after_batch`

tcapelle:
can you check on master if it is fixed now?

Issue #3373, opened 2021-05-13
DataLoaders.from_dsets does not apply after_item transforms to the validation dataset


backdoorbreaker:
I'm admittedly fairly new to fastai2, but I did search in the forums for the `from_dsets` keyword and found no useful results. 

Please confirm you have the latest versions of fastai, fastcore, and nbdev prior to reporting a bug (delete one): YES

**Describe the bug**
When the `DataLoaders.from_dsets` classmethod is used, the `after_item` transforms are not applied to the validation dataset. This results in a `TypeError` within the `fa_collate` function because in my specific use case, the `ToTensor` transform is not called on my validation dataset which consists of `PILImage` types. This `TypeError` occurs at the start of the validation portion of the training loop.

This is different than the behavior of some of the other functions which accept `after_item`. (e.g. the `dataloaders` instance method on `TfmdLists`)

When inspecting the source code for [`DataLoaders.from_dsets`](https://github.com/fastai/fastai/blob/master/fastai/data/core.py#L182), we can see that the function builds up a list of `kwargs` for each of the datasets. When printing these `kwargs` out, `after_item: Pipeline: ToTensor` appears in the `kwargs` for both the training and validation dataset which seems to contradict the behavior that I see.

**To Reproduce**
It's a bit challenging to exactly reproduce my local setup, but I will describe how to view similar behavior in the siamese tutorial.
1. Open the [siamese tutorial colab](https://colab.research.google.com/github/fastai/fastai/blob/master/nbs/24_tutorial.siamese.ipynb)
2. Run the cells up to cell 34 (at the time of writing). cell contents: `dls = DataLoaders.from_dsets(...)`
3. Add an `after_item=[ToTensor]` to that `from_dsets` call
4. Print out `dls.train.after_item` and `dls.valid.after_item`
5. You should see that `dls.valid.after_item` is an empty pipeline

**Expected behavior**
I expected `ToTensor` to also be applied to my validation dataset during training.

It was simple enough to manually add the transform back in like so:
```
dls.valid.after_item = Pipeline([ToTensor])
```
Everything worked with that addition - I can't help but feel like I'm doing something incorrect though.

**Error with full stack trace**
```
    learner.fit_one_cycle(1, lr_steep)
  File "/usr/local/lib/python3.6/dist-packages/fastai/callback/schedule.py", line 112, in fit_one_cycle
[...]

PR #3397, opened 2021-06-02
changes DataLoaders.from_dsets

arampacha:
Linked issues #3396 #3373 
@riven314 and myself came up with possible solution. This PR changes `DataLoaders.from_dsets` to resamble `Datasets.dataloaders` ([link](https://github.com/fastai/fastai/blob/301016c5d3de2bdb5269121bd0716538d85f7409/fastai/data/core.py#L221)). This resolves aforementioned issue and allows to pass custom arguments to valid dataloader (e.g. val_res for `SortedDL`).

Open questions:
- How do we deal with situation when only one dataset is passed to constructor. As of now it will trigger error when trying to access `dls.valid`?
- It's still not quite clear what exactly was causing the issue. It was not observed with `do_setup=False` so probably something went wrong when setting up `Pipeline`s. But `Pipeline.setup` in isolation doesn't seem to cause same effect.

### DIFF
--- a/nbs/03_data.core.ipynb
+++ b/nbs/03_data.core.ipynb
@@ -585,7 +585,7 @@
     "        val_kwargs={k[4:]:kwargs.pop(k) for k in list(kwargs.keys()) if k.startswith('val_')}\n",
     "        def_kwargs = {'bs':bs,'shuffle':shuffle,'drop_last':drop_last,'device':device}\n",
     "        dl = dl_type(ds[0], **merge(def_kwargs, kwargs))\n",
-    "        def_kwargs = {'bs':bs if val_bs is None else val_bs,'shuffle':val_shuffle,'drop_last':False}\n",
+    "        def_kwargs = {'bs':val_bs,'shuffle':val_shuffle,'drop_last':False}\n",
     "        dls = [dl] + [dl.new(ds[i], **merge(def_kwargs,kwargs,val_kwargs))\n",
     "                      for i in range(1,len(ds))]\n",
     "        return cls(*dls, path=path, device=device)\n",
@@ -683,23 +683,12 @@
    "cell_type": "code",
    "execution_count": null,
    "metadata": {},
-   "outputs": [
-    {
-     "data": {
-      "text/plain": [
-       "(Pipeline: _TestTfm, Pipeline: _TestTfm)"
-      ]
-     },
-     "execution_count": null,
-     "metadata": {},
-     "output_type": "execute_result"
-    }
-   ],
+   "outputs": [],
    "source": [
     "dls3 = DataLoaders.from_dsets(start, start, bs=4, val_bs=8, after_item=[_TestTfm()], val_after_batch=[_TestTfm()])\n",
     "test_eq(len(dls3.train.after_batch.fs),0)\n",
     "test_eq(len(dls3.valid.after_batch.fs),1)\n",
-    "dls.train.after_item, dls.valid.after_item"
+    "test_eq(len(dls3.train.after_item.fs), len(dls3.valid.after_item.fs))"
    ]
   },
   {
--- a/nbs/03_data.core.ipynb
+++ b/nbs/03_data.core.ipynb
@@ -357,7 +357,7 @@
     {
      "data": {
       "text/markdown": [
-       "<h4 id=\"DataLoader.one_batch\" class=\"doc_header\"><code>DataLoader.one_batch</code><a href=\"https://github.com/fastai/fastai/tree/master/fastai/data/load.py#L148\" class=\"source_link\" style=\"float:right\">[source]</a></h4>\n",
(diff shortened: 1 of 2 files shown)


nbs/03_data.core.ipynb
```diff
@@ -674,6 +679,29 @@
...
+     "execution_count": null,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "dls3 = DataLoaders.from_dsets(start, start, bs=4, val_bs=8, after_item=[_TestTfm()], val_after_batch=[_TestTfm()])\n",
+    "test_eq(len(dls3.train.after_batch.fs),0)\n",
+    "test_eq(len(dls3.valid.after_batch.fs),1)\n",
+    "dls.train.after_item, dls.valid.after_item"
```
riven314, line 702: personally I think its better for this line to be a test:  
```
test_eq(len(dls.train.after_item.fs), len(dls.valid.after_item.fs))
```

Additionally, shall we also test `bs` is set properly on train & valid?  
```
test_eq(dls.train.bs, 4)
test_eq(dls.valid.bs, 8)
```

jph00:
Why not just use `deepcopy`? A smaller change like that would be easier for me to review. Are there significant downsides?

arampacha:
Using `deepcopy` was the first solution that came to both @riven314 and me and it fixes the issue :)

The solution proposed in this PR has advantage of handling kwargs specific to `valid` dataloader. Example use case when this can be useful - providing `val_res` for speeding up `SortedDL` creation. It is possible to modify current version of `.from_dsets` to take `val_*` keyword arguments as well. But that will also introduce substantial changes to review.
The proposed version makes `Dataloaders.from_dsets` more consistent with `Datasets.dataloaders`. These two have similar purpose and structure. In my opinion this should make those easier to maintain.

There is also an option to construct `Datasets` object from datasets passed to `Dataloaders.from_dsets` and use its `.dataloaders` method. But it might be unnecessary complication for the task. I can give it a try if you think it is a worthy idea.

tcapelle:
why not just do something like this:
```
    @classmethod
    def from_dsets(cls, *ds, path='.',  bs=64, device=None, dl_type=TfmdDL, **kwargs):
        default = (True,) + (False,) * (len(ds)-1)
        defaults = {'shuffle': default, 'drop_last': default}
        tfms = {k:tuple(Pipeline(kwargs[k]) for i in range_of(ds)) for k in _batch_tfms if k in kwargs}
        kwargs = merge(defaults, {k: tuplify(v, match=ds) for k,v in kwargs.items() if k not in _batch_tfms}, tfms)
        kwargs = [{k: v[i] for k,v in kwargs.items()} for i in range_of(ds)]
        return cls(*[dl_type(d, bs=bs, **k) for d,k in zip(ds, kwargs)], path=path, device=device)
```

arampacha:
@tcapelle, from previous message:
> The proposed version makes Dataloaders.from_dsets more consistent with Datasets.dataloaders. These two have similar purpose and structure. In my opinion this should make those easier to maintain.

I think this can help to avoid the kind of issues when we have an unexpected behavior in one method, while other method with similar functionality works fine. But that's just my opinion and it might be wrong. I can close the PR if proposed changes are not considered useful.

tcapelle:
Maybe `from_dsets` should not exist, and the user should be pushed towards creating explicitly their dataloaders with transforms using the `TfmDL` constructor. If you have plain datasets without transforms on the input, the needed `kwargs` on `from_dsets` can quickly become messy if you want to customize the full pipeline.

tcapelle:
The `from_df` method need reworking, maybe adding delegates also to replace the `**kwargs.`  Probably the `_batch_tfms = ('after_item','before_batch','after_batch' )` should also be explicit for train and valid. If we do so, this method would end up with like 15 arguments. Don't know what the right solution is, but the actual one is confusing and not expressive enough.

arampacha:
Using delegates here is tricky. We want to delegate to `dl_type` which is passed as argument itself. Delegating to `TfmdDL` may be confusing if one uses other dataloader type.
The version proposed here allows to pass transforms as:
```
dls = DataLoaders.from_dsets(train_ds, valid_ds, 
                             after_item=[TrainOnlyTfm, ToTensor],
                             val_after_item=[ToTensor], 
                             after_batch=[IntToFloatTensor, Normalize.from_stats(*imagenet_stats)])
```
Keyword args starting with `val_` will be used for valid dataloader and if no `val_` argument provided, regular argument will be used for both train and valid dl. Knowing this one can finely control the behavior of `from_dsets`. This pattern is used in other places too, so it shouldn't be completely unfamiliar. But I agree this might be not ideal and possibly better solutions can be found.
Would you share examples where this is not expressive enough? We can try to improve solution based on that.

Also I remember Jeremy mentioned it's possible that v3 can possibly switch to using native pytorch `DataLoader`. If that's correct it will impact methods like `.from_sets`.

PR #3399, opened 2021-06-03
Updated layer_info to work with seq2seq huggingface transformer models

ohmeow:
Tested with ohmeow-blurr.  This simple modification allows folks to use  with huggingface seq2seq tranformer models.


jph00:
Many thanks

PR #3401, opened 2021-06-06
Add `log_freq` param to WandbCallback

andrewtruong:
Previously, `WandbCallback` used the default `log_freq=1000`, which may prevent any grads or params from being logged if there are too few steps.


jph00:
Thanks @andrewtruong . Can you please follow the steps here to clean your notebook and build the lib? https://docs.fast.ai/dev-setup.html

Issue #3364, opened 2021-05-07
interp.plot_top_losses() not working with v 2.3.1


polyrand:
Please confirm you have the latest versions of fastai, fastcore, and nbdev prior to reporting a bug (delete one): YES

Versions installed:

```python
import fastcore
fastcore.__version__
# '1.3.20'

import torch
torch.__version__
# '1.8.1'

import torchvision
torchvision.__version__
# '0.9.1'

import fastai
fastai.__version__
# '2.3.1'
```

**Describe the bug**
The exact same notebook / experiment stopped working after updating to fastai v 2.3.1

**To Reproduce**

* Train resnet18 cnn_learner
* `interp = ClassificationInterpretation.from_learner(learn)`
* `interp.plot_top_losses(5)`

Full example [here](https://gist.github.com/polyrand/66cd557b25a40651e6c0a89a94047fb1)

The other `interp` functions I have used seem to be working (`most_confused`, `print_classification_report`, `plot_confusion_matrix`)

**Expected behavior**

**Error with full stack trace**

Place between these lines with triple backticks:

```
----> 1 interp.plot_top_losses(5)

~/miniconda3/envs/torch2/lib/python3.7/site-packages/fastai/interpret.py in plot_top_losses(self, k, largest, **kwargs)
     36         losses,idx = self.top_losses(k, largest)
     37         if not isinstance(self.inputs, tuple): self.inputs = (self.inputs,)
---> 38         if isinstance(self.inputs[0], Tensor): inps = tuple(o[idx] for o in self.inputs)
     39         else: inps = self.dl.create_batch(self.dl.before_batch([tuple(o[i] for o in self.inputs) for i in idx]))
     40         b = inps + tuple(o[idx] for o in (self.targs if is_listy(self.targs) else (self.targs,)))

~/miniconda3/envs/torch2/lib/python3.7/site-packages/fastai/interpret.py in <genexpr>(.0)
     36         losses,idx = self.top_losses(k, largest)
     37         if not isinstance(self.inputs, tuple): self.inputs = (self.inputs,)
---> 38         if isinstance(self.inputs[0], Tensor): inps = tuple(o[idx] for o in self.inputs)
     39         else: inps = self.dl.create_batch(self.dl.before_batch([tuple(o[i] for o in self.inputs) for i in idx]))
     40         b = inps + tuple(o[idx] for o in (self.targs if is_listy(self.targs) else (self.targs,)))

[...]

PR #3405, opened 2021-06-17
Fix #3404 #3364

### LINKED ISSUE #3404: `SegmentationInterpretation.plot_top_losses` ignores `vmin`, keeps it set to 1
When using a simple unet for image segmentation with a binary mask (which has only two values, 0 and 1), I noticed that `plot_top_losses` shows the mask and prediction as a solid blue rectangle, although the mask obviously contains something.

I am using fastai v2 in the latest release on my own hardware. OS: Ubuntu 20.04 with CUDA 11.3 and pytorch 1.8.1

*How to reproduce*:
- Create a new unet learner using the high-level API (as described [here](https://docs.fast.ai/tutorial.vision.html#Segmentation---Using-the-high-level-API))
- Use custom generated `ImageMask`s, with only two values, 0 and 1.
- After training the model, create a `SegmentationInterpretation` using the `from_learner` function.
- View the output of `plot_top_losses`. The masks and predictions look blank when they are not.

*Expected behaviour*:
- When using `SegmentationDataLoaders.show_batch` with the same mask data, I can pass an argument `vmin=0`. This causes the values labeled with 1 to be displayed in color.
- Not passing `vmin=0` also shows the mask as blank (see also [this forum post](https://forums.fast.ai/t/segmentationdataloaders-show-batch-not-showing-mask-in-other-color/66071/2))
- I would expect the same behaviour for `SegmentationInterpretation.plot_top_losses`. However, I can pass the argument `vmin=0`, but it is ignored.

*Known Workaround*:
- I could mitigate the problem by using the value 2 in my `ImageMask` instead of 1. While this works, I find it irritating. Being able to use the same parameters as for `show_batch` would be more consistent.  I also seem (?) to have to include a third bogus value in my `codes` list for the dataloader to avoid `AssertionErrors` when training the learner, however I was not able to verify this.

tcapelle:
Add a missing `kwargs` to top losses of `TensorMask`


jph00:
Many thanks

PR #3407, opened 2021-06-20
fixed typo in text tutorial

ama-gii:
Creates a TextDataLoaders with text_vocab but doesn't use on text_classifier_learner.


jph00:
Thanks!

PR #3408, opened 2021-06-20
Fix typos in LR finder suggestion methods docs

M-Salti:
This PR fixes the default algorithm for `Learner.lr_find` in the docs from `steep` to `valley`, and removes a little typo.


jph00:
Thanks

PR #3414, opened 2021-07-01
Add EndSplitter

tcapelle:
Add a simple Splitter to split train/valid keeping valid at the end. Useful for ordered data like timeseries, text, etc...

```python
EndSplitter(0.2, valid_last=True)(range(10))
>> ([0,1,2,3,4,5,6,7], [8,9])
```


jph00:
How about changing it to `EndSplitter(..., valid_last=True)`? That way it can split at either end...

jph00:
Ah apologies - that `'base'` error isn't actually anything to do with your code, it's caused by the latest pandas. I've fixed that now.

Issue #3418, opened 2021-07-02
ML_100k - Broken Link, needs to be HTTPS


mmichelli:
ML_100k has stopped working, it needs to be HTTPS 

https://github.com/fastai/fastai/blob/ab154927696338741e59e0ffc4774777c4a9781c/fastai/data/external.py#L83

'https://files.grouplens.org/datasets/movielens/ml-100k.zip'

PR #3419, opened 2021-07-02
Fix ML_100k Link 

mmichelli:
Fixed the ML_100k link, HTTP needs to be HTTPS
https://github.com/fastai/fastai/issues/3418


jph00:
Many thanks.

PR #3427, opened 2021-07-08
Cast lr returned by valley and slide as float

vtecftwy:
lr suggestion returned by slide and valley is a tensor while lr suggested by minimum and steep are floats.
To keep consistency, and after exchange on the forum, this simple PR casts the value returned by valley and slide as float.

### DIFF
--- a/fastai/callback/schedule.py
+++ b/fastai/callback/schedule.py
@@ -218,7 +218,7 @@ def valley(lrs:list, losses:list, num_it:int):
     sections = (max_end - max_start) / 3
     idx = max_start + int(sections) + int(sections/2)
 
-    return lrs[idx], (lrs[idx], losses[idx])
+    return lrs[idx].item(), (lrs[idx].item(), losses[idx])
 
 # Cell
 def slide(lrs:list, losses:list, num_it:int, lr_diff:int=15, thresh:float=.005, adjust_value:float=1.):
@@ -234,7 +234,7 @@ def slide(lrs:list, losses:list, num_it:int, lr_diff:int=15, thresh:float=.005, 
         r_idx -= 1
         l_idx -= 1
 
-    suggestion = local_min_lr * adjust_value
+    suggestion = local_min_lr.item() * adjust_value
     idx = np.interp(np.log10(suggestion), np.log10(lrs), losses)
     return suggestion, (suggestion, idx)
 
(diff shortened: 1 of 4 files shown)


fastai/data/external.py
```diff
@@ -80,7 +80,7 @@ class URLs():
     IMDB                = f'{S3_NLP}imdb.tgz'
     IMDB_SAMPLE         = f'{URL}imdb_sample.tgz'
     ML_SAMPLE           = f'{URL}movie_lens_sample.tgz'
-    ML_100k             = 'https://files.grouplens.org/datasets/movielens/ml-100k.zip'
+    ML_100k             = 'http://files.grouplens.org/datasets/movielens/ml-100k.zip'
```
jph00, line 83: I believe this change should be reverted - https is the correct protocol to use AFAIK.

fastai/callback/schedule.py
```diff
@@ -218,7 +218,7 @@ def valley(lrs:list, losses:list, num_it:int):
     sections = (max_end - max_start) / 3
     idx = max_start + int(sections) + int(sections/2)
 
-    return lrs[idx], (lrs[idx], losses[idx])
+    return lrs[idx].item(), (lrs[idx].item(), losses[idx])
```
jph00, line 221: I think this will fail if the suggestion isn't a tensor (which it might not be). Instead I think this should always work:

```suggestion
    return float(lrs[idx]), (float(lrs[idx]), losses[idx])
```

fastai/callback/schedule.py
```diff
@@ -234,7 +234,7 @@ def slide(lrs:list, losses:list, num_it:int, lr_diff:int=15, thresh:float=.005,
         r_idx -= 1
         l_idx -= 1
 
-    suggestion = local_min_lr * adjust_value
+    suggestion = local_min_lr.item() * adjust_value
```
jph00, line 237:
```suggestion
    suggestion = float(local_min_lr) * adjust_value
```

jph00 (changes requested):
Looking pretty good now! Just a couple of requests to look at please.

fastai/data/external.py
vtecftwy: Yes, indeed. I did not realize there was a change made there.

fastai/callback/schedule.py
vtecftwy: Yes. Will use that and try to run the test on my computer before next step.

vtecftwy:
Updated the code to float() and reverted to https for movielens. I used two separate commit because these are two different issues.
Tests that failed above passed on my computer ` nbdev_test_nbs --flags '' --pause 1.0 --fname "nbs/[0-2][0-2]*.ipynb"`
Hope all is good now

jph00:
Many thanks!

PR #3429, opened 2021-07-13
Fix Pillow dtype issue

tcapelle:
The CI is broken with latest pillow.
pillow 8.3.1 has been released (does not fix the isse either), ~~but it is not available on all channels yet.~~
modified `tensor` to cast to array before converting to tensor on images.

We are not that impacted as we use our own custom tensor transform.


jph00:
@tcapelle this change impacts stuff other than just images, and breaks the tests (as you see from CI). Can you suggest an alternative that is more constrained?

jph00:
OK sorted that out myself - turns out that line wasn't needed at all! :)

tcapelle:
how did you sorted this out,? my changes did not break the test, it was the pandas thingy....

PR #3433, opened 2021-07-19
Fix tabular_learner to accept custom embedding sizes

rbrtwlz:
There is a typo in tabular_learner, which makes it impossible to set custom embedding sizes.


rbrtwlz:
Since my modification of tabular_learner contains an embedding matrix, which is not associated to a class in to (TabularObject) I tried to set embed_szs as a list of tuples:
`emb_szs_list = [(10,6),(17,8),(8,5),(16,8),(7,5),(6,4),(4,4)]`

But when converting the list to a dict
`emb_szs_dict = {to.cat_names[i]:t[1] for i,t in enumerate(emb_szs_list)}`
it works, like intended.

So my fix just fixes a problem I had with a modification of tabular learner...

jph00:
Thanks for letting us know.

Issue #3403, opened 2021-06-16
Depreciate Mish JIT support for native PyTorch 1.9 version


digantamisra98:
With Mish now being supported out of the box in PyTorch 1.9, the JIT version present in Fast AI should be removed to keep it synchronous and use the [`nn.Mish()`](https://pytorch.org/docs/stable/generated/torch.nn.Mish.html?highlight=mish#torch.nn.Mish) version instead. 

Happy to submit a PR if this will be approved.

tcapelle:
you made Mish to nn, congratz!

digantamisra98:
Thank you! @tcapelle

jph00:
Thanks @digantamisra98. In order to avoid changing the version dependency, how about having fastai delegate to the PyTorch version if it's 1.9 or greater, and use its internal implementation otherwise? If that sounds OK, I'd be happy to take a PR that does that.

digantamisra98:
Sure, that makes sense. I will submit a PR at the earliest, thanks!

tmabraham:
@jph00 I think this issue can be closed thanks to your commit https://github.com/fastai/fastai/commit/f09e11907646d262f5161176045aa575891a244d

PR #3436, opened 2021-07-22
Updated Mish to Torch 1.9

digantamisra98:
Fixes [3403](https://github.com/fastai/fastai/issues/3403)


jph00:
@digantamisra98 looks like you may need to update your version of nbdev and rebuild

digantamisra98:
@jph00 did a pip upgrade and shows that I'm on the latest version. Should I build from source?

jph00:
> @jph00 <https://github.com/jph00> did a pip upgrade and shows that I'm on the latest version. Should I build from source?
> 

That shouldn't be needed. I think somehow you have 2 versions installed or something. Or maybe you've got multiple environments.

I just released a new version which you could try, but the difference you're seeing is from code that was removed many months ago.

digantamisra98:
@jph00 Installed the latest NBDev and did a rebuild. Hopefully this is good now.

jph00:
Oh I think I know what's happening. Your notebooks have `nbdev_comment` lines added. Perhaps at some point you did `nbdev_update_lib` on all the nbs, with an old version of nbdev, so now all your NBs have been changed.

Would you prefer I just made the change manually on my end? Otherwise, might be easiest for you to just delete your local fastai folder and then make the change in the NB and build the lib.

digantamisra98:
Oh okay, sure, if that's convenient for you then that would be great, otherwise I'm happy to follow your recommended steps.

jph00:
Ouch that sure didn't work! I'll do it over here - thanks :)

digantamisra98:
Thanks!

jph00:
I've just pushed that change - please test it when you have a moment.

digantamisra98:
Looks good to me, thanks!

PR #3439, opened 2021-07-26
Bugfix for DiceLoss

thomasfermi:
Hi all, when I worked on this [kaggle notebook](https://www.kaggle.com/thomasfermi/fastai-sample-solution) I found that the fastai DiceLoss is not working properly. When I was using it, the network achieved a very low loss, but it was not learning to do proper segmentation. Basically everything was predicted to be "background class". I read the code and found that the current implementation sums over all dimensions (batch_dimension, class_dimension, image_width, image_height). However it shall only sum over the dimensions image_width and image_height. It should take the mean over the class_dimension! Otherwise it is not working properly, especially in case of class imbalance.

I fixed the problem with this PR. I had to remove a test case, and introduced several new test cases together with some quite long explanations in the notebook. Let me know if that explanation is too long, or if there are any other issues. This is my first PR to fastai and I am new to nbdev.


thomasfermi:
I am seeing some errors in the github actions regarding nbdev. It is complaining about cfg_path. I had the same issue with my local install until I installed an older version of nbdev (1.1.11) in my conda environment. Maybe there is some issue with the current version of nbdev?

jph00:
@thomasfermi this is looking great! Can you please explain why you needed to remove the test that was already there? Was it incorrect?

Apologies for the errors in CI - if you merge the latest changes from master that should fix them.

jph00:
> I am seeing some errors in the github actions regarding nbdev. It is complaining about
> 
> `__init__() missing 2 required positional arguments: 'cfg_path' and 'cfg_name'`
> 
> I had the same issue with my local install until I installed an older version of nbdev (1.1.11) in my conda environment. Maybe there is some issue with the current version of nbdev?

The latest nbdev also requires the latest fastcore. I've updated nbdev now to ensure that the correct fastcore is installed.

tcapelle:
let me check this tomorrow @jph00

tcapelle:
I am currently working on this, would you like to join the fastai discord and solve this together?

thomasfermi:
Hi all, 

@jph00: I removed the test case because the value it was asking for was wrong in my opinion (because I think that "mean" should be used instead of "sum"). Also the example did not really have the form of an image, so I thought that it did not make much sense.

@tcapelle : You are right that "mean" vs "sum" only leads to a scale factor and hence will not affect the learning (I was wrong in this regard). Your suggestion with the 'reduction' argument sounds good. I would prefer to set 'reduction' to 'True' per default since it makes the interpretation of the loss a bit easier.
I looked into my kaggle dataset again and found that the other change that I had made to the loss function is more important. I am squaring the terms in the denominator as I had learned in [this blog post](https://www.jeremyjordan.me/semantic-segmentation/). 
If I do not square, then I get super bad performance on my dataset. Someone else seems to have experienced the same issue on another dataset with class imbalance (see [stackoverflow](https://stackoverflow.com/questions/66536963/dice-loss-working-only-when-probs-are-squared-at-denominator))
In the [monai library](https://docs.monai.io/en/latest/_modules/monai/losses/dice.html#DiceLoss) there is a toggle 'squared_pred' to switch on the squaring. I implemented such a toggle as well. Please let me know what you think of the newest version.

P.S. Just saw your message regarding fastai discord after I updated this PR. I will try to join the fastai discord now.

thomasfermi:
@tcapelle : This is my latest version taking into account our discussion here and on discord. Please let me know what you think.

tcapelle:
I would do some modifications:
- I would make method to check and flatten the vectors, it is important to check that the tensors have the same size and raise an error. Something like this:
```python
    def flatten_check_channel(self, p, t):
        "keeps channels dim first"
        channels = p.shape[self.axis]
        def _flat(x): return TensorBase(x.transpose(0,self.axis).contiguous()).view(channels,-1)
        p,t = map(_flat, [p,t])
        assert p.shape == t.shape, 'input and target dimensions differ, DiceLoss expects non one-hot targs'
        return p, t
```
- I would make the defaults different, I did some testing today, and ideally you want the loss to have values that are not too small. So I would sum the dice coefficients over the channels and average over the batch.
- smooth should be 1e-6 as default.
- square union should be off by default ( I would remove this altogether as for me it makes no sense), also square the target is not necessary, only the preds.
- You need to cast the tensors to `TensorBase` to avoid methods incomptaiblity between tensor subclasses.

thomasfermi:
Hi @tcapelle, thanks for your suggestions! I incorporated most of them, but I did not add the 'flatten_check_channel' method. 
Instead I added a casting to TensorBase and a shape-assert into the __call__ method. The code is simpler that way in my opionion (no transpose, contiguous, and view). Please have a look at the updated code and let me know what you think!

Regarding defaults: 
* Why would you want to sum over the channels but then take the mean over the batch? We can just sum over everything in this case, can't we?
* This paper here has the squares in the union: https://arxiv.org/abs/1606.04797 (it might even be the paper that introduced dice loss in deep learning). Also empirically with my dataset, my unet does not learn properly without the square. Did you try with and without square for your application? Did it make a difference?

tcapelle:
@jph00  you can merge this

PR #3449, opened 2021-08-08
bugfix for learner summary

lispsil:
In Chapter 4: Under the Hood: Training a Digit Classifier (04_mnist_basics.ipynb) Jeremy creates Dataloaders from lists:

```python
train_x = torch.cat([stacked_threes, stacked_sevens]).view(-1, 28*28)
train_y = tensor([1]*len(threes) + [0]*len(sevens)).unsqueeze(1)
dset = list(zip(train_x,train_y))
dl = DataLoader(dset, batch_size=256)
```

as a result, there is no Dataset object and we can't use `summary` method of a learner

Here is a fix (and a test) that works the same for objects that have "n_inp" attribute and work with lists if there is no.

### DIFF
--- a/fastai/callback/hook.py
+++ b/fastai/callback/hook.py
@@ -201,7 +201,7 @@ def module_summary(learn, *xb):
 @patch
 def summary(self:Learner):
     "Print a summary of the model, optimizer and loss function."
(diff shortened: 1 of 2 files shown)


jph00:
Nice PR - thanks!

PR #3450, opened 2021-08-08
Tutorial for inteference

jdupl123:
At the moment Fast the moment the documentation is a bit spare on production so I'm proposing to add tutorials for it starting with inference.

Should I be making this pr against the `docs submodule` ?
Any thoughts on what I have so far would also be apreciated.

### DIFF
--- a/docs_src/sidebar.json
+++ b/docs_src/sidebar.json
@@ -37,6 +37,11 @@
                 "Lightning": "/migrating_lightning.html",
                 "Catalyst": "/migrating_catalyst.html"
             }
+        },
+        "empty4": {
+            "Production": {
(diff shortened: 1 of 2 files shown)


muellerzr:
I love the idea of this! Great initiative! I'm thinking we make a subsection in the docs under tutorials for inference. I'll give your notebook a thorough review with some feedback later today, but it looks really good! Just a few touch ups I think and it'll be set.

jdupl123:
Thanks @muellerzr I appreciate the feedback.

As the pr stands it adds a production tab with inference as a subsection.

image

muellerzr:
I'd recommend something like this:

- Inference
  - Tabular

Given your tutorial is tabular specific

Issue #3452, opened 2021-08-11
Incorrect behavior of crop_pad for images when image needs to be both cropped and padded


johan12345:
Please confirm you have the latest versions of fastai, fastcore, and nbdev prior to reporting a bug (delete one): YES

**Describe the bug**
The implementation of `crop_pad` for images seems to be seriously flawed for cases in which the image needs to be both cropped and padded. In this case, it does not crop the image to the given size.

**To Reproduce**
```python
>>> from fastai.vision.all import *
>>> img = PILImageBW.create(np.random.random((100, 100)))
>>> img.crop_pad((20, 20), (-10, 10))
<PIL.Image.Image image mode=F size=30x20 at 0x7FF4E7045AC0>
```
The resulting image has a size of 30x20.

```python
>>> img.crop_pad((20, 20), (90, 10))
<PIL.Image.Image image mode=F size=10x20 at 0x7FF5FF17F8E0>
```

The resulting image has a size of 10x20.

**Expected behavior**
The first example should crop the image to a size of (20, 20) at the position (-10, 10), which means adding 10 pixels of padding on the left side and then taking 10x20 pixels from the image, resulting in a 20x20 image.

**Additional context**
If `crop_pad` is used with the `resize_to` option, it is not immediately apparent that something goes wrong, as the image is output with the correct size given under  `resize_to`. But it will be squished in order to fit that size. That also makes it inconsistent with the `crop_pad` implementations for e.g. `TensorPoint`s, which are of course not squished in this way.

johan12345:
Found an additional edge case which is definitely wrong:

```python
>>> from fastai.vision.all import *
>>> img = PILImageBW.create(np.random.random((100, 100)))
>>> img.crop_pad((20, 20), (-1, -1))
<PIL.Image.Image image mode=F size=101x101 at 0x7F785AB67130>
```

PR #3453, opened 2021-08-11
Fix incorrect behavior of crop_pad

johan12345:
This is my attempt to fix the bug explained in #3452. I added new tests to verify the output size and content for additional test cases.

I'm new to the fastai/nbdev development workflow, so I hope I did it correctly :)


jph00:
Many thanks!

PR #3454, opened 2021-08-15
Fix MixHandler to restore original loss function of the learner

roboserg:
If MixUp (derived from MixHandler) is assigned as a callback during learner instantiation, running lr_find() will cause MixUp/MixHandler to change the loss function of the learner without restoring it back. This will lead to an error during training (as discussed in the #fastbook-and-courses Discord channel).

This PR solves this issue by restoring the original loss function of the learner in the after_cancel_train event, which will be invoked by lr_find().


fastai/callback/mixup.py
```diff
@@ -25,6 +25,9 @@ def before_train(self):
     def after_train(self):
         if self.stack_y: self.learn.loss_func = self.old_lf
 
+    def after_cancel_train(self):
+        if self.stack_y: self.learn.loss_func = self.old_lf
```
jph00, line 29:
```suggestion
        self.after_train()
```

jph00 (changes requested):
One little suggestion

fastai/callback/mixup.py
roboserg: yep, much cleaner, committed below.

jph00:
Many thanks!

PR #3456, opened 2021-08-16
Add explicit links to the migrating scripts [Documentation]

muellerzr:
Makes sure that each Migrating* tutorial has an explicit link to the various script they utilize

### DIFF
--- a/nbs/examples/migrating_catalyst.ipynb
+++ b/nbs/examples/migrating_catalyst.ipynb
@@ -40,7 +40,7 @@
    "source": [
     "We're going to use the MNIST training code from Catalyst's README (as at August 2020), converted to a module.\n",
     "\n",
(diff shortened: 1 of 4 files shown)


jph00:
The links are only needed for folks using Colab. So I'd suggest saying something like they're "in the `examples` subdirectory of this folder, if you checked out the repo from git, or can be downloaded from [here](...) if you're using an online viewer such as Colab."

muellerzr:
Makes sense, will make that adjustment today

jph00:
> Makes sense, will make that adjustment today

Any news?

muellerzr:
Yup! Sorry, life got busy unexpectedly. Will push some changes in an hour
or so.

muellerzr:
@jph00 made adjustments. Went with:

> Note: The source script for `X` is in the `examples` subdirectory of this folder if you checked out the `fastai` repo from git, or can be downloaded from [here](https://github.com/fastai/fastai/blob/master/nbs/examples/X.py) if you're using an online viewer such as Colab.

PR #3457, opened 2021-08-16
Add verbose migrating pytorch tutorial [Documentation]

muellerzr:
Adds a new tutorial to the migration guides that mimics the original Pytorch tutorial, but is a bit more explicit in what is going on.

There is benefit in having both so I've added this as a (Verbose) example. 

Eg: this one is geared towards beginners who want hand-holding in what each part does and interact with the framework, while the other is a nice quick walkthrough and script for them to mimic


jph00:
Thanks!

PR #3458, opened 2021-08-16
Fix lr_find to not error out with num_it < 6

roboserg:
Using `leaner.lr_find` with **num_it** < 6 will error out with `IndexError: index 0 is out of bounds for dimension 0 with size 0`.
**lrs** and **losses** variables will be empty due to [this line](https://github.com/fastai/fastai/blob/0bcde97e60ae28c9643112a95d1635147a80f703/fastai/callback/schedule.py#L285)
Those variables will be accessed [here](https://github.com/fastai/fastai/blob/0bcde97e60ae28c9643112a95d1635147a80f703/fastai/callback/schedule.py#L221), throwing out of bounds error, since those variables are empty. 

See [this notebook](https://colab.research.google.com/drive/1grgn6ClgTDNzZ-pShmsv2CYCTHsHboaD?usp=sharing) for examples.

It is worth considering hard capping **num_it** to not be below 50-100, as lower numbers are of questionable use, as seen in the google colab notebook.


muellerzr:
@jph00 not 100% sure why the CI is failing here, as running the notebook myself in Colab worked top -> down (even with CUDA).

roboserg:
Marking it for review since, as Zachary mentioned, notebooks diffs are "\r" only and the notebook runs without errors on my local machine as well.

jph00:
Many thanks. I think this isn't quite the right fix, so I've just pushed a fix myself.

PR #3465, opened 2021-08-20
Add comment about import Learner

ayasyrev:
nbs/examples/migrating_pytorch_verbose.ipynb 
Add comment about import Learner


jph00:
Thanks!

PR #3472, opened 2021-08-30
fix documentation for Datasets

k0ala:
the reference to `item_tfms` here is confusing since it does not pertain to the `Datasets` class.


jph00:
Thanks

Issue #3476, opened 2021-09-03
Learning Rate Suggestions are inaccurate with NaN [BUG]


muellerzr:
Currently if we run the LRFinder and there are NaN's present in the results (such as from gradient blowup) it will give us extremely unreasonable suggestions such as the example below:

![image](https://user-images.githubusercontent.com/7831895/131935022-6d622171-d8ff-4d77-abb9-b1dafa7a3c80.png)

The solution is to filter out the NaN losses and run the algorithms based on what is present there. Below is a much better looking suggestion threshold between all four current algorithms after this fix:

![image](https://user-images.githubusercontent.com/7831895/131935012-8c84f126-9ba3-4c21-83b3-c811d3b2f8a2.png)

PR #3477, opened 2021-09-03
Fix lr suggestion bugs

muellerzr:
Fixes bug brought up in this issue: https://github.com/fastai/fastai/issues/3476

The root of the cause is the fact that there can be NaN's present in the Losses if the gradients blew up too quickly. The solution is to simply filter out all of these nan's **before** calculating suggested learning rates.

Also valley wasn't affected by this because it inheritably can handle these edge cases as the algorithm finishes before ever reaching that point.

Below is a before/after picture with applying this filtering

image

image


muellerzr:
cc @jph00

jph00:
Good job!

PR #3481, opened 2021-09-15
Update 05_data.transforms.ipynb

maze1980:
While _test_splitter(f) works in this example,  _test_splitter(f, items=src) is better, because it doesn't break the example when changing the number of list items and labels.


jph00:
Many thanks

PR #3489, opened 2021-09-29
Remove redundant color specification

renato145:
Remove redundant color specification on `plot_lr_find`, it was causing warning messages from matplotlib.


jph00:
Thanks!

PR #3493, opened 2021-10-05
Fix #3277

### LINKED ISSUE #3277: pd.DataFrame.from_dicoms doesn't handle empty Sequences

This bug is present in latest fastai 2.5.3 fastcore 1.3.26.

**Describe the bug**
When processing a DICOM with empty `Sequence` tags, an exception is thrown inside `pd.DataFrame.from_dicoms`.

**To Reproduce**
```python
from fastai.basics import *
from fastai.medical.imaging import *
from pydicom.data import get_testdata_file
from pydicom.sequence import Sequence


ds = Path(get_testdata_file("CT_small.dcm")).dcmread()
ds.ReferencedImageSequence = Sequence([])
# Fails
ds.as_dict()
```

**Expected behavior**
No exception should be thrown.

```
/tmp/ipykernel_115321/2697590605.py in <module>
      9 ds.ReferencedImageSequence = Sequence([])
     10 # Fails
---> 11 ds.as_dict()

~/Documents/fastai/fastai/medical/imaging.py in as_dict(self, px_summ, window)
    378         for f in stats: res['img_'+f] = 0
    379         print(res,e)
--> 380     for k in res: res[k] = _cast_dicom_special(res[k])
    381     return res
    382 

~/Documents/fastai/fastai/medical/imaging.py in _cast_dicom_special(x)
    352     if not cls.__module__.startswith('pydicom'): return x
    353     if cls.__base__ == object: return x
--> 354     return cls.__base__(x)
    355 
    356 def _split_elem(res,k,v):

TypeError: __init__() missing 1 required positional argument: 'iterable'
```

thomsonn:
`_split_elem` is supposed to strip all MultiValues, but fails for sequences of length 0 as the value is never overwritten by the for loop.

### DIFF
--- a/fastai/medical/imaging.py
+++ b/fastai/medical/imaging.py
@@ -353,10 +353,16 @@ def _cast_dicom_special(x):
     if cls.__base__ == object: return x
     return cls.__base__(x)
 
-def _split_elem(res,k,v):
-    if not isinstance(v,DcmMultiValue): return
-    res[f'Multi{k}'] = 1
-    for i,o in enumerate(v): res[f'{k}{"" if i==0 else i}']=o
+def _split_elem(vals):
+    res = dict()
+    for val in vals:
+        k, v = val.keyword, val.value
+        if not isinstance(v,DcmMultiValue):
+            res[k] = v
+            continue
+        res[f'Multi{k}'] = 1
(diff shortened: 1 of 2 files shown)


jph00:
Many thanks!

PR #3496, opened 2021-10-08
Missing shapes for activations and Flatten in learner.summary

marcinczeczko:
When I was going from fastbook `13_convolutions.ipynb` notebook I found that the `learn.summary()` method was printing out the shapes for activation functions. The one that was really useful for myself was the output shape of the `Flatten()` module. 
image

However, going through the notebook with the latest fastai, i found out that output shape of the Flatten module is empty, what I found misleading. I found that the format of the summary is changed which is not a problem, but I found that missing shapes of non-trainable modules (like flatten) might be misleading.

Here is my attempt to fix that situation. Apparently, shapes wasn't calculated for non-trainable modules.


jph00:
Many thanks.

PR #3497, opened 2021-10-08
minor corrections in tutorial documentation

vivekh2000:
**Reference Issues/PR**
None

**What does this implement/fix?**
These are just minor corrections in the tutorial documentation and fixes [(Data](https://docs.fast.ai/tutorial.datablock.html) block tutorial and [Tutorial - Training a model on Imagenette](https://docs.fast.ai/tutorial.imagenette.html)) that I noticed while reading through the documentation.

**Any other comments?**
While going through the tutorials in fastai docs there seem to be discrepancies in the text description and code outputs. For example, the following screenshot is from the datablock tutorial. 
image

the text reads, “_how it changed the name “cat” to an index_.” However, if you see the output above, it says `TensorCategory(1)`, which means as per the sorted vocab should be the `dog` category.
 
Also, from the [imagenette tutorial](https://docs.fast.ai/tutorial.imagenette.html), the screenshot is shown below.
image

The text reads that “how it changed the name “cat” to an index,” but there is no “`cat`” category in the imagenette dataset. I believe it should be “`cassette player`” as per the sorted vocab and output `TensorCategory(2)`


jph00:
Many thanks.

PR #3501, opened 2021-10-15
Minor typo fixes in tabular tutorial

xairos:
Thanks for the great fastai docs!

I just spotted these two typos as I was reading the tabular tutorial.


jph00:
Many thanks

PR #3511, opened 2021-10-27
Support both torch.solve & torch.linalg.solve

### LINKED ISSUE #3469: torch.solve is deprecated in favour of torch.linalg.solve
Hi!
I just got the following warning when using PyTorch version 1.9.0:

```
/home/jdg/.local/lib/python3.9/site-packages/torch/_tensor.py:1023: UserWarning: torch.solve is deprecated in favor of torch.linalg.solveand will be removed in a future PyTorch release.
torch.linalg.solve has its arguments reversed and does not return the LU factorization.
To get the LU factorization see torch.lu, which can be used with torch.lu_solve or torch.lu_unpack.
X = torch.solve(B, A).solution
should be replaced with
X = torch.linalg.solve(A, B) (Triggered internally at  ../aten/src/ATen/native/BatchLinearAlgebra.cpp:752.)
  ret = func(*args, **kwargs)
```
The source of this deprecation warning seems to be here:

https://github.com/fastai/fastai/blob/52d6302eb4487e86382e663ef5c10ee950c07ad1/fastai/vision/augment.py#L611

I presume that a suitable fix would be to do something like they say.

Unfortunately, though, `torch.linalg.solve` only seems to have been introduced in version 1.9.0 (commit https://github.com/pytorch/pytorch/commit/8be205ae1364ff3c48b41200282bafd4805cb63c on Mon Dec 21 10:09:28 2020 -0800), which means that changing this code will break fastai if PyTorch <= 1.8 is used but leaving it will break fastai if PyTorch >= 1.10 (once it is released) is used. Having the function be conditional on which version of PyTorch is being used would probably be even uglier, but is perhaps a reasonable resolution.

warner-benjamin:
Resolves #3469. Tested on both 1.9.0+cu111 and 1.8.2+cu102.

Also removed the PyTorch 1.4 check in `_grid_sample` since minimum version is 1.7.

### DIFF
--- a/fastai/vision/augment.py
+++ b/fastai/vision/augment.py
@@ -298,11 +298,7 @@ def _grid_sample(x, coords, mode='bilinear', padding_mode='reflection', align_co
         d = min(x.shape[-2]/coords.shape[-2], x.shape[-1]/coords.shape[-1])/2
         # If we're resizing up by >200%, and we're zooming less than that, interpolate first
         if d>1 and d>z:
-            # Pytorch > v1.4.x needs an extra argument when calling nn.functional.interpolate to preserve previous behaviour
-            if (int(torch.__version__[0:4].replace(".", "")) > 14):
-                x = F.interpolate(x, scale_factor=1/d, mode='area', recompute_scale_factor=True)
-            else:
-                x = F.interpolate(x, scale_factor=1/d, mode='area')
+            x = F.interpolate(x, scale_factor=1/d, mode='area', recompute_scale_factor=True)
     return F.grid_sample(x, coords, mode=mode, padding_mode=padding_mode, align_corners=align_corners)
 
 # Cell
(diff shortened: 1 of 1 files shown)


fastai/vision/augment.py
```diff
@@ -596,6 +592,16 @@ def __init__(self,min_zoom=1., max_zoom=1.1, p=0.5, draw=None, draw_x=None, draw
...
         super().__init__(aff_fs, size=size, mode=mode, pad_mode=pad_mode, align_corners=align_corners)
 
+# Internal Cell
+def _linalg_solve(A,B):
+    return torch.linalg.solve(A,B)
+
+def _solve(A,B):
+    return torch.solve(B,A)[0]
+
+if torch.__version__ >= '1.9': solve = _linalg_solve
```
Rebolforces, line 602: @jph00   won't this string compare for version fail? ie '1.10.0' >= '1.9' is False

Rebolforces (commented):
Won't string compare for version will fail. e.g '1.10' >= '1.9' == False

jph00:
Yes good point. @Rebolforces @warner-benjamin perhaps one of you could fix it with a PR that does something like shown here:
https://stackoverflow.com/questions/11887762/how-do-i-compare-version-numbers-in-python/21065570

warner-benjamin:
Yeah. I can submit a PR for that. There's another incorrect version string comparison in [layers](https://github.com/fastai/fastai/blob/f8b74ef5b320512a2bb4a6c3cb17a5e917b7d6a3/fastai/layers.py#L598) which I copied for this PR.

Would it be preferred to add a get_torch_version method which returns a packaging.version.Version object in torch_core, or do the entire version check where needed?

There's also a PyTorch 0.4 check in [fp16_utils](https://github.com/fastai/fastai/blob/ab154927696338741e59e0ffc4774777c4a9781c/fastai/fp16_utils.py#L183) and a PyTorch 1.4 check in [examples/train_imdbclassifier.py](https://github.com/fastai/fastai/blob/ab154927696338741e59e0ffc4774777c4a9781c/nbs/examples/train_imdbclassifier.py#L36) which both can be removed.

jph00:
I think a function is a great idea

warner-benjamin:
Both PRs fastai/fastcore#378 and #3518 resolve this issue. Thanks for catching it @Rebolforces.

PR #3512, opened 2021-10-27
Updated wandb callback to used validation data loader to inspect input

ohmeow:
Still not sure why, but after unfreezing the model, calling  causing computation to be added to backprop in some cases (e.g., training transformer models with Blurr). Changing this to instead use the validation data loader fixes the issue; but I'm still not sure why.

### DIFF
--- a/nbs/70_callback.wandb.ipynb
+++ b/nbs/70_callback.wandb.ipynb
@@ -2,7 +2,7 @@
  "cells": [
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": null,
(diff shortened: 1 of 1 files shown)


ohmeow:
Well, finally figured this out with some help from Zach ...

Having a call to `dls.train.one_batch` in a callback will undo setting all the random seeds right before `fit` because the dataset is shuffled on each iteration, thereby changing the order of items when training starts.

Seems like a best practice for callback authors is to use `dls.valid.one_batch` should they need a batch to work with.

PR #3516, opened 2021-10-29
Fix nested tensors predictions compatibility with fp16

tcapelle:
Modifies retrieving first tensor to check float16 dtype to support nested tensors.
> Warning: This PR needs the flatten function that is added in this other [PR](https://github.com/fastai/fastcore/pull/375).

### DIFF
--- a/fastai/callback/fp16.py
+++ b/fastai/callback/fp16.py
@@ -19,7 +19,7 @@ class MixedPrecision(Callback):
     def before_fit(self): self.learn.scaler,self.scales = GradScaler(**self.kwargs),L()
     def before_batch(self): self.autocast.__enter__()
     def after_pred(self):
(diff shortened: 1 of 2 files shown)


jph00:
@tcapelle can you please try changing `listify(flatten(...))` to `concat(...)`, since that's the same thing now

tcapelle:
we have a conflict with `torch_core.concat`

jph00:
OK happy to leave it as-is then. Well spotted on the `concat` name clash - that's actually been there for ages but no-one had noticed before!

jph00:
Just another thought: what if instead of `listify(flatten(self.pred))[0]` we did `next(flatten(self.pred))` ? That would avoid needlessly allocating extra memory.

jph00:
Looks like it's all sorted! I'll update the fastcore version dep too.

PR #3518, opened 2021-11-02
Fix torch version checks, remove clip_grad_norm check

warner-benjamin:
Resolves issues with torch version checks brought up in #3511. Requires fastai/fastcore#378.

`clip_grad_norm` in fp16_utils.py is unused, so removed it and the version check.


warner-benjamin:
@jph00, I updated this PR per your comments on fastai/fastcore#378.

jph00:
Many thanks!

PR #3520, opened 2021-11-05
Fix _get_shapes to work with dictionaries

ohmeow:
Latest improvement to Learner.summary broke blurr (and maybe other hugging face transformer models where dictionaries are used to pass the inputs).  This fixes that.


jph00:
Many thanks. Do you want to also add a test for this so it doesn't break again in the future?

ohmeow:
Will do. Do you mind if I add that in a later PR?

-wg

jph00:
Sure thing

PR #3540, opened 2021-12-18
Added docs for WeightedDL

kwsp:
WeightedDL's `wgts` parameter takes an array of weights of the same length as the training data (excluding validation data). This was previously undocumented and caused a lot of confusion for me.


jph00:
@kwsp sorry it has taken me a while to get to this.

Could you please add the `ImageBlock` import to a non-exported cell? Otherwise it become part of the exported lib, resulting in circular references.

Also, would you be able to add a bit of Markdown prose explaining the code you added?

kwsp:
@jph00 fixed both

jph00:
@kwsp looks like you didn't merge in the changes of today's merged PRs so they're appearing as commits here too.

kwsp:
@jph00 I tried to rebase the new commits earlier and I guess GH created a merge commit. I've removed those.

jph00:
Many thanks!

PR #3547, opened 2021-12-31
Update 10_tutorial.pets.ipynb

jss367:
My goal for this PR was to fix the `IndexError: list index out of range` near the end of the [pets tutorial](https://docs.fast.ai/tutorial.pets.html). To do so I had to rerun the notebook, which changed all the images. This makes the PR appear to change much more than I had intended. I would like to try to fix the problem so if there's a cleaner way to do this, please let me know and I'll be happy to do so.


jph00:
Thanks for the fix!

PR #3554, opened 2022-01-15
Fix: Learner.show_results uses passed dataloader via dl_idx or dl arguments

warner-benjamin:
Changes `Learner.show_results` from being hardcoded to use the training dataloader's `show_results` to use the dataloader passed to `Learner.show_results` via `dl_idx` or `dl` arguments.

Resolves an issue when the train and valid dataloaders differ, which could result in the incorrect decodes being applied to valid inputs. For example, when training on image sub-patches and during validation combining patches to form a whole image.


jph00:
Many thanks!

PR #3558, opened 2022-01-21
Updates Interpretation class to be memory efficient

warner-benjamin:
`Interpretation` is now memory efficient and should be able to process any size dataset, provided the hardware could train the same model.

`Interpretation` now calls `get_preds` for each item to generate inps, targs, etc.

`GatherPredsCallback` has two new arguments, `with_preds` and `with_targs` to control whether it returns predictions or targets. These are used by `Interpretation` for memory efficient creation.

Also modifies `GatherPredsCallback` to use `torch.save`, which handles both tensors and tuples of tensors.


jph00:
Great stuff!

PR #3563, opened 2022-01-25
Fix Classification Interpretation

### LINKED ISSUE #3562: TypeError: __init__() missing 2 required positional arguments: 'decoded' and 'losses'

**Describe the bug**
When I try running `ClassificationInterpretation.from_learner(test_learner)` I get an error as of #3558.  This error appears to be due to the init function of Interpretation changing but the init function of Classification staying the same.  

**To Reproduce**
1. go to notebook 20_interpret
2. Create a new cell after ClassificationInterpretation is defined
3. `classification_interp = ClassificationInterpretation.from_learner(test_learner)`
4. Using the test learner created higher up in the notebook, I believe this should work with ClassificationInterpretation without issues

**Expected behavior**
ClassificationInterpretation is usable with no error

```
/tmp/ipykernel_22/2455746672.py in <module>
----> 1 ClassificationInterpretation.from_learner(test_learner)

/tmp/ipykernel_22/1144871617.py in from_learner(cls, learn, ds_idx, dl, act)
     21         _,_,losses = learn.get_preds(dl=dl, with_input=False, with_loss=True, with_decoded=False,
     22                                      with_preds=False, with_targs=False, act=act)
---> 23         return cls(learn, dl, losses, act)
     24 
     25     def top_losses(self, k=None, largest=True, items=False):

TypeError: __init__() missing 2 required positional arguments: 'decoded' and 'losses'

```

warner-benjamin:
Resolves #3562 and adds simple test to make sure `ClassificationInterpretation` doesn't fall through the cracks again.

Implementation does generate the entire datasets' worth of predictions, targets, and decoded predictions, which is required for the confusion matrix. It might be possible to make this more memory efficient, but the labels and predictions for classification problems are not usually that large.

### DIFF
--- a/nbs/20_interpret.ipynb
+++ b/nbs/20_interpret.ipynb
@@ -8,7 +8,7 @@
    "source": [
     "#hide\n",
     "#skip\n",
-    "! [ -e /content ] && pip install -Uqq fastai  # upgrade fastai on colab"
+    "# ! [ -e /content ] && pip install -Uqq fastai  # upgrade fastai on colab"
    ]
   },
   {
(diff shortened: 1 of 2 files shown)


nbs/20_interpret.ipynb
```diff
@@ -8,7 +8,7 @@
    "source": [
     "#hide\n",
     "#skip\n",
-    "! [ -e /content ] && pip install -Uqq fastai  # upgrade fastai on colab"
+    "# ! [ -e /content ] && pip install -Uqq fastai  # upgrade fastai on colab"
```
jph00, line 11: Could you please uncomment this? It's needed to work on colab

jph00 (changes requested):
Many thanks - just one request

warner-benjamin:
Done

jph00:
Great.

rsomani95:
@warner-benjamin With the changes made here, one would be re-running inference on each call of plotting the confusion matrix or fetching the `skm` classification report, right? If so, that could get very cumbersome esp. with larger test sets, and if you wanted to use the stored predictions for anything else.

warner-benjamin:
@rsomani95 It's on my todo list to add back a cached option

marcocuccarini:
Hi, i see that the guy that you answer had my same problem but I did't understand the answer. For my example how I can fix the error? I'm sono sorry for the silly question.


`for benchmark in benchmarks:

          training_id, learner = self.create_learner(train_index, valid_index, training_id, 'test' == benchmark)

          print('*'*30)
          print('%s metrics for %s' % (benchmark, training_id))

          metrics = learner.validate()
          
          for (metric, value) in zip(metricsL, metrics):
            results[benchmark][metric].append(value)
            

            interp = DSClassificationInterpretation.from_learner(learner)

          confusion_matrices[benchmark].append(interp.confusion_matrix())
          classification_reports[benchmark].append(interp.classification_report())
`

PR #3568, opened 2022-01-28
Update import for show_install

fr1ll:
show_install was moved at some point to test_utils from utils.collect_env.
This updates the contributor guide so people will know how to
print version information.


jph00:
Thank you!

Issue #3561, opened 2022-01-22
docker-compose changes file permissions


kevinbird15:
https://github.com/fastai/fastai/blob/13b4734af41a3ac24d7ef3819f4103093174d204/docker-compose.yml#L36

I noticed that the docker-compose.yml file changes the permissions of the files in docs and I wanted to check if that is the intended functionality and if so, why is this required?  The reason this came up for me is because I ran docker-compose up and I noticed a bunch of file permissions had changed and this is what I traced it back to.  I tried removing this snippet and the documentation still seems to run fine, but I'm wondering if I'm missing another case where it was broken before.

kevinbird15:
My concern that I had with this PR is that it would somehow hurt the automatic docs deployment so I was watching close for the next PR after this was merged.  #3573 was the next PR and it adds a `name` option to AccumMetric.  I just checked the docs and it shows up properly so I am pretty comfortable now saying this change is valid and could probably be replicated in the rest of the docker-compose.yml files.

![image](https://user-images.githubusercontent.com/7451178/153334481-c27988c2-7c37-426d-916b-c570d4b22c89.png)

The only other concern that I have is with this warning that I first noticed because I was looking very close at CI/CD pipeline after #3571.  `WARNING: Error loading config file: /root/.docker/config.json: open /root/.docker/config.json: permission denied`

But when I looked closer, I see this has been happening for quite some time so it appears to be unrelated and also doesn't seem to matter?  I will investigate separately, but closing this as resolved for the fastai repo

PR #3571, opened 2022-02-06
removing permissions change

kevinbird15:
Checking to see if it is possible to remove the `chmod -R u+rwx .` from the docker-compose.yml file.  This would be beneficial when using the docker-compose file locally to view docs or for development because currently using this will cause many permission changes and git thinks this is a difference.  If we are able to remove it, this is a pretty clean way to spin up a development environment locally and view the docs as changes are made.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -33,4 +33,4 @@ services:
      && pip install .
      && nbdev_build_docs && cd docs
      && bundle i
-     && chmod -R u+rwx . && bundle exec jekyll serve --host 0.0.0.0"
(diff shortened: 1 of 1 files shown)


kevinbird15:
closes #3561

kevinbird15:
I haven't seen any issues with this change locally and it seems to be fine in the ci pipeline.  I think if it is going to fail, it would be on the docs action which I don't think runs until merging to master.  I tried running that test on my end but wasn't able to get it running.

jph00:
OK I'll merge this now - but please keep an eye on CI for future commits to see if there's an issue

PR #3573, opened 2022-02-08
Custom name for metrics

bdsaglam:
This PR adds a setter to the name property of AccumMetric so that we can give a custom name for a metric. This is useful when we want to use the same metric type with different arguments, such as threshold or average arguments. For instance, 
in multi-label classification, when we want to have F1 scores with `macro` and `samples` averaging, this is what we currently have:
metrics-without-custom-name
In epoch results, both F1 scores are named as `f1_scores`.

With this PR, we'll able to distinguish them:
metrics-with-custom-name


jph00:
That's a good idea! Thanks for the PR

PR #3575, opened 2022-02-10
Update 03_data.core.ipynb

jss367:
I often find myself wanting to know have many `loaders` are in `DataLoaders` but there is no `__len__` method currently. This seems counterintuitive to me because they have a `__getitem__` method so I can access individual loaders with, for example, `my_dataloader[0]`. I didn't see other comments about this in the forum but I thought I would make the suggestion.

### DIFF
--- a/fastai/data/core.py
+++ b/fastai/data/core.py
@@ -142,6 +142,7 @@ class DataLoaders(GetAttr):
         if device is not None or hasattr(loaders[0],'to'): self.device = device
 
     def __getitem__(self, i): return self.loaders[i]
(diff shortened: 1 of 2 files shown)


jph00:
That makes sense! Can you please `nbdev_build_lib` as discussed here: https://docs.fast.ai/dev-setup.html .

jss367:
Great! I ran `nbdev_build_lib` and everything seems to be working (although I'll check the build again after it's finished). Let me know if there's anything else I need to do.

jph00:
Great!

PR #3577, opened 2022-02-11
add LIB_INSTALL_TYPE env file to docker-compose 

kevinbird15:
Wanted to throw this idea out based on some conversations in discord.  This would allow users to fairly quickly switch from the normal `pip install -e .` and `pip install -e .[dev]` by changing a new environment variable (LIB_INSTALL_TYPE) that I am proposing is set in docker-compose.yml.


jph00:
lgtm!

PR #3581, opened 2022-02-16
Fix bug in 04_data.external

malligaraj:
In `URLs.path()`, there is a check to see if the file to download exists in folders `data` or `models` in `URLs.LOCAL_PATH`.

The key to the `Config` is set to `models` instead of `model` when forming the local path. The key should be one of ['archive', 'data', 'model', 'storage'] according to  [URLs docs](https://docs.fast.ai/data.external.html#URLs)

I have tried the following:

1. Current working director is empty (path pointed by LOCAL_PATH = Path.cwd())
2. Run URLs.path() with different values for c_key ['data', 'archive', 'storage', 'model']

```python
for k in ['data', 'archive', 'storage', 'model']:
    print(URLs.path(URLs.MNIST_TINY,c_key=k))

"""
Output:

/root/.fastai/data/mnist_tiny.tgz
/root/.fastai/archive/mnist_tiny.tgz
/root/.fastai/tmp/mnist_tiny.tgz
/root/.fastai/models/mnist_tiny.tgz
"""
```

Then, 

1. Copied `mnist_tiny.tgz` to my current directory under folders `data` and `models`
2. Ran the same code as above
```python
for k in ['data', 'archive', 'storage', 'model']:
    print(URLs.path(URLs.MNIST_TINY,c_key=k))

"""
Output:

/content/data/mnist_tiny.tgz
/content/data/mnist_tiny.tgz
/content/data/mnist_tiny.tgz
/content/data/mnist_tiny.tgz  # <---
"""
```
I was expecting the last output line in the second run to be `/content/models/mnist_tiny.tgz` (marked by `<---`).


malligaraj:
Sorry, messed up the branching. Will create another PR with correct commit

PR #3590, opened 2022-02-24
deleted unused add_vert arg in 08_vision.data:get_grid

ChrisRichardMiles:
One thing to note as I look at the changes. There is this change in the notebook which I don't understand: 
```
   "display_name": "Python 3",
   "display_name": "Python 3 (ipykernel)",
```
I don't know if it matters, or if I can do anything about it.


jph00:
Many thanks. The change to `display_name` can be ignored.

PR #3602, opened 2022-03-03
Fixing ColReader

hiromis:
I came across a similar issue as https://github.com/fastai/fastai/issues/3164 with a column named `'mask'`. This time, it was a conflict with DataFrame's function called `mask` (https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.mask.html). 

Here is an example I took from [a tutorial](https://github.com/fastai/fastai/blob/master/nbs/examples/app_examples.ipynb) that uses ColReader to reproduce the issue:
https://gist.github.com/hiromis/e371603e71543cd7cca9cea7dfcc7609

I could have added 'mask' to a list of special cases along with 'name' and 'cat', but I thought checking whether `r` is a namedtuple containing a field `c` may be cleaner.

The existing tests ran fine, and I also did a sanity check using the example notebook:
https://gist.github.com/hiromis/7a0863072694faa5a0f6fb0bbcb3d2da


jph00:
Many thanks!

PR #3604, opened 2022-03-03
Add Python 3.9 to `setup.py` for PyPI

nzw0301:
https://pypi.org/project/fastai says fastai supports Python 3.6 -- 3.8, but I suppose fastai works with python 3.9. I've installed fastai with python 3.9.

Screenshot 2022-03-03 at 22 30 17


jph00:
Many thanks.

PR #3606, opened 2022-03-14
Fix pin_memory=True breaking (batch) Transforms

### LINKED ISSUE #2795: pin_memory=True breaks (batch) Transforms
```
from fastai.data.all import *
path = untar_data(URLs.PETS)
fnames = get_image_files(path/"images")
pets = DataBlock(blocks=(ImageBlock, CategoryBlock), 
                 get_items=get_image_files, 
                 splitter=RandomSplitter(),
                 get_y=Pipeline([attrgetter("name"), RegexLabeller(pat = r'^(.*)_\d+.jpg$')]),
                 item_tfms=Resize(128),
                 batch_tfms=aug_transforms())
dls = pets.dataloaders(untar_data(URLs.PETS)/"images",pin_memory=True)
dls.show_batch(max_n=9)
```
breaks: `pin_memory=True` makes DL return a list instead of a tuple and `Transforms` (e.g. `batch_tfms`) expect batches to be tuples.

Why is this important? `pin_memory=True` is 1st performance recommendation here https://twitter.com/karpathy/status/1299921324333170689

johan12345:
A simple workaround that fixes #2795.

The reason for the bug is that PyTorch's `pin_memory` ([v1.10.0](https://github.com/pytorch/pytorch/blob/v1.10.0/torch/utils/data/_utils/pin_memory.py#L57-L58), [v1.11.0](https://github.com/pytorch/pytorch/blob/v1.11.0/torch/utils/data/_utils/pin_memory.py#L61-L62)) converts tuples into lists - so we need to convert back to a tuple.


johan12345:
Sorry for the late reply! I think 08c2abf should fix the test.

jph00:
Many thanks!

PR #3607, opened 2022-03-18
Draft: Add Activity State to Learner

warner-benjamin:
# Problem
There is an issue where `Callback.before_fit` and `Callback.after_fit` are called in multiple contexts, during training, validating, and prediction, without any information available to the callback regarding the context is calling it.

For inner loop events,
```python
_inner_loop = "before_batch after_pred after_loss before_backward before_step after_step after_cancel_batch after_batch"
```
there is the training flag, but it's set by `TrainEvalCallback` after `before_fit` is called and ending before `after_fit` is called. And it cannot distinguish between validating during `Learner.fit` or validating during `Learner.get_preds`.

# Solution
This PR resolves this issue by adding an activity state `Learner.state` populated from a new `LearnerState` enum. It currently has four states: Fit, Eval, LRFind, and Idle. 

Fit is set during `Learner.fit`. Eval is set during `Learner.get_preds` and `Learner.validate`. LRFind is set when fitting the `LRFinder` callback.

The state is set via a context manager, so Callbacks could hypothetically override or set their own custom states which will automatically be set back to `LearnerState.Idle` when training/evaluation is finished.

# ToDo

- [ ] Update documentation
- [ ] Modify existing callbacks to use `Learner.state`
- [ ] Rework state pending feedback


warner-benjamin:
Closed due to fastai v2 no longer accepting major changes due to imminent development of fastai v3.

PR #3611, opened 2022-03-25
Add in exceptions and verbose errors

muellerzr:
Includes a variety of verbose errors for commonly situated problems. Due to the complexity of some issues, some of the logic in said errors is a bit much in order to provide a clear picture of what went wrong (such as the `collate_error`).

Open to removing any you're not comfortable with @jph00


muellerzr:
Aware there's a small bug with the tests in there I think. Will get to it in the next few days

jph00:
@muellerzr can you take a look at this CI failure? Might just be a case of needing to merge lastest fastai changes - not sure.

muellerzr:
@jph00 okay, that was an exploration and a half but we got callbacks working. One thing I did have to do though was move up where the exception classes were defined so they can be pulled into `Callback.__call__`. Since we already use `show_doc` for them all though, I think it would make sense to keep the documentation in the same spot

fastai/callback/core.py
```diff
@@ -42,7 +53,12 @@ def __call__(self, event_name):
         _run = (event_name not in _inner_loop or (self.run_train and getattr(self, 'training', True)) or
                (self.run_valid and not getattr(self, 'training', False)))
         res = None
-        if self.run and _run: res = getattr(self, event_name, noop)()
+        if self.run and _run:
+            try: res = getattr(self, event_name, noop)()
+            except (CancelBatchException, CancelEpochException, CancelFitException, CancelStepException, CancelTrainException, CancelValidException): raise
+            except Exception as e:
```
muellerzr, line 59: We could also rewrite this as something like:
```python
except Exception as e:
  if len(e.args) > 0: 
    e.args = [f'Exception occured in `{self.__class__.__name__}` when calling event `{event_name}`:\n\t{e.args[0]}']
  raise
```

jph00:
Thanks @muellerzr !

PR #3626, opened 2022-04-22
fix loading of some timm models by fixing `create_head`

tmabraham:
changed `if concat_pool: nf *= 2` to `if pool and concat_pool: nf *= 2`  in the `create_head` function so the number of input features isn't doubled if `pool` is False.

This fixes usage of `vision_learner` with timm models that didn't need pooling.


jph00:
Many thanks!

PR #3644, opened 2022-05-11
[Enhancement] New Callback event, before and after backward

muellerzr:
This PR introduces a new Callback event: `after_backward`. This is designed so custom backward pass implementations can be used (such as accelerate).

`after_backwards` is the same as `before_step`, but I opted to keep both because there isn't a clean way to combine the two into one without introducing too many changes. 

If a callback implements their own backward pass, they just need to attach a `RaiseBackwardException` in the implementation.

cc @jph00


muellerzr:
ci failed unrelated to the current notebook

nbs/13a_learner.ipynb
```diff
@@ -24,7 +24,20 @@
    "cell_type": "code",
    "execution_count": null,
    "metadata": {},
-   "outputs": [],
+   "outputs": [
+    {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "/opt/conda/lib/python3.9/site-packages/torch/cuda/__init__.py:145: UserWarning: \n",
```
jph00, line 32: Can you fix this? It'll mess up the docs

nbs/13a_learner.ipynb
```diff
@@ -723,7 +735,32 @@
    "cell_type": "code",
    "execution_count": null,
    "metadata": {},
-   "outputs": [],
+   "outputs": [
+    {
+     "ename": "RuntimeError",
```
jph00, line 740: ...and this

jph00 (changes requested):
Looks like you need to re-run this after fixing your cuda env

nbs/13a_learner.ipynb
muellerzr: Ack, yup sorry! Will do so momentarily

warner-benjamin:
Could you update the Callback documentation to mention `after_backward` and what it is used for? Currently it reads:
```
- `before_backward`: called after the loss has been computed, but only in training mode (i.e. when the backward pass will be used)
- `before_step`: called after the backward pass, but before the update of the parameters. It can be used to do any change to the gradients before said update (gradient clipping for instance).
```

muellerzr:
@warner-benjamin `after_backward` is synonymous with `before_step`. But everyone should use `before_step` really, hence why it was left unmentioned.

warner-benjamin:
That strikes me as an important thing to mention.

PR #3646, opened 2022-05-12
Integrate Accelerate into fastai

muellerzr:
This PR introduces Accelerate as the main handler for distributed training in fastai. 

A new tutorial was also included to show off how to use `notebook_launcher` to run distributed code from a notebook.

cc @jph00


muellerzr:
I wasn't sure what would be considered "too much" publicity/discussion on accelerate as a whole, let me know what you think about adding something like this  to the start of 20 distrib:

> When using multiple GPUs, you will most probably want to fit using distributed training. fastai utilizes [Accelerate](https://github.com/huggingface/accelerate) to make sure the process is as smooth as possible, and ensure that code works on not just multiple GPU's, but TPU's as well.  See [examples/distrib.py](https://github.com/fastai/fastai/blob/master/nbs/examples/distrib.py) for a complete example. To use distributed training, there are only two required steps:

muellerzr:
We're moving `setup_accelerate` into it's on util in accelerate as it's more useful there. Will update this PR once that change has been made

jph00:
Got it @muellerzr -- I've made this a draft PR in the meantime. Let me know when it's ready to merge. (Which should wait until there's a pypi release of Accelerate.)

muellerzr:
@jph00 new release is now out, updated the tutorial w/ the new directions 😄

PR #3654, opened 2022-05-18
Add built-in integration to Comet

rgaiacs:
Fastai has built-in integration with Neptune, see https://docs.fast.ai/callback.neptune.html.

I use [Comet.ml](https://www.comet.ml/) instead of Netpune and I want to have Comet available from Fastai similar to Neptune.

Is this contribution welcome?

PR #3660, opened 2022-05-24
Refactor integration notebook names and fix comet integration

muellerzr:
# Refactor integration notebook names, and fix recently added comet integration

## What does this add?

- Renames all integration notebooks to follow a related series of `70a_...`, since they are all related.
- Properly makes the comet_ml notebook export to a comet module, instead of overriding `neptune` like it currently does

## Why is it needed?

I was glancing at recent PR merges and the result of [this](https://github.com/fastai/fastai/pull/3654/) PR smelled fishy, as exports were used but no .py's were being added in the pull request. Upon investigating I found that `default_exp callback.neptune` was being used instead of `callback.comet`.

I can't quite explain why `nbdev_diff_nbs` didn't catch this however, which is quite scary!

cc @jph00


jph00:
Well spotted!

PR #3664, opened 2022-05-27
plot_top_losses: fix its being limited to 9 items

johan12345:
A change in #3558 removed the `max_n` argument from one of the `_pre_show_batch` calls within `Interpretation.plot_top_losses` (probably by mistake?). This caused the number of items in `its` to default to 9 even if a different number `k` of losses to plot was set. This may lead to only 9 items being plotted even if more were requested.


jph00:
Well spotted - thanks!

Issue #3666, opened 2022-05-29
Fix Tokenizer.get_lengths in fastai.text.core


karotchykau:
Please confirm you have the latest versions of fastai, fastcore, and nbdev prior to reporting a bug (delete one): YES.

**Describe the bug**
The `get_lengths` method of `Tokenizer` from `fastai.text.core` does not work properly. This method contains two bugs:
1. There is no such column as `text_lengths`. There is a mix of `text_lengths` and `text_length`. The latest one is correct.
2. When `Datasets` is created using `Tokenizer` from `fastai.text.core`, `get_lengths` will always return `None`. 

**To Reproduce**
Steps to reproduce the behavior of the second bug:
```
from operator import attrgetter

import pandas as pd
from fastai.data.core import Datasets
from fastai.text.core import Tokenizer

dataset = Datasets(pd.DataFrame({"a": ["abc", "123 456"]}), [[attrgetter("text"), Tokenizer.from_df("a")]])
tokenizer = dataset.tokenizer
print(dataset.items)
print(tokenizer.lengths)
print(tokenizer.get_lengths(dataset.items))
```
Output:
```
                text  text_length
0       [xxbos, abc]            2
1  [xxbos, 123, 456]            3
None
None
```

Please note that the first bug isn't described here because it's just a typo in the code.

**Expected behavior**
```
                text  text_length
0       [xxbos, abc]            2
1  [xxbos, 123, 456]            3
None
[2 3]
```

**Error with full stack trace**
No errors have occurred.

**Additional context**
No additional context to add.

PR #3667, opened 2022-05-29
Fix Tokenizer.get_lengths

karotchykau:
Fixes [#3666](https://github.com/fastai/fastai/issues/3666).


jph00:
Many thanks for the PR. Note that fastai is an nbdev project, therefore edits need to be made in the notebooks, not the .py files -- the .py files are auto-generated. A walk-thru is here, if needed: https://docs.fast.ai/dev-setup.html

Also, could you explain more about the reasoning for the first change you made, to reverse the check for `None`?

karotchykau:
> Many thanks for the PR. Note that fastai is an nbdev project, therefore edits need to be made in the notebooks, not the .py files -- the .py files are auto-generated. A walk-thru is here, if needed: https://docs.fast.ai/dev-setup.html

Got it. Only fixed it in `nbs/30_text.core.ipynb`.

> Also, could you explain more about the reasoning for the first change you made, to reverse the check for `None`?

I think we don't need to reverse but the current version is incorrect anyway. Here is what I mean. There are two modes:
* `df`, it will save `text_length` into some DataFrame, however, it doesn't save these lengths into `self.lengths`.
* `folder`, it doesn't have `text_length` at all (and can't have it), however, it saves these lengths into `self.lengths`.

If `if self.lengths is None: return None` comes first, `if self.mode == 'df'` will never be executed because `self.lengths` is always `None` for this mode, these lengths are stored as a column in this DataFrame. Please note that `if self.mode == 'folder'` will work fine.

I see two solutions here:
* Save `self.lengths` for the `df` mode as well. Not great because we basically repeat it two times.
* Put `if self.lengths is None: return None` between `if self.mode == 'df'` and `if self.mode == 'folder'`. In this way, all the conditions above will be satisfied.

Sorry, that was my bad. I didn't check how the `folder` mode behaves. The changes were amended.

karotchykau:
> Save `self.lengths` for the `df` mode as well.

To be honest, this solution can also be applied without any problems. Saving of lengths can be added to `Tokenizer.setups`. When these lengths are saved there, `if self.lengths is None: return None` will not be triggered in `get_lengths` if everything is initialized. Now `Tokenizer` with the `df` mode will have `self.lengths` that is not `None` as before. This solution is more "general" because it allows to call `self.lengths`.

I also encountered a bug when `tok_text_col` is different from the default name in `tokenize_df`. `'text_length'` is specified inside `get_lengths`, therefore, it will not be executed if you provide a different name.

Implemented this "general" solution. It works for every mode now. Please see the latest amended commit.

jph00:
Nice - that's looking much better!

PR #3670, opened 2022-05-31
Some missing bits on the W&B Callback

tcapelle:
There is an error on the CB, and I was adding some more tests. I was also thinking about adding some ThroughPutCallback to get `samples/sec` and `times` for train/validation, etc..


jph00:
Looks like there's a CI failure

Issue #3674, opened 2022-06-02
find_bs doesn't work with Tabular


muellerzr:
To recreate:
```python
from fastai.tabular.all import *

path = untar_data(URLs.ADULT_SAMPLE)
df = pd.read_csv(path/'adult.csv')
to = TabularPandas(df, y_names="salary",
            cat_names = ['workclass', 'education', 'marital-status', 'occupation',
                         'relationship', 'race'],
            cont_names = ['age', 'fnlwgt', 'education-num'],
            procs = [Categorify, FillMissing, Normalize])
find_bs(to)
```

Self-assigned, fix coming momentarily

Issue #3665, opened 2022-05-27
Accelerate notebook launcher fails with PyTorch dataloaders


DanteOz:
Please confirm you have the latest versions of fastai, fastcore, and nbdev prior to reporting a bug (delete one): YES

**Describe the bug**
The Accelerate notebook launcher fails when using PyTorch dataloaders. The equivalent script version of the notebook successfully runs using `accelerate launch`.

**To Reproduce**

1. Configure accelerate.
Using default configuration create with `write_basic_config()` from [notebook tutorial](https://docs.fast.ai/tutorial.distributed).

2. Run successful script to verify environment. 
https://gist.github.com/DanteOz/2713a7b993b696a29676051515a5873c

3. Run failing notebook launcher to reproduce bug.
https://gist.github.com/DanteOz/fc1fb96b917dd6bec19e71aca07339ca

**Expected behavior**
I would expect the notebook launcher to successfully launch `train()` function across multiple GPUs. 

**Error with full stack trace**

```
Some weights of the model checkpoint at bert-base-uncased were not used when initializing BertModel: ['cls.predictions.transform.LayerNorm.bias', 'cls.seq_relationship.weight', 'cls.predictions.decoder.weight', 'cls.predictions.transform.dense.bias', 'cls.predictions.transform.LayerNorm.weight', 'cls.predictions.transform.dense.weight', 'cls.seq_relationship.bias', 'cls.predictions.bias']
- This IS expected if you are initializing BertModel from the checkpoint of a model trained on another task or with another architecture (e.g. initializing a BertForSequenceClassification model from a BertForPreTraining model).
- This IS NOT expected if you are initializing BertModel from the checkpoint of a model that you expect to be exactly identical (initializing a BertForSequenceClassification model from a BertForSequenceClassification model).
Some weights of the model checkpoint at bert-base-uncased were not used when initializing BertModel: ['cls.predictions.transform.LayerNorm.bias', 'cls.seq_relationship.weight', 'cls.predictions.decoder.weight', 'cls.predictions.transform.dense.bias', 'cls.predictions.transform.LayerNorm.weight', 'cls.predictions.transform.dense.weight', 'cls.seq_relationship.bias', 'cls.predictions.bias']
- This IS expected if you are initializing BertModel from the checkpoint of a model trained on another task or with another architecture (e.g. initializing a BertForSequenceClassification model from a BertForPreTraining model).
[...]

PR #3675, opened 2022-06-02
[Enhancement] Fix CUDA spawn issues in Distributed, Tabular having no shape, and App Examples with notebook_launcher

muellerzr:
# Fix all bugs related to distributed learning with `notebook_launcher`

## What does this add?

* Moves spacy import errors that otherwise would allocated CUDA memory, not allowing the use of `notebook_launcher` when training on NLP tasks
* Adds a capability to `find_bs` that lets `TabularPandas` objects have a true length by checking if we have a `shape` or not, and if not just do the `len()`
* Adds a notebook mimicking the app_examples but uses the `notebook_launcher` instead

## Who is it for?

* Closes https://github.com/fastai/fastai/issues/3665
* Closes https://github.com/fastai/fastai/issues/3674

## Why is it needed?

I was running tests to see how `notebook_launcher` does with all of the examples fastai has, and noticed multiple bugs. This PR addresses them

### User-facing:
- `find_bs` now can accept objects that do not have a `shape` attribute

### Internal structure:

- Imports for spacy were moved into `SpacyTokenizer.__init__` directly to avoid automatic CUDA memory allocation

## Basic Usage Example(s):

With `find_bs`:
```python
test_eq(find_bs([1,2,3]), 3)
```

cc @jph00


jph00:
Thank you!

PR #3683, opened 2022-06-10
Workaround for performance bug in PyTorch with subclassed tensors

### LINKED ISSUE #3682: workaround pytorch subclass performance bug

warner-benjamin:
Resolves #3682 by adding `CastToTensor` callback as a workaround for performance bug in PyTorch with subclassed tensors.

Fixes https://github.com/fastai/fastai/commit/4cff258df422aa8577ec98f9b0519e3e4c5a2c96 by testing if learn.xb & learn.yb are tuples and applying the cast to `Tensor` if they are. (Unless I made a mistake in testing, b[:i] and b[i:] are tuples).

Unlike https://github.com/fastai/fastai/commit/4cff258df422aa8577ec98f9b0519e3e4c5a2c96, this PR adds the workaround as a callback so callbacks which use the input tensor type still can before it is casted to `Tensor` for the training performance increase. 

It also allows turning off the `Tensor` casting by removing the callback should it ruin a workflow. Although anyone who does this is encouraged to reimplement their own custom callback which casts to `Tensor` to get the free training performance increase.

Currently `CastToTensor.order` is right before `MixedPrecision`.


warner-benjamin:
I should have fixed the sync error, but if a nbdev `_all_` isn't on one line, it won't be added to the module's `__all__`.

jph00:
Much better - thanks!

PR #3684, opened 2022-06-11
chore: Included githubactions in the dependabot config

naveensrinivasan:
This should help with keeping the GitHub actions updated on new releases. This will also help with keeping it secure.

Dependabot helps in keeping the supply chain secure https://docs.github.com/en/code-security/dependabot

GitHub actions up to date https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot

https://github.com/ossf/scorecard/blob/main/docs/checks.md#dependency-update-tool
Signed-off-by: naveen <172697+naveensrinivasan@users.noreply.github.com>

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"


jph00:
Thanks @naveensrinivasan . Can you explain how this is different to the dependabot stuff that runs by default?

naveensrinivasan:
> Thanks @naveensrinivasan . Can you explain how this is different to the dependabot stuff that runs by default?

The actions aren't getting updated. For example https://github.com/fastai/fastai/blob/ff215d4f853b206b7f6a85320f5efc77b4b28399/.github/workflows/main.yml#L29 is at `v2` where as there is `v3.1.0`

With this change Dependabot would open PR's to the actions.

HTH

PR #3686, opened 2022-06-13
Add small doc for splitter

v-ahuja:
It was difficult to use `splitter` (especially with a `dataframe`) without seeing the usage in the course textbook.

I've added a small sentence here that will at least provide some intuition for how it should be used. Ideally, I'd like more documentation including how splitter processes a `dataframe` or some other data source, but this is the start that i could think of. suggestions welcome.

Also:

Ideally, instead of showing up as `No content` in the table, we could have the docs for `splitter`, `get_x` and `get_y` show up along with the rest of the args. However, this is more involved and looks like it needs changes to `nbdev`. I suggest this as a follow up enhancement to `nbdev`, which I can take a crack at.


v-ahuja:
@jph00 thanks in advance for reviewing.

fastai/data/block.py
```diff
@@ -77,7 +77,9 @@ def _zip(x): return L(x).zip()
 @docs
 @funcs_kwargs
 class DataBlock():
-    "Generic container to quickly build `Datasets` and `DataLoaders`"
```
jph00, line 80: Please don't change the formatting of existing fastai code, and when you add code please try to follow fastai standards. Our standard for docstrings is to use a single line and a single pair of double quotes.

jph00 (changes requested):
Thanks for your contribution. Just one requested change please.

v-ahuja:
changed. tks.

jph00:
Thanks!

PR #3697, opened 2022-06-19
Add correct Pillow dependency to fastai

v-ahuja:
`Image.Resampling` was added in pillow 9.1.0


jph00:
Let's not add a dep - will break anyone using pillow-simd. I'll do a workaround. Thanks anyway!

PR #3711, opened 2022-06-27
use py310 style union annotations

seeM:
This uses py310 style union annotations (e.g. `A|B`) over tuple style (e.g. `(A,B)`).

This should be released together with the corresponding PRs to https://github.com/fastai/nbdev/pull/636 and https://github.com/fastai/fastcore/pull/421.

I found all uses of the tuple convention using `rg --multiline-dotall -t py -t jupyter '\([^\(\)]*:\s*\(.*\).*\)`.

nbdev + fastcore + fastai tests pass locally on py38 and py310 on these branches.

cc @hamelsmu @jph00


seeM:
Ah, I thought I'd fixed nbdev export indexing but it still doesn't look right. I've made all of these draft PRs while I fix that...

jph00:
Is this ready for review now?

seeM:
@jph00 A question before this is ready for review: In https://github.com/fastai/fastai/pull/3711/commits/408c1060edf5d531e53b34c13ddbf114be0d681f I (1) fixed union annotations[^update-union-annots] and (2) reran all notebooks (using execnb, thanks to the Jupyter consistency updates). But that makes for a _huge_ diff. Would it be better to only fix the annotations?

[^update-union-annots]: I ended up doing this programmatically using `ast`, as you originally suggested, which I quite like! [Here's the notebook](https://wasimlorgat.com/tils/update-fastai-union-annotations-using-ast/). It uses `nbprocess` to export a tiny runnable package.

jph00:
Yes just fix the annotations - thanks for checking!

jph00:
Oh @seeM two more things:

- All notebooks with `|` type annotations will need `from __future__ import annotations` at the top, if they don't already
- There's a CI failure; I'm not sure why. I'm guessing there's some type mismatch going on...

seeM:
@jph00 done! And tests pass in [fork's CI](https://github.com/seeM/fastai/runs/7297988501) + locally

jph00:
Awesome job!

PR #3712, opened 2022-06-27
Update load_learner docs

misza222:
Documentation contradicting default values for cpu parameter


jph00:
Many thanks. Note that this is an *nbdev* project so the source code is in the notebooks, not the .py files - they're auto-synced when you run `nbdev_build_lib`. Would you mind update the nb file and building the lib? (No worries if not - I can do it at my end instead).

PR #3717, opened 2022-06-30
WIP: Added support for timm in unet

madhavajay:
This PR attempts to add `timm` models to the `unet_learner`, as per conversations during Live Coding 17: https://forums.fast.ai/t/live-coding-17/97166

I have several issues so far:
1) I don't know if theres a way to get the cut preferences from timm, I have added some code which tries to match them to fastai model types as a backup but they don't seem to match up so perhaps this is pointless and instead we either get them somewhere else or rely on manual user input

2) I tried training with them, but it seems like they use up a tonne of memory meaning my batch size can only be 1, not sure whats going on but something seems wrong especially considering i tried a smaller timm model `resnet18` than my default unet fastai model `resnet32`.

I am sure I have done something wrong, and would appreciate some direction on what to do next.


madhavajay:
So, I tried training it just to make sure its still working.

I got this far before my paperspace machine shutdown:
Screen Shot 2022-07-02 at 7 56 00 am

So I guess its definitely training but extremely slowly. For comparison I can do the same dataset, resnet34 with batch size 4 and get epochs of about 6 minutes on that free paperspace gpu. So assuming the architecture was equally complex and the batch size was 4 times lower that should only take 24 minutes per epoch.

It seems like when the model is allocated theres only 7% gpu memory in use, and then once training with 1 batch starts it goes to 87+%
Screen Shot 2022-07-01 at 6 06 02 pm

I guess perhaps I don't understand the model cutting code and how to use it with `timm` models. Any advice on how to debug this?

madhavajay:
Okay, I have changed the code use `timm.create_model` `features_only=True`.
So far it seems to be training a `convnext_tiny`
Screen Shot 2022-08-08 at 11 36 37 am

Issue #3727, opened 2022-07-06
LICENSE Tag is missing


bonysureliya:
Craete a license tag in README.md and add link it to LICENCE file

AnMeiDP:
I would like to work on this issue

tmabraham:
Since Jeremy does not want to add this, this issue can be closed
@jph00

PR #3729, opened 2022-07-06
Adding license

JGonong:
Resolves Issue: https://github.com/fastai/fastai/issues/3727


jph00:
Thanks but I'd rather not add this.

PR #3733, opened 2022-07-07
Update Optimizer docs & fix SGD+Mom WD comparison

warner-benjamin:
Update Optimizer docs with additional clarity relative to PyTorch optimizers.

Fix the SGD with Momentum Weight Decay comparison test/example.

Move `OptimWrapper` usage example to right after the `OptimWrapper` definition.

### DIFF
--- a/nbs/12_optimizer.ipynb
+++ b/nbs/12_optimizer.ipynb
@@ -243,7 +243,9 @@
    "source": [
     "`params` will be used to create the `param_groups` of the optimizer. If it's a collection (or a generator) of parameters, it will be a `L` containing one `L` with all the parameters. To define multiple parameter groups `params` should be passed as a collection (or a generator) of `L`s.\n",
     "\n",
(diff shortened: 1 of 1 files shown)


nbs/12_optimizer.ipynb
```diff
@@ -243,7 +243,9 @@
    "source": [
```
jph00, line 243: <p>I think this addition could be a bit distracting/confusing. Especially with the "important" at the start. For most people I don't think it's important at all! Generally speaking I don't think you'd want wd applied to normalization bias terms, would you?</p><p> </p><p>If there are situations where someone wants to make a fastai optimizer work just like pytorch's, how about we instead create a little section in this notebook called "differences to pytorch optimizers", or something like that. That way, we can remove these notes scattered through the notebook, and have all that info in one place.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/fastai/fastai/pull/3733/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='13'/>

PR #3737, opened 2022-07-15
Add split value argument to ColSplitter

DanteOz:
Adds an optional parameter to ColSplitter to specify which value to split on. Done with intention of simplifying the use of a k-fold validiation index column with the DataBlocks API.  Example usage would be:

```python
df = pd.DataFrame({
    'a': [0,1,2,3,4,5], 
    'fold': [1,2,3,1,2,3]
})
splits = ColSplitter('fold', 3)(df)
```


DanteOz:
@jph00 Ready for review. I added support for splitting on a list of values, given the discord discussion with Benjamin.

jph00:
Good stuff!

PR #3743, opened 2022-07-24
fix same file error message when resizing image

cvergnes:
Issue : https://github.com/fastai/fastai/issues/3744


jph00:
Many thanks!

PR #3760, opened 2022-08-01
Fixed fastai tutorial link in README.md

khaled5321:


jph00:
Thanks - this file is auto-generated so I'll fix the notebook.

PR #3766, opened 2022-08-06
Doc bug

Isaac-Flath:
@jph00 

I believe that `doc` should be importing `show_doc` from nbdev.   It's currently importing `nbdev.showdoc.doc`, which doesn't exist so I think it should be importing `show_doc` instead.

Should the `doc` function also be renamed `show_doc` to have consistency with `nbdev`?


jph00:
No, `doc` and `show_doc` are different. I guess we need to add `doc`.

PR #3772, opened 2022-08-09
Fix 04_data.external.ipynb copy-paste issue

rayman1104:
While reading [the docs](https://docs.fast.ai/data.external.html), that was very confusing that running the same piece of code lead to different results.

But I believe it's just a copy-paste typo.


jph00:
Oops! Thanks for the keen eye and the fix.

PR #3793, opened 2022-09-13
_array2tensor() now accepts **kwargs

### LINKED ISSUE #3781: _array2tensor() got an unexpected keyword argument 'requires_grad'

```
fastai  : 2.7.9
fastcore: 1.5.14
```

**Describe the bug**
The code below throws an error when using `requires_grad=True`

**To Reproduce**
```py
import fastai
import torch

this_is_fine = fastai.torch_core.tensor([3.], requires_grad=True)
this_is_not = fastai.torch_core.tensor(3., requires_grad=True)
this_is_also_fine = torch.tensor(3., requires_grad=True)
```

**Expected behavior**
It should not throw an error.

```
File ~\.virtualenvs\fastbook--K4-LsvW\lib\site-packages\fastai\torch_core.py:152, in tensor(x, *rest, **kwargs)
    144     if len(rest): x = (x,)+rest
    145     # There was a Pytorch bug in dataloader using num_workers>0. Haven't confirmed if fixed
    146     # if isinstance(x, (tuple,list)) and len(x)==0: return tensor(0)
    147     res = (x if isinstance(x, Tensor)
    148            else torch.tensor(x, **kwargs) if isinstance(x, (tuple,list))
    149            else _array2tensor(x) if isinstance(x, ndarray)
    150            else as_tensor(x.values, **kwargs) if isinstance(x, (pd.Series, pd.DataFrame))
    151 #            else as_tensor(array(x, **kwargs)) if hasattr(x, '__array__') or is_iter(x)
--> 152            else _array2tensor(array(x), **kwargs))
    153     if res.dtype is torch.float64: return res.float()
    154     return res

TypeError: _array2tensor() got an unexpected keyword argument 'requires_grad'
```

**Additional context**
There are two problems:
* The source passes `**kwargs` to `_array2tensor`, but that function doesn't accept `**kwargs`
* But also, the logic within `tensor` doesn't handle the case of a single float being passed in. Perhaps better logic would be to test for all the special cases that fastai wants to handle, then fall back to passing all args to `torch.tensor`

SaadAhmedGit:
fixes #3781 

The aforementioned function did not previously accept **kwargs even though it was being passed to it in line 152 which caused it to throw an error when only a single float was passed to torch_core.tensor() as mentioned in issue #3781. This adds **kwargs to _array2tensor() and uses it to set the requires_grad attr of the returned tensor from the value given through **kwargs

### DIFF
--- a/fastai/torch_core.py
+++ b/fastai/torch_core.py
@@ -129,13 +129,13 @@ def __array_eq__(self:Tensor,b):
     return torch.equal(self,b) if self.dim() else self==b
 
 # %% ../nbs/00_torch_core.ipynb 39
-def _array2tensor(x):
+def _array2tensor(x, **kwargs):
     if x.dtype==np.uint16: x = x.astype(np.float32)
     # windows default numpy int dytpe is int32, while torch tensor default int dtype is int64
     # https://github.com/numpy/numpy/issues/9464
     if sys.platform == "win32":
         if x.dtype==np.int: x = x.astype(np.int64)
-    return torch.from_numpy(x)
+    return torch.from_numpy(x).requires_grad_(kwargs.get("requires_grad"))
 
 # %% ../nbs/00_torch_core.ipynb 40
 @use_kwargs_dict(dtype=None, device=None, requires_grad=False, pin_memory=False)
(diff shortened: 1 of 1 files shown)


jph00:
Many thanks @SaadAhmedGit . This is actually an nbdev project so changes need to be made in the notebook. I'm also not sure the approach here is the best -- I'll fix it at our end.

SaadAhmedGit:
> This is actually an nbdev project so changes need to be made in the notebook.

Thanks @jph00 for educating me about fastai being an nbdev project. I will try to commit to the notebooks in the future.

> I'm also not sure the approach here is the best

Yes I too felt like there was something wrong with the previous solution. I have come up with another solution but I'll commit it in the notebook if you can have a quick look at it and tell me what's wrong with my approach so I can write something better and commit that instead if, that sounds ok.
```
def _array2tensor(x, **kwargs):
    if x.dtype==np.uint16: x = x.astype(np.float32)
    # windows default numpy int dytpe is int32, while torch tensor default int dtype is int64
    # https://github.com/numpy/numpy/issues/9464
    if sys.platform == "win32":
        if x.dtype==np.int: x = x.astype(np.int64)

    # removing and storing "requires_grad" from kwargs since torch.as_tensor() doesn't accept it
    req_grad_from_kwargs = kwargs.pop("requires_grad")
    t = torch.as_tensor(x, **kwargs)
    t.requires_grad_(req_grad_from_kwargs)
    return t
```

jph00:
I actually committed a fix today you can look at

jph00:
Sorry my bad I forgot to push it! here's the line:

https://github.com/fastai/fastai/blob/master/fastai/torch_core.py#L148

PR #3794, opened 2022-09-13
update README with nbdev

kurianbenoy:
The commands to contribute in README of fastai looks outdated and are not available in nbdev2.

image

**Note to reviewer**

- I noticed the `index.ipynb` is updated, but the README has not been kept upto date with it. 
- I am not sure, if the `_quarto.yml`  change I made is wrong, yet because of that the `nbdev_readme` command was throwing an error.


jph00:
Many thanks.

PR #3797, opened 2022-09-16
Added **kwargs to _array2tensor()

SaadAhmedGit:
As discussed in the comments of #3781,
You could not create a tensor using an ndarray with named arguments. The aforementioned function now accepts **kwargs and creates a tensor using torch.as_tensor() which shares memory with the ndarray.

In the following code snippet:
```
x = np.array([3., 4., 5.])
tensor = fastai.torch_core.tensor(x, requires_grad=True, pin_memory=True)

print(f"Is pinned: {tensor.is_pinned()}")
print(f"Requires gradient: {tensor.requires_grad}")
```
## Previous behaviour:
**Expected:**
```
Is pinned: True
Requires gradient: True
```
**Output:**
```
Is pinned: False
Requires gradient: False
```
## After commit:
**Expected:**
```
Is pinned: True
Requires gradient: True
```
**Output:**
```
Is pinned: True
Requires gradient: True
```


SaadAhmedGit:
I just found out that you have you to build the notebooks too (I should read the nbdev stuff in more detail). So I'll open this once I have done that.

jph00:
That's a clever approach. Would this work and be a bit cleaner?:

```python
def _array2tensor(x, requires_grad=False, pin_memory=False, **kwargs):
    if x.dtype==np.uint16: x = x.astype(np.float32)
    # windows default numpy int dtype is int32, while torch tensor default int dtype is int64
    # https://github.com/numpy/numpy/issues/9464
    if sys.platform == "win32" and x.dtype==np.int: x = x.astype(np.int64)
    t = torch.as_tensor(x, requires_grad=requires_grad, pin_memory=pin_memory, **kwargs)
    t.requires_grad_(requires_grad)
    if pin_memory: t.pin_memory()
    return t
```

SaadAhmedGit:
@jph00 yeah your approach is far cleaner but your code doesn't work because `as_tensor()` doesn't accept `requires_grad` and `pin_memory` keyword arguments. I just erased them in `as_tensor()`'s arguments and it works correctly as we are already changing these attributes after that. I have also made the commit

Just changed this from your code:
```
t = torch.as_tensor(x, requires_grad=requires_grad, pin_memory=pin_memory, **kwargs)
```
to this:
```
t = torch.as_tensor(x, **kwargs)
```

PR #3799, opened 2022-09-18
Update documentation

kurianbenoy:
- Removed some of the old nbdev commands with the equivalent commands in nbdev2
- Fix broken link


jph00:
Thanks!

PR #3803, opened 2022-09-26
Show Quick Start imports

### LINKED ISSUE #3771: Quick_start should show the import cell
https://docs.fast.ai/quick_start.html hides the "imports" cell.  

This leads to an immediate and confusing set of errors when attempting to reproduce locally. 

Suggested fix:  
1. remove the '#hide' in https://github.com/fastai/fastai/blob/master/nbs/quick_start.ipynb in cell 2.  
2. Move the cell down a bit.

Why I am not submitting a PR:  You all will be faster at doing it, and moving the cells around than I will be :), given that it's a ipynb.

Ben-Karr:
Fixes #3771.

### DIFF
--- a/nbs/quick_start.ipynb
+++ b/nbs/quick_start.ipynb
@@ -27,7 +27,6 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "#|hide\n",
     "from fastai.vision.all import *\n",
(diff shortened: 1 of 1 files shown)


jph00:
Thanks!

PR #3806, opened 2022-09-27
Migrate callouts to Quarto format

warner-benjamin:
I migrated callouts from nbdev v1 to Quarto format. Docs appear to render correctly using `nbdev_preview`.


warner-benjamin:
No script, I ran `nbdev_migrate` against the notebooks and the callout migration worked without issue.

I did lightly modify `migrate_nb` before running because it was converting the markdown H1 titles and blockquote subtitles to frontmatter and was unsure if that's desired behavior.

PR #3808, opened 2022-09-28
Add `img_cls` param to `ImageDataLoaders`

tcapelle:
This PR adds a param to the `ImageDataLoaders` class so we can pass an Image class to the underlying `DataBlock`.
This is particularly useful for passing `PILImageBW` for MNIST.

```python
dls = ImageDataLoaders.from_folder(path, img_cls=PILImageBW)
x,y = dls.one_batch()
x.shape
> [64, 1, 28,28]
```


jph00:
Good idea!

PR #3827, opened 2022-10-30
Add torch_[(save)|(load)]_kwargs to fastai.[(save)|(load)]_model

JonathanGrant:
This lets us do nice things like set `pickle_module` to `cloudpickle`.

I went with `**kwargs` approach so that any updates to `torch.save` or `load` are automatically supported.
However, I decided to keep the existing pass through kwargs as is in case someone is relying on them as positional arguments. We can make this a breaking change if that is preferred, though.

(btw thank you for the excellent library and lectures!)

### DIFF
--- a/fastai/learner.py
+++ b/fastai/learner.py
@@ -34,13 +34,13 @@ def mk_metric(m):
     return m if isinstance(m, Metric) else AvgMetric(m)
 
 # %% ../nbs/13a_learner.ipynb 15
-def save_model(file, model, opt, with_opt=True, pickle_protocol=2):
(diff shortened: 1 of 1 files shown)


jph00:
Many thanks for your PR. fastai is an nbdev project, which means that the notebooks are the source files, and the .py files are generated from them. Could you please update the source notebooks with your changes?

JonathanGrant:
Fixed in #3831

JonathanGrant:
Wait, I think you wanted me to have both changes. My bad! Will merge this branch into the other.

PR #3828, opened 2022-10-31
PyTorch 1.13 Compatibility

warner-benjamin:
PyTorch 1.13 introduced a subclassed Tensor formatting issue. A potential workaround is pytorch/pytorch#82766 which I use for the fastai `TensorBase` compatibility patch in `torch_core`.

PyTorch 1.13 also adds "differentiable" as another `optim` hyperparameter and it looks like PyTorch 1.14 will add "fused" for at least a subset of optimizers. Since I modified `OptimWrapper` not to error out in #3821, I changed the `OptimWrapper` tests to only check for the subset of hyperparameters fastai cares about.

Tested locally running the cuda and slow flags without issue.


jph00:
Thanks so much!

PR #3830, opened 2022-10-31
nbdev CI: add notebook-library sync check

warner-benjamin:
I believe this PR should resolve the issue with the CI not checking if the notebooks and library source code are in sync. I based it on my [fastxtend nbdev CI](https://github.com/warner-benjamin/fastxtend/blob/main/.github/workflows/nbdev.yml), which as far as I can tell is working without issue.

Since the fastai library and notebooks are currently out of sync, this PR should error out. Assuming the PR CI runs the changed `main.yml` and not the master branch `main.yml`.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,9 +1,9 @@
 name: CI
-on: 
+on:
   workflow_dispatch: #allows repo admins to trigger this workflow from the Actions tab
   pull_request:
   push:
-    branches: 
+    branches:
       - master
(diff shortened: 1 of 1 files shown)


warner-benjamin:
It errored out as I anticipated, so I think this means it can be merged.

warner-benjamin:
Updated to use the [nbdev-ci workflow](https://github.com/fastai/workflows/tree/master/nbdev-ci) per feedback on the fastai discord

PR #3831, opened 2022-11-01
Add torch save and load kwargs

JonathanGrant:
This lets us do nice things like set pickle_module to cloudpickle.

I went with **kwargs approach so that any updates to torch.save or load are automatically supported.
However, I decided to keep the existing pass through kwargs as is in case someone is relying on them as positional arguments. We can make this a breaking change if that is preferred, though.

(btw thank you for the excellent library and lectures!)


warner-benjamin:
The CI error here should be resolved when #3828 is merged into the master branch.

jph00:
Thanks for getting this all sorted out! :D

PR #3823, opened 2022-10-23
minor typos correction in 00_torch_core, and 01_layers

EmbraceLife:
The details can be found in this forum [thread](https://forums.fast.ai/t/help-reading-fastcore-and-fastai-docs/100168/8?u=daniel)

PR #3832, opened 2022-11-01
Sync notebooks with library and rollback `to_float` change

warner-benjamin:
This PR resolves the out of sync issue with fastai notebooks and library.

It also rolls back the issue unintentionally introduced in #3823 to `to_float`. fastai uses `to_float` to convert fp16 to fp32 in mixed precision training, so the unsynced change of the function to match the doc string would have resulted in `NonNativeMixedPrecision` not convert fp16 predictions to fp32 predictions as expected.

I also clarified a few doc strings, as `to_half` and `to_float` only modify [floating point](https://pytorch.org/docs/stable/generated/torch.is_floating_point.html) Tensors, and both along with `to_cpu` recursively apply to more than just lists of tensors, so I removed the doc string reference to lists.

### DIFF
--- a/fastai/layers.py
+++ b/fastai/layers.py
@@ -73,7 +73,7 @@ def Flatten(self, x):
 # %% ../nbs/01_layers.ipynb 15
 @module(tensor_cls=TensorBase)
 def ToTensorBase(self, x):
-    "Remove `tensor_cls` to x"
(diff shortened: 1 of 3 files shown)


jph00:
Many thanks :D

PR #3833, opened 2022-11-01
Fix image src at README.md file

yaserabdelaziz:


jph00:
Many thanks. The README is autogenerated from index.ipynb, so we need to fix it there too. Would you be able to make that change?

Issue #3836, opened 2022-11-09
Save model callback breaks multi-gpu training


caleb-depotanalytics:
Hey!

Not a huge deal, but the SaveModelCallback breaks multi-gpu training because suddenly all threads try to write to the same .pth file.

My workaround is:
```
    # redefine the save function of the model saving callback to be a 'multicore save' where there won't be a file-writing conflict
    # in which every core tries to write to the same file at the same time...
    save_cb = SaveModelCallback(monitor='valid_loss', min_delta=0.0, with_opt=False)
    def mc_save(self,name): self.last_saved_path = rank0_first(lambda: self.learn.save(name, with_opt=self.with_opt))
    save_cb._save = functools.partial(mc_save, save_cb)
```

but I'd rather not do that.

caleb-depotanalytics:
```
ubuntu@ip-172-31-71-13:~/soybean-production-train$ pip freeze | grep fastai
fastai==2.7.6
fastai2==0.0.30
```

muellerzr:
@jph00 feel free to assign to me here!

muellerzr:
@caleb-depotanalytics can you verify using `pip install git+https://github.com/muellerzr/fastai/fix-save` works for you? Thanks!

PR #3839, opened 2022-11-14
Fix load hanging in distributed processes

muellerzr:
Should solve https://github.com/fastai/fastai/issues/3836

Basically similar to the `export` function we need to include a `distrib_barrier()` in here so that all the processes are waiting and synced here before loading in the model. The `wait_for_everyone()` I found actually doesn't apply inside the `Learner` because since it's added during the context manager that version of the Learner is stored differently (so it doesn't have `learn.accelerator` for instance). Fun memory things 🙃 

cc @jph00


muellerzr:
@jph00 not sure where the diff is coming in, `nbdev_export` shows nothing new changing when building from 2.3.9 and 2.3.10 😕

PR #3842, opened 2022-11-27
Add learner.predict() usage example

bencoman:
2022 Part 1 Lesson 1 has the following example which throws away the middle return value, which is useful when students start doing their own experiments when the predicted answer is not `probs[0]`.

image

but as pointed out here: https://forums.fast.ai/t/how-to-output-confidence-of-classiifcation-models/102221?u=bencoman
the documentation is not clear on how to use that middle return value - so adding a beginner example to the doc (not sure if this is the best way to do this.)


bencoman:
I'm not sure why this dff showed up. I didn't consciously make a change there.
image

jph00:
Many thanks @bencoman . I think you need a blank line before your list to make it appear correctly. I suggest removing the name of each output, since they don't really have names (it's an unnamed tuple). Also you might want to add the `with_input` optional output to the list.

jph00:
Closing due to lack of activity - @bencoman feel free to reopen if you have a chance to get back to this.

PR #3843, opened 2022-12-03
Allow passing a pre-initialized `OptimWrapper` to `Learner`

warner-benjamin:
The [`OptimWrapper` documentation](https://docs.fast.ai/optimizer.html#optimwrapper) suggests that using a pre-defined optimizer with `Learner`'s opt_func argument should work:

> Or if you already have an existing one, pass in only opt:
> ```python
> opt = torch.optim.SGD([tensor([1,2,3])], lr=1e-2)
> opt_func = OptimWrapper(opt=opt)
> ```

however, following this example and passing `opt_func` to `Learner.opt_func` results in an error, as `Learner.create_opt` only accepts callable methods and not initialized objects.

This PR modifies `Learner` to accept a pre-initialized `OptimWrapper` to the opt_func argument by modifying  `Learner.create_opt` to set a pre-initialized `OptimWrapper` to `Learner.opt`. To mimic the current behavior of always initializing an optimizer from scratch, `create_opt` will call `clear_state` on a pre-initialized `OptimWrapper`.

This allows a straightforward method of using custom parameter groups in a PyTorch style optimizer:

```python
param_groups = custom_param_group_method(model)
opt = OptimWrapper(opt=optim.AdamW(param_groups, ...))
learn = Learner(dls, model, opt_func=opt, ...)
```

To prevent a pre-initialized optimizer's state from being cleared on the first fit, users can use the current behavior of passing the optimizer directly to `Learner.opt`. However, in this case manually calling `Learner.fit(..., reset_opt=True)` will cause an error. Instead of using reset_opt, users will need to redefine the optimizer.

This PR also updates some of the existing `Learner` docments to better match the docment style and more accurately describe the argument behavior.


warner-benjamin:
When I run the tests locally, CI error does not occur. I also did not change anything related to default callbacks. Investigating.

warner-benjamin:
@jph00 The CI failure appears to be random and due to a change made in the development branch of nbdev, which the fastai CI currently uses.

My updated passed the CI, but would have been safe to merge even if it did not, as I tested it with nbdev 2.3.9 and it passed every time.

fastai/learner.py
```diff
@@ -97,22 +97,22 @@ def before_epoch(self):
...
-                 path=None,  # Parent directory to save, load, and export models
-                 model_dir='models',  # Subdirectory to save and load models
-                 wd=None,  # Weight decay
-                 wd_bn_bias=False,  # Apply weight decay to batchnorm bias params?
-                 train_bn=True,  # Always train batchnorm layers?
-                 moms=(0.95,0.85,0.95),  # Momentum
-                 default_cbs:bool=True  # Include default callbacks?
-                ):
+        dls:DataLoaders, # `DataLoaders` containing fastai or PyTorch `DataLoader`s
+        model:callable|nn.Module, # PyTorch model for training or inference
```
jph00, line 101: `nn.Module` is `callable` so I don't think this change is neede

fastai/learner.py
```diff
@@ -97,22 +97,22 @@ def before_epoch(self):
...
+        loss_func:nn.Module|None=None, # Loss function. Defaults to `dls` loss
```
jph00, line 102: I believe any callable is OK as a loss func, so I think this change is incorrect

fastai/learner.py
```diff
@@ -97,22 +97,22 @@ def before_epoch(self):
...
+        opt_func:Optimizer|OptimWrapper=Adam, # Optimization function for training
```
jph00, line 103: I think (IIRC) any callable works, although not all features work with any callable. We should use the most general type which can work in some situation

fastai/learner.py
```diff
@@ -97,22 +97,22 @@ def before_epoch(self):
...
+        lr:float|slice=defaults.lr, # Default learning rate
+        splitter:callable=trainable_params, # Split model into parameter groups. Defaults to one parameter group
+        cbs:Callback|list|None=None, # `Callback`s to add to `Learner`
```
jph00, line 106: I think any collection works - doesn't have to be a `list`.

fastai/learner.py
```diff
@@ -97,22 +97,22 @@ def before_epoch(self):
...
+        metrics:callable|list|None=None, # `Metric`s to calculate on validation set
```
jph00, line 107: Ditto

fastai/learner.py
```diff
@@ -97,22 +97,22 @@ def before_epoch(self):
...
+        path:str|Path|None=None, # Parent directory to save, load, and export models. Defaults to `dls` `path`
+        model_dir:str|Path='models', # Subdirectory to save and load models
+        wd:float|None=None, # Default weight decay
```
jph00, line 110: I think any numeric might be OK here?

fastai/learner.py
```diff
@@ -97,22 +97,22 @@ def before_epoch(self):
...
+        wd_bn_bias:bool=False, # Apply weight decay to normalization and bias parameters
+        train_bn:bool=True, # Train frozen normalization layers
+        moms:tuple[float,...]=(0.95,0.85,0.95), # Default momentum for schedulers
```
jph00, line 113: We don't use generic types in any fast.ai lib

fastai/learner.py
```diff
@@ -272,17 +276,17 @@ def validate(self, ds_idx=1, dl=None, cbs=None):
...
-                  with_decoded: bool=False, # Whether to decode based on loss function passed
-                  with_loss : bool=False, # Whether to return losses
-                  act: Any=None, # Option to pass Activation function to predict function
-                  inner : bool=False, # Tells that it's used internally used anywhere like in another training loop
-                  reorder : bool=True, # To order the tensors appropriately
-                  cbs: list | None =None, # Option to pass `Callbacks` to predict function
-                  **kwargs
-                 )-> tuple:
+        ds_idx:int=1, # `DataLoader` to use for predictions if `dl` is None. 0: train. 1: valid
+        dl:DataLoader|None=None, # `DataLoader` to use for predictions, defaults to `ds_idx=1` if None
```
jph00, line 280: I suspect other iterables might work - e.g. a progress bar wrapped DL?

fastai/learner.py
```diff
@@ -272,17 +276,17 @@ def validate(self, ds_idx=1, dl=None, cbs=None):
...
+        with_input:bool=False, # Return inputs with predictions
+        with_decoded:bool=False, # Return decoded predictions
+        with_loss:bool=False, # Return per item loss with predictions
+        act:Any=None, # Apply activation to predictions, defaults to `self.loss_func`'s activation
```
jph00, line 284: We can actually remove `:Any` since that's always implied

fastai/learner.py
```diff
@@ -272,17 +276,17 @@ def validate(self, ds_idx=1, dl=None, cbs=None):
...
+        inner:bool=False, # If False, create progress bar, show logger, use temporary `cbs`
+        reorder:bool=True, # Reorder predictions on dataset indicies, if applicable
+        cbs:Callback|list|None=None, # Temporary `Callback`s to apply during prediction
```
jph00, line 287: I know you didn't put `list` here - but we should fix it anyway!

jph00 (changes requested):
This looks great @warner-benjamin . My only request is to avoid restricting types more than necessary. You've added some restrictions which are too tight -- and it's also made me realise some existing ones are also too tight!

fastai/learner.py
warner-benjamin: I added `nn.Module` since I regularly see questions about what kind of model fastai accepts and thought specifying it is a PyTorch Module via the type hint might be useful. Can remove.

fastai/learner.py
warner-benjamin: At minimum, `opt_func` needs to be a callable which accepts param groups from a splitter, and a learning rate per, `self.opt = self.opt_func(self.splitter(self.model), lr=self.lr)`, and then have a `step` implementation, etc. So I thought `Optimizer|OptimWrapper` was more useful than callable. Can change.

fastai/learner.py
warner-benjamin: The convention of `Object|list` is incorrectly used all over the codebase to represent a collection/iterable of the first object following the docment sprint. I kept with the current codebase since we don't have a defined method of any iterable/collection. What should it be changed to?

fastai/learner.py
warner-benjamin: fastai has `moms:tuple=(0.95,0.85,0.95)` in multiple places throughout the library. Prefer that formatting?

fastai/learner.py
warner-benjamin: Not sure what this should to be changed to. `DataLoader` seems to be useful information to have in the type hint.

warner-benjamin:
Will change any of the type hints to what you want. I tried to follow the convention set by #3639 and attempted to add some clarity. I added a few comments and questions on the code review.

fastai/learner.py
jph00: The types aren't really just hints - they have to handle at minimum all the types that can be accepted, otherwise people that use them will get errors. In this case, I believe the only type that would be accurate is `iterable` -- or just have no type, in which case it defaults to `Any`.

fastai/learner.py
jph00: Yes exactly - just remove the generic constraint and leave it as `tuple`.

fastai/learner.py
jph00: Thanks for letting me know - I think we should change `list` to `collections.abc.MutableSequence` when it's used in this way.

fastai/learner.py
jph00: Ah yes you're right - let's leave it as you have it then.

fastai/learner.py
jph00: For providing that kind of tip to the user, we should use markdown docs or docments, not types.

jph00:
Many thanks @warner-benjamin -- let me know if you have any outstanding questions.

fastai/learner.py
warner-benjamin: Technically, it should be `Callable[...,Optimizer|OptimWrapper]|OptimWrapper`, but per fastai/nbdev#1179 this would probably cause `show_doc` to error out, hence my original suggestion.

fastai/learner.py
warner-benjamin: Changing all of them is outside the scope of this PR. I can change these few instances to `cbs:Callback|MutableSequence[Callback]|None`, the less technically correct `cbs:Callback|MutableSequence|None`, or leave it as `Callback|list` to match the rest of the library.

Alternatively, before merging #3639 with the current `Object|list` convention, you came up with the custom type hint of [`Listified[Object]`](https://github.com/fastaidocsprint/styleguide#custom-type-aliases) to represent `Object|MutableSequence[Object]`. If you prefer, I could use that instead. And perhaps add `MutableSequence` to the custom Listified type hint.

fastai/learner.py
jph00: Right - we don't use generics.

fastai/learner.py
jph00: We can't use `Listified[Object]` since it's a generic type. But we can use `Listified`. Since `list` is a `MutableSequence` we can simply change the `Listified` type -- no need to add to it.

warner-benjamin: I'm still not completely sure what you want the type hint to be changed too.

jph00: `Callback|MutableSequence|None`

jph00 (approved):

jph00:
Many thanks!

jph00:
I'm replacing `list` in all types I can find with `MutableSequence` now FYI and will push shortly.

PR #3849, opened 2022-12-20
Fix broken image in README

harlantwood:


jph00:
Many thanks for the PR. The README is actually built from a notebook, and making this change in the notebook would break the image there! I'll copy the image over to fix it.

harlantwood:
Sounds good, my pleasure.  I am blown away by your intro course, just starting the videos, already life changing, so empowering, after years of reading ML books… ✨🙏✨

PR #3850, opened 2022-12-21
Changed to int (it is the same as np.int) but will work on later vers…

Mazchoo:
Changed to np.int to int (np.int is an alias of int) and will still work on later versions of numpy.


jph00:
Thanks! This is an nbdev project, so the source code lives in the notebooks, and the py modules are generated from there. Would you be able to modify the source notebook? (If not, just let me know, and I'll do it myself).

Mazchoo:
Thanks for replying, I changed the line in a notebook.

jph00:
Many thanks!

PR #3852, opened 2022-12-30
Add option to (optionally) save confusion matrix plot

aspiringastro:
This PR adds an optional parameter `save_plot` to the `plot_confusion_matrix` function which allows offline analysis of models and their confusion matrix across several tuning or iterations.


jph00:
Thanks for the PR. I think I'd rather just let people use plt.savefig directly themselves rather than add this to the method.

Issue #3824, opened 2022-10-23
albumentations tutorial typo


levyvix:
https://docs.fast.ai/tutorial.albumentations.html#using-different-transform-pipelines-and-the-datablock-api

![image](https://user-images.githubusercontent.com/9471166/197395133-f85a6a1c-03c5-47b5-97cb-0b75a69af7f1.png)

PR #3855, opened 2023-01-04
fixing a micro typo

Shubhamai:
It fixes the issue originally described in https://github.com/fastai/fastai/issues/3824


jph00:
Thanks!

PR #3856, opened 2023-01-10
Removed line with no effect from tabular_learner

pablogps:
Removed `to = dls.train_ds` line (41) from `tabular_learner`, since it looks like it is never used and has no other effect.


jph00:
Thanks for the PR. Since this is an nbdev project, any changes need to be made in the source notebook, and then exported to the .py files.

pablogps:
Thanks for taking the time, it seems like I'm a bit obsolete! Maybe this is a good excuse to finally look into nbdev a bit.

PR #3857, opened 2023-01-12
Move training-only step to separate function in `Learner`

kunaltyagi:
* Separates difference between post-training and post-validation into a single function
* Allows overloading a much simpler function to change learner behavior


kunaltyagi:
@jph00 Can we rerun the CI?

jph00:
Thank you!

Issue #3785, opened 2022-09-02
"default_device" logic is repeated twice, related to `mps` / OSX support.


gregglind:
https://github.com/fastai/fastai/blob/c5b9aa050e1ed382b40a7f772a07d74453fdcacb/fastai/torch_core.py#L259

@jph00, previous fixes in #3719 repeated the logic.  

While there could we add a docstring for how to get a 'cpu only' build?  `use=0`, or use='cpu') would be great for that.

Re-assign it to me as a bug if you like, and I will be happy to submit a patch.

jph00:
Assigned!

PR #3858, opened 2023-01-14
Fix default_device to correctly detect + use mps (Apple Silicon)

wolever:
This fixes two minor bugs preventing `fastai.torch_core.default_device` from detecting `mps` (Apple Silicon) GPUs.

1. The `torch.backends.mps.is_available` attribute does not exist in PyTorch 1.13.1. If this attribute is unavailable, this patch uses `torch.has_mps` instead.
2. A minor logic bug in `default_device` prevented `_has_mps` from being called.

Additionally, this fixes https://github.com/fastai/fastai/issues/3785 by removing the duplicate `default_device` function.

### DIFF
--- a/fastai/torch_core.py
+++ b/fastai/torch_core.py
@@ -258,28 +258,22 @@ def to_float(b):
 defaults.use_cuda = None
 
 # %% ../nbs/00_torch_core.ipynb 71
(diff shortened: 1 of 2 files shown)


jph00:
Many thanks.

Issue #3861, opened 2023-01-15
dataloader error: 'numpy' has no attribute 'int'


rohanselvaradov:
Please confirm you have the latest versions of fastai, fastcore, and nbdev prior to reporting a bug (delete one): YES

**Describe the bug**
I'm running the [simple image classifier](https://www.kaggle.com/code/jhoward/is-it-a-bird-creating-a-model-from-your-own-data) from the fastai course in a Jupyter Notebook on my PC's IDE (PyCharm), but I'm getting the message

> Could not do one pass in your dataloader, there is something wrong in it.

**To Reproduce**
1. Copy the Kaggle notebook into PyCharm 
2. Run upto and including the codeblock which creates the dataloader
```
dls = DataBlock(
    blocks=(ImageBlock, CategoryBlock),
    get_items=get_image_files,
    splitter=RandomSplitter(valid_pct=0.2, seed=21),
    get_y=parent_label,
    item_tfms=[Resize(192, method='squish')]
).dataloaders(path, bs=32)
```

**Expected behavior**
Makes the dataloader without errors

**Error with full stack trace**
```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[8], line 1
----> 1 dls = DataBlock(
      2     blocks=(ImageBlock, CategoryBlock),
      3     get_items=get_image_files,
      4     splitter=RandomSplitter(valid_pct=0.2, seed=21),
      5     get_y=parent_label,
      6     item_tfms=[Resize(192, method='squish')]
      7 ).dataloaders(path, bs=32)
      9 dls.show_batch(max_n=6)

File ~\Documents\Coding\fastai\venv\lib\site-packages\fastai\data\block.py:157, in DataBlock.dataloaders(self, source, path, verbose, **kwargs)
    155 dsets = self.datasets(source, verbose=verbose)
    156 kwargs = {**self.dls_kwargs, **kwargs, 'verbose': verbose}
--> 157 return dsets.dataloaders(path=path, after_item=self.item_tfms, after_batch=self.batch_tfms, **kwargs)

File ~\Documents\Coding\fastai\venv\lib\site-packages\fastai\data\core.py:334, in FilteredBase.dataloaders(self, bs, shuffle_train, shuffle, val_shuffle, n, path, dl_type, dl_kwargs, device, drop_last, val_bs, **kwargs)
    332 dl = dl_type(self.subset(0), **merge(kwargs,def_kwargs, dl_kwargs[0]))
[...]

PR #3865, opened 2023-01-23
solved error with new numpy version issue :- #3861

sanjaybora04:
new numpy version has removed numpy.int, use numpy.int32 instead


jph00:
This is an nbdev project so changes need to be made in the source notebook. Shall I do that, or would you like to?

sanjaybora04:
I have exams from Tommorow so it might take some time, you can do it
yourself if you have time

sanjaybora04:
I am not familiar with nbdev, you do it please

PR #3868, opened 2023-02-01
Allows user to provide the start/stop epoch in `SaveModelCallback`

kunaltyagi:
Allows different saving frequencies in different epoch range, eg: every 10 epochs for first 100, then every 5 epochs till the end


jph00:
Looks like there's a CI failure - do you want to take a look?

kunaltyagi:
I'll take a look. Maybe i had some uncommited code so didn't see this failure locally

PS: the output on running locally doesn't match the curent cached outputs on unchanged cells. Is something an issue with my local setup?

PR #3871, opened 2023-02-12
Change remaining np.int types to int

warner-benjamin:
Completes #3850 by changing the remaining `np.int` types to `int`. The current version of NumPy (1.24) requires this change.


jph00:
Many thanks.

PR #3872, opened 2023-02-12
Allow Pillow Image to be passed to PILBase.create

warner-benjamin:
This PR modifies `PILBase` to accept a Pillow `Image` in `create`, which makes using `ImageBlock` more intuitive at both training and inference.

I also added more doc strings to PILBase derived classes.


jph00:
Great!

PR #3873, opened 2023-02-13
Set DataLoaders device if not None and to exists

warner-benjamin:
This PR changes `DataLoaders.__init__` to set device only if `to` exists and a device is passed on init. Currently if `to` exists and no device is passed, init will overwrite any existing DataLoader device with None.

I left the device.setter alone but could change check if `to` exists in the DataLoader and warn if there isn't one. Which could lead to multiple warnings, one per DataLoader. Currently setting DataLoaders device will error out if a DataLoader doesn't have a `to` method.


jph00:
Hmmm yes the previous logic does seem a bit odd...

PR #3875, opened 2023-02-14
Add support for a batch transforms `to` method

warner-benjamin:
This PR modifies `TfmdDL.to` to support calling a `to` method on batch transforms as an alternate means of moving the transform to a device. The current approach of looking for PyTorch Parameters and calling `to` doesn't work for moving `nn.Module` based transforms such as Torchaudio's [AmplitudeToDB](https://pytorch.org/audio/stable/generated/torchaudio.transforms.AmplitudeToDB.html).

### DIFF
--- a/fastai/data/core.py
+++ b/fastai/data/core.py
@@ -179,7 +179,8 @@ class TfmdDL(DataLoader):
     ):
         self.device = device
         for tfm in self.after_batch.fs:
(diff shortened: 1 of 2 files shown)


warner-benjamin:
I missed that TabularPandas and tabular Batch Transforms set `to` as an object, so my followup commit should be compatible with TabularPandas.

jph00:
Good idea :)

PR #3876, opened 2023-02-14
ChannelsLast Callback Improvements, Additional Documentation, & Bug Fix

warner-benjamin:
This PR upstreams my [ChannelsLast callback](https://fastxtend.benjaminwarner.dev/callback.channelslast.html) from fastxtend. This version of the callback is simpler, and it fixes the current `ChannelsLast` callback bug of adding the labels to the model input batch.

The new Channels Last notebook adds two convenience methods for enabling and disabling channels last (`to_channelslast` and `to_contiguous`), adding `MixedPrecision` by default for best channels last performance. It also expands the channels last documentation and adds a test to make sure model outputs are in channels last format.

Keeping with the current `MixedPrecision` behavior, I modified `load_learner` to remove the `ChannelsLast` and `MixedPrecison` callbacks.

I also moved the Mixed Precision and Channels Last notebooks under the Training sidebar header for easier visibility, since they struck me as more similar to the Distributed Trainer notebook, which is directly under Training, than other callbacks.


jph00:
Many thanks!

PR #3881, opened 2023-02-22
Fixed CONTRIBUTING.md Links & Commands

nglillywhite:
- Both 'nbdev_install_hooks' & 'nbdev_test' were outdated

I also think

` or `make test`` and ` or `make fastai``

aren't valid anymore since there's no makefile, should this also be deleted? I'll update the PR if so, just asking for a sanity check.

I also noticed the "https://nbdev.fast.ai/sync#nbdev_diff_nbs" link is dead and I can't see an obviously named equivalent api as I think the [general approach to managing diffs has changed ](https://nbdev.fast.ai/blog/posts/2022-08-25-jupyter-git/#the-jupytergit-problem). Should this line be removed entirely and instead be directed to the [nbdev getting started page?](https://nbdev.fast.ai/getting_started.html)?


jph00:
Many thanks. You're right there's no Makefile any more.

jph00:
Many thanks for these updates.

PR #3882, opened 2023-02-24
TensorBase deepcopy Compatibility

warner-benjamin:
This PR updates `TensorBase` to be compatible with current PyTorch `deepcopy` requirements. It should also be backwards compatible with at least PyTorch 1.9.

As of pytorch/pytorch#73078, all Tensor subclasses are required to override `new_empty` and `clone` for `deepcopy` to work. As `new_empty` is an overloaded method, I implemented both versions in `TensorBase`.

This PR does not fix deepcopy for a quantized `TensorBase`, as the error message as of PyTorch 1.13.1 reads:

> The default implementation of `__deepcopy__()` for quantized tensors expects the tensor returned by `torch._utils._rebuild_qtensor()` to match the type of the instance being copied. If you encounter this, please open an issue on PyTorch's GitHub

so I am leaving `TensorBase.is_quantized` compatibility for a future fastai PR.

This PR also adds a test for `TensorBase` deepcopying to catch future subclassed tensor copying changes.


jph00:
Many thanks @warner-benjamin  -- great job getting this sorted out.

Issue #3880, opened 2023-02-22
`Learn.predict` Errors Out if Passed a `PILImage`


warner-benjamin:
As of the fastai 2.7.11 release, there are reports on the fastai [discord](https://discord.com/channels/689892369998676007/748330524556132364/1075861042875334738) and [forums](https://forums.fast.ai/t/beginner-basics-of-fastai-pytorch-numpy-etc/96285/47) that the [Is it a bird? Creating a model from your own data](https://www.kaggle.com/code/jhoward/is-it-a-bird-creating-a-model-from-your-own-data) Kaggle notebook errors out on the prediction step.

Specifically, here, due to passing in a `PILImage` instead of a Pillow Image or file name:

```python
is_bird,_,probs = learn.predict(PILImage.create('bird.jpg'))
```

Rolling back to fastai 2.7.10 is supposed to [prevent the issue from occurring on Kaggle](https://forums.fast.ai/t/lesson-1-official-topic/95287/469), but when I created a new conda env locally with 2.7.10 I still ran into the error. It's possible that this error occurred due to #3872, or maybe a difference in Pillow versions as I think `PILBase.create` should behave the same in 2.7.10 and 2.7.11 if passed a `PILImage`.

There is a PR to fix the course notebook fastai/course22#52, but I think `learn.predict` should accept a `PILBase` instance, and this issue should be fixed in fastai.

nglillywhite:
Happy to take this one up, will checkout the links and suggest a path forward 😀

nglillywhite:
Hey @warner-benjamin,

I've managed to recreate the error, 2.7.10 works, and 2.7.11 creates this `AttributeError: read`

I'll dig into vision_core and make the changes and tests to cover this.

Is it appropriate to create a predict() test within the vision_learner class and replicate this circumstance despite changes being within the vision_core code? Seems like a cheap way to check/cover a vision learner taking in multiple types but I also don't want to break testing conventions if this isn't the fastai way to test and I should only be testing within the module I've made changes.

warner-benjamin:
@nglillywhite [`vision_learner`](https://github.com/fastai/fastai/blob/master/fastai/vision/learner.py#L209) isn't a class, but rather a method which returns a `Learner` instance.

Assuming the issue is with vision_core, then it would probably suffice to create a test like this current one, except passing a `PILImage` to `PILImage.create`

```python
timg = TensorImage(image2tensor(im))
tpil = PILImage.create(timg)
```
[...]

PR #3884, opened 2023-02-27
Issue #3880 Possible Fix

nglillywhite:
The `self.types` loop in [`infer_idx`](https://github.com/fastai/fastai/blob/4d1834cb0b6ac20b068de55cf57f40a0c2296cd4/fastai/data/core.py#L400-L409) hits the break on line [404](https://github.com/fastai/fastai/blob/4d1834cb0b6ac20b068de55cf57f40a0c2296cd4/fastai/data/core.py#L404) one element earlier than the commit prior to https://github.com/fastai/fastai/pull/3872 and the idx gets set to one value lower. This means that the [rm_tfms](https://github.com/fastai/fastai/blob/4d1834cb0b6ac20b068de55cf57f40a0c2296cd4/fastai/data/core.py#L514) variable within https://github.com/fastai/fastai/blob/4d1834cb0b6ac20b068de55cf57f40a0c2296cd4/fastai/data/core.py#L504-L523 contains the [create method](https://github.com/fastai/fastai/blob/4d1834cb0b6ac20b068de55cf57f40a0c2296cd4/fastai/vision/core.py#L117-L125) of [PILBase](https://github.com/fastai/fastai/blob/4d1834cb0b6ac20b068de55cf57f40a0c2296cd4/fastai/vision/core.py#L111) when it previously didn't.

Making the [create method](https://github.com/fastai/fastai/blob/4d1834cb0b6ac20b068de55cf57f40a0c2296cd4/fastai/vision/core.py#L117-L125) of [PILBase](https://github.com/fastai/fastai/blob/4d1834cb0b6ac20b068de55cf57f40a0c2296cd4/fastai/vision/core.py#L111) method safely accept a PILImage means this `rm_tfms` variable that gets set to the `test_tl.tfms.fs` attribute is executed successfully when [compose_tfms ](https://github.com/fastai/fastcore/blob/f7fea257626106e2016d4a55d280f8b876f6dcb4/fastcore/transform.py#L153-L159) is called.

Whilst I think its good that a `PILImage` can be created from another PILImage, this fix doesn't preserve the original behavior of how [`infer_idx`](https://github.com/fastai/fastai/blob/4d1834cb0b6ac20b068de55cf57f40a0c2296cd4/fastai/data/core.py#L400-L409) previously functioned. **_Not sure if this is a bigger problem or if change this is sufficient_**

### DIFF
--- a/fastai/vision/core.py
+++ b/fastai/vision/core.py
@@ -121,7 +121,7 @@ class PILBase(Image.Image, metaclass=BypassNewMeta):
         if isinstance(fn,Tensor): fn = fn.numpy()
         if isinstance(fn,ndarray): return cls(Image.fromarray(fn))
         if isinstance(fn,bytes): fn = io.BytesIO(fn)
-        if isinstance(fn,Image.Image) and not isinstance(fn,cls): return cls(fn)
+        if isinstance(fn,Image.Image): return cls(fn)
         return cls(load_image(fn, **merge(cls._open_args, kwargs)))
 
     def show(self, ctx=None, **kwargs):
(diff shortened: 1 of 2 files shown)


warner-benjamin:
I think this simple fix is the correct fix, as modifying [`infer_idx`](https://github.com/fastai/fastai/blob/4d1834cb0b6ac20b068de55cf57f40a0c2296cd4/fastai/data/core.py#L400-L409) and related code to perform a strict type check would probably cause more issues in the future.

We want to merge and release this soon @jph00, as there are a lot of questions as to why the tutorial notebooks on Kaggle are breaking and this will fix those issues.

jph00:
Many thanks @nglillywhite and @warner-benjamin

PR #3885, opened 2023-02-27
TypeError: new_empty(): argument 'layout' must be torch.layout, not NoneType

csaroff:
This resolves an issue where calling `TensorBase.new_empty` with the default parameters results in an exception.

[Tensor.new_empty](https://pytorch.org/docs/stable/generated/torch.Tensor.new_empty.html) uses `layout=torch.strided` by default, so I replicated those arguments here.

```python
│ /home/jovyan/fastai/fastai/vision/augment.py:573 in _draw_mask                                   │
│                                                                                                  │
│    570 │   │   assert len(draw)>=x.size(0)                                                       │
│    571 │   │   res = tensor(draw[:x.size(0)], dtype=x.dtype, device=x.device)                    │
│    572 │   else: res = x.new_zeros(x.size(0)) + draw                                             │
│ ❱  573 │   return TensorBase(mask_tensor(res, p=p, neutral=neutral, batch=batch))                │
│    574                                                                                           │
│    575 # %% ../../nbs/09_vision.augment.ipynb 126                                                │
│    576 def affine_mat(*ms):                                                                      │
│                                                                                                  │
│ /home/jovyan/fastai/fastai/vision/augment.py:560 in mask_tensor                                  │
│                                                                                                  │
│    557 │   if p==1.: return x                                                                    │
│    558 │   if batch: return x if random.random() < p else x.new_zeros(*x.size()) + neutral       │
│    559 │   if neutral != 0: x.add_(-neutral)                                                     │
│ ❱  560 │   mask = x.new_empty(*x.size()).bernoulli_(p)                                           │
│    561 │   x.mul_(mask)                                                                          │
│    562 │   return x.add_(neutral) if neutral != 0 else x                                         │
│    563                                                                                           │
│                                                                                                  │
│ /home/jovyan/fastai/fastai/torch_core.py:406 in new_empty                                        │
│                                                                                                  │
│   403 │                                                                                          │
│   404 │   def new_empty(self, *size, dtype=None, layout=None, device=None, pin_memory=False, r   │
│   405 │   │   cls = type(self)                                                                   │
│ ❱ 406 │   │   return self.as_subclass(Tensor).new_empty(*size, dtype=dtype, layout=layout, dev   │
│   407                                                                                            │
│   408 # %% ../nbs/00_torch_core.ipynb 105                                                        │
│   409 class TensorImageBase(TensorBase):                                                         │
╰──────────────────────────────────────────────────────────────────────────────────────────────────╯
[...]

PR #3887, opened 2023-03-01
Better version support for `TensorBase.new_empty`

warner-benjamin:
This PR updates #3882 to support PyTorch 1.12, as @csaroff pointed out it failed in #3885.

The pre-parsed PyTorch versions decrease the speed hit of `TensorBase.new_empty` from ~18.75x slower than `Tensor.new_empty` when parsing in `new_empty` to ~4.4x slower.

I tested this PR on PyTorch 1.13, 1.12, 1.11. 1.10, and 1.9 using the following script:

```python
from copy import deepcopy

t = TensorBase([1,2,3,4])
x = deepcopy(t)

imagenette = untar_data(URLs.IMAGENETTE_160)

block = DataBlock(
    blocks=(ImageBlock, CategoryBlock),
    splitter=GrandparentSplitter(valid_name='val'),
    get_items=get_image_files, get_y=parent_label,
    item_tfms=Resize(128),
    batch_tfms=[*aug_transforms(flip_vert=True, max_rotate=45, max_warp=0.3, xtra_tfms=[Hue(), Saturation()]),
                Normalize.from_stats(*imagenet_stats)],
)
dls = block.dataloaders(imagenette, bs=128, num_workers=num_cpus(), pin_memory=True)

learn = Learner(dls, xresnext18(n_out=dls.c)).to_channelslast()
learn.fit_one_cycle(1, 3e-4)
```

### DIFF
--- a/fastai/torch_core.py
+++ b/fastai/torch_core.py
@@ -307,6 +307,12 @@ def to_concat(xs, dim=0):
                           for i in range_of(o_)) for o_ in xs], L())
 
 # %% ../nbs/00_torch_core.ipynb 84
+# Parsed PyTorch versions for faster version checking
+_torch_version = parse(torch.__version__)
+_torch_113 = parse('1.13')
+_torch_112 = parse('1.12')
+
+# %% ../nbs/00_torch_core.ipynb 85
 @patch
 def set_meta(self:Tensor, x, as_copy=False):
     "Set all metadata in `__dict__`"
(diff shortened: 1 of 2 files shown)


jph00:
Thankyou :)

csaroff:
Thanks for the fix @warner-benjamin!

Probably not worth updating again either way, but just curious if this could have been resolved better by passing in kwargs to new_empty which would have allowed skipping the version check and hardcoding strided?

warner-benjamin:
@csaroff `new_empty` is regularly called by PyTorch without setting `layout`, so `TensorBase.new_empty` needs to distinguish between 1.13 & 2.0 calling it expecting a `layout` default of `None` and 1.12 and earlier which expect a default of `strided`. Some form of PyTorch version hardcoding is required.

PR #3889, opened 2023-03-11
Adding clarity on config.ini's directory in a docstring. 

nesaboz:
I wanted to change the folder where datasets are being downloaded using `untar_data`, this can be done by changing env. variable FASTAI_HOME (`~/.fastai` by default). This is useful to know since on remote machines `~/.fastai` gets deleted with every restart, hence the need to download data repeatedly. So I'm adding clarity to docstrings. No code changes.


jph00:
Many thanks.

PR #3890, opened 2023-03-15
PyTorch 2.0 Compatibility

warner-benjamin:
This PR updates fastai to be compatible with PyTorch 2.0.


warner-benjamin:
I changed `TensorBase` to use `Tensor`'s `__reduce_ex__` and didn't notice any errors when running tests with the slow and cuda flags turned on. An Imagenette training and augmentation test also appears to work correctly.

I'm not sure why we have our own `__reduce_ex__` implementation for `TensorBase`, so I left it alone for PyTorch 1.x versions.

jph00:
Awesome job @warner-benjamin

PR #3895, opened 2023-03-21
fix callout renders

deven-gqc:
This PR fixes a few callout-note rendering issues in the docs.


jph00:
Many thanks for these fixes @deven-gqc

PR #3899, opened 2023-03-27
Pytorch 2.0 compiler

ggosline:
Pytorch 2.0 compiled models fail when used with fastai.  Two fixes are necessary to run a basic Convnext model.
fastcore L.__eq__ requires a chack for a function being passed in, and the Flatten layer has to have the Tensorbase type cast removed.  
After this a simple model=torch.compile(model) increases training throughput (it/s) by 50%+.
The compiled models can't have layers frozen/unfrozen, so the train head / train all strategy is not useable, but didn't make much difference in my limited tests.


jph00:
Many thanks for this - and congrats on figuring it out!

Issue #3901, opened 2023-04-10
Some fastai Loss Functions Cannot be Exported due to Pickling Error


warner-benjamin:
Some fastai loss functions, such as `LabelSmoothingCrossEntropyFlat` are not picklable, and thus not exportable via `Learner.export`. But most, including `LabelSmoothingCrossEntropy` and other flattened losses, appear pickle without issue.

This minimal reproduction on Python 3.10 with the `Learner.export` default `protocol=2`
```python
try:
    pickle.dumps(LabelSmoothingCrossEntropyFlat(), protocol=2)
    print("The class is picklable")
except pickle.PicklingError as e:
    print(e)
```
results in `TypeError: cannot pickle 'code' object` for `LabelSmoothingCrossEntropyFlat` but no errors for `LabelSmoothingCrossEntropy` or `CrossEntropyLossFlat`.

We should probably add a test for pickling `LabelSmoothingCrossEntropyFlat` to catch future changes to pickle.

nglillywhite:
Hey @warner-benjamin, let me know if this still needs picking up, I'll put in a PR if so.

warner-benjamin:
@nglillywhite I am unaware of this being worked on

nglillywhite:
Great, I'll have a look and put in a PR, thanks

nglillywhite:
[PR 3910](https://github.com/fastai/fastai/pull/3910) adds these guards for current loss functions, I'd like to investigate why these loss functions can't be pickled with this issue but happy to open a new issue & PRs if you'd like to keep them separate.

dharik-arsath:
Hi, I am facing a same issue on Focal Loss, unable to pickle focal loss. am I missing out something, is there any solution to this problem.

jordiju:
Bump. Same issue here.

kevin-vitro:
It seems like the issue here is somewhere in `Module`. That is why some of the loss functions work and others don't.

I dug into this a bit and got to the following test: 
```
class FocalLossTest(Module):
    def __init__(self):
        pass

    def forward(self, inp:Tensor, targ:Tensor) -> Tensor:
        return 0

#Works
pickle.dumps(FocalLossTest())

#Works
BaseLoss(FocalLossTest)

# Fails
pickle.dumps(BaseLoss(FocalLossTest))

# Fails
pickle.dumps(FocalLossTest().__annotations__)

# Fails
pickle.dumps(FocalLossTest().__annotations__['_modules'])
```

[...]

PR #3910, opened 2023-05-03
#3901 - Added Guards for Pickling Losses

nglillywhite:
Started with guards for current loss functions & python 3.10.8. Will investigate why these losses can't be pickled in further commits


nglillywhite:
Relates to Issue #3901

Issue #3911, opened 2023-05-05
fastai.vision.augment.mask_tensor() crashes runtime for 'mps' accelerator (apple silicon chip)


MSciesiek:
Please confirm you have the latest versions of fastai, fastcore, and nbdev prior to reporting a bug (delete one): YES

Problem is visible in fastai==2.7.12, fastcore==1.5.29 and torch==2.0.0

**Describe the bug**
This line crushes the python runtime on MacOS with M1 apple silicon chips for the 'mps' accelerator https://github.com/fastai/fastai/blob/master/fastai/vision/augment.py#L560
```mask = x.new_empty(*x.size()).bernoulli_(p)```
It seems that the type has to be cast implicitly to avoid the error.

**To Reproduce**
on a macbook with M1 chip run:
```python -c "import torch; x=torch.zeros([1], dtype=torch.int64, device=torch.device('cpu')).bernoulli_(0.5);print(x))"```

**Expected behavior**
Augmentations running without crashing the runtime

**Error with full stack trace**

```
loc("mps_less"("(mpsFileLoc): /AppleInternal/Library/BuildRoots/97f6331a-ba75-11ed-a4bc-863efbbaf80d/Library/Caches/com.apple.xbs/Sources/MetalPerformanceShadersGraph/mpsgraph/MetalPerformanceShadersGraph/Core/Files/MPSGraphUtilities.mm":248:0)): error: input types 'tensor<1xf32>' and 'tensor<1xsi64>' are not broadcast compatible
LLVM ERROR: Failed to infer result type(s).
```

**Additional context**

this works:
```python -c "import torch; x=torch.zeros([1], dtype=torch.int64, device=torch.device('cpu')).bernoulli_(0.5);print(x))"```
this works:
```python -c "import torch; x=torch.zeros([1], dtype=torch.int64, device=torch.device('cpu')).float().bernoulli_(0.5).long();print(x))"```
this works:
```python -c "import torch; x=torch.zeros([1], dtype=torch.int64, device=torch.device('mps:0')).float().bernoulli_(0.5).long();print(x))"```
this does not work:
```python -c "import torch; x=torch.zeros([1], dtype=torch.int64, device=torch.device('mps:0')).bernoulli_(0.5);print(x))"```

MSciesiek:
[...]

PR #3912, opened 2023-05-05
Fix casting types for mps

MSciesiek:
Fix for the issue https://github.com/fastai/fastai/issues/3911


jph00:
Thanks for the contribution! This is an nbdev project, so the actual source is in the notebooks, and the .py files are auto-generated. Could you please make your edit in the notebook too? Also, please add a comment above explaining why the extra casts are there?

MSciesiek:
Sure, I will do a proper local setup and will do then the requested changes. Most probably over the coming weekend.

jph00:
@MSciesiek just following up on this

jph00:
Thank you!

PR #3917, opened 2023-05-25
Fixed link to layers image on README

bkowshik:
Fixed link to layers image on README.


jph00:
That would break the notebook unfortunately. I've copied the image over now.

PR #3926, opened 2023-07-10
Update learner.py - fixed type in warning

astewartau:
Fixed a typo in a UserWarning - "Saved filed" changed to "Saved file"


jph00:
This needs to be changed in the notebook too - this is an nbdev project so the notebooks are the actual source.

jph00:
Many thanks!

PR #3934, opened 2023-08-05
Fixed format string, Closes #3920

### LINKED ISSUE #3920: SegmentationInterpretation.plot_top_losses() gives "unsupported format string" error
```
fastai==2.7.12
fastcore==1.5.29
fastdownload==0.0.7
fastprogress==1.0.3
torch==2.0.1
torchvision==0.15.2
```
**Describe the bug**

Working through the FastAI Segmentation tutorial at https://docs.fast.ai/tutorial.vision.html#segmentation, everything is fine until the cell:
```
interp = SegmentationInterpretation.from_learner(learn)
interp.plot_top_losses(k=3)
```
which gives:
```
TypeError: unsupported format string passed to TensorBase.__format__
```
from the `interp.plot_top_losses()` line.

The first "input" and "target" images appear, but what should be the "pred" image and all subsequent images do not.

**To Reproduce**
1. Install latest versions of PyTorch and FastAI
2. Run through the above tutorial.

**Expected behavior**

Not an error

**Additional context**

[This Stack Overflow answer](https://stackoverflow.com/questions/69021335/typeerror-unsupported-format-string-passed-to-tensor-format#answer-69021370) suggests that `torch.Tensor` no longer supports format operations and gives the examples:
```
print(f"{torch.rand(1):.2f}")
```
which gives the same error as above:
```
TypeError: unsupported format string passed to Tensor.__format__
```
and 
```
print(f"{torch.rand(1).item():.2f}")
```
which works as expected.

A quick look through the code suggests that the issue may lie in https://github.com/fastai/fastai/blob/master/nbs/21_vision.learner.ipynb in the function `plot_top_losses()`, in the line:
```
if title=="pred": title += f"; loss = {l:.4f}"
```
where `l` appears to be of type `TensorBase`.

The fix may just be to replace `l` with `l.item()`.

Presumably this was fine with earlier versions of PyTorch which supported format strings.

If this does prove to be the issue, there may be other instances of format strings being applied to tensors in the codebase which would be worth flushing out.

bkowshik:
Closes https://github.com/fastai/fastai/issues/3920

Had this error today and went ahead to create a PR with the fix suggested by @shrdluk 

```python
interp = SegmentationInterpretation.from_learner(learn)
interp.plot_top_losses(k=2)
```

## After the fix

image-segmentation

---

### DIFF
--- a/fastai/vision/learner.py
+++ b/fastai/vision/learner.py
@@ -354,6 +354,6 @@ def plot_top_losses(x:TensorImage, y:TensorMask, samples, outs, raws, losses, nr
     for axs,s,o,l in zip(axes, samples, outs, losses):
         imgs = (s[0], s[1], o[0])
         for ax,im,title in zip(axs, imgs, titles):
-            if title=="pred": title += f"; loss = {l:.4f}"
+            if title=="pred": title += f"; loss = {l.item():.4f}"
             im.show(ctx=ax, **kwargs)
             ax.set_title(title)
--- a/nbs/21_vision.learner.ipynb
+++ b/nbs/21_vision.learner.ipynb
@@ -1051,7 +1051,7 @@
     "    for axs,s,o,l in zip(axes, samples, outs, losses):\n",
     "        imgs = (s[0], s[1], o[0])\n",
     "        for ax,im,title in zip(axs, imgs, titles):\n",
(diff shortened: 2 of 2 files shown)


jph00:
Thanks! Can you please sync your notebooks using nbdev!

jph00:
Excellent!

PR #3936, opened 2023-08-07
Minimum Requirement of Python 3.8 & PyTorch 1.10

warner-benjamin:
This PR updates the fastai to require a minimum of Python 3.8 and PyTorch 1.10 as Python 3.7 is EOL'd and PyTorch 1.9 is over two years old (this matches Accelerate's minimum requirements). It also increases dependency versions to similar requirements. Updated dependencies for the default fastai install are all part of the current Colab & Kaggle images.

Code and checks for PyTorch<1.10 have been removed.

I elected to keep the TorchScript versions of Mish and SiLU (Swish) in case they are still being used, but renamed them to `MishJit` and `SwishJit`. `Mish` and `Swish` will use the much faster native PyTorch implementations.

The CI has been updated to test on Python 3.8, 3.9, & 3.10. PyTorch also supports 3.11 from pip install, it could be added to the test matrix if wanted in this or a future PR.


warner-benjamin:
I ran a local test with slow and cuda flags. I am unable to test multicuda, so assuming the Accelerate bump works fine since there was no maximum version cap.

jph00:
Thank you!

PR #3942, opened 2023-08-17
Fix the Deploy to GitHub Pages Action

warner-benjamin:
This PR attempts to fix the Deploy to GitHub Pages action by adding ipykernel to the dev-requirements since it is no longer being installed from other packages' requirements.


jph00:
Great- thank you for fixing that.

PR #3943, opened 2023-08-17
Slim & Speedup fastai Test Action

warner-benjamin:
This PR updates the fastai CI to cache pip installs and download PyTorch-CPU instead of PyTorch-Cuda. Removing Cuda dependencies will decrease the cache size by ~2GB.

Test run is working [here](https://github.com/warner-benjamin/fastai/actions/runs/5894438987).


warner-benjamin:
Requesting a review @jph00

jph00:
Smart idea! :D

PR #3944, opened 2023-08-17
Support TorchVision's Multi-Weight API

warner-benjamin:
This PR updates `vision_learner` and `unet_learner` to support TorchVision's multi-weight API using the new `weights` argument. The default of `pretrained=True, weights=None` will download the DEFAULT weights from TorchVision, which at the time of this PR are IMAGENET1K_V2.

To randomly initialize a model's weights, users will need to set both `pretrained=False, weights=None`.

If using pre TorchVision 0.13 or timm models setting `weights` will have no effect.

### DIFF
--- a/fastai/vision/learner.py
+++ b/fastai/vision/learner.py
@@ -2,12 +2,15 @@
 
 # %% ../../nbs/21_vision.learner.ipynb 1
 from __future__ import annotations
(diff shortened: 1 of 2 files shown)


jph00:
This is great! :D

PR #3948, opened 2023-08-28
fix torch.jit.script on TimmBody

johan12345:
Creating a model based on a `timm` backbone and then converting it to TorchScript currently fails.

```python
import torch
from fastai.vision.learner import create_timm_model

model = create_timm_model('resnet18', 1)
torch.jit.script(model)
``` 
```
RuntimeError: 
Could not cast value of type Tuple[int, int] to bool:
  File "/tmp/ipykernel_114724/2853927905.py", line 8
    def forward(self,x): return self.model.forward_features(x) if self.needs_pool else self.model(x)
                                                                  ~~~~~~~~~~~~~~~ <--- HERE
```

TorchScript can't implicitly convert the tuple into a boolean. So we should do it explicitly instead with `is not None`.


jph00 (changes requested):
Many thanks - this is great! Can you please remove the printing of the whole module in the notebook, since that will appear in the docs. Instead, just assert that it exists.

johan12345:
Oh of course, that makes sense. Just pushed the change.

jph00:
Perfect! :D

PR #3955, opened 2023-09-02
Make DataLoader.__init__() respect num_workers

### LINKED ISSUE #3954: DataLoader.__init__() ignores num_workers
as reported originally by **benkarr** on the [fast.ai forums](https://forums.fast.ai/t/how-to-set-num-workers-to-zero-for-fastai-learner-at-the-inference-stage/98760/2)

Please confirm you have the latest versions of fastai, fastcore, and nbdev prior to reporting a bug (delete one): OK
fastai==2.7.12 fastcore==1.5.29 nbdev==2.3.12

**Describe the bug**
`DataLoader.__init__()` ignores `num_workers`, [using 1 as default](https://github.com/fastai/fastai/blob/master/fastai/data/load.py#L103), contrary to [documentation](https://docs.fast.ai/data.load.html#dataloader).

**To Reproduce**
1. Create a dataloader `dls` in any way, e.g. using `DataBlock.dataloaders()`
2. check that `dls.num_workers` is always `1`

**Expected behavior**
`dls.num_workers` should respect the argument `num_workers`, including the default (`None` meaning that `ncpu` is used) and windows/jupyter/darwin quirks.

xalexalex:
fixes #3954 
furthermore, I moved the win32/darwin check earlier, where I think it should be to have any effect.

### DIFF
--- a/fastai/data/load.py
+++ b/fastai/data/load.py
@@ -100,9 +100,9 @@ class DataLoader(GetAttr):
             try: n = len(dataset)
             except TypeError: pass
         store_attr('dataset,bs,shuffle,drop_last,indexed,n,pin_memory,timeout,device')
-        self.rng,self.num_workers,self.offs = random.Random(random.randint(0,2**32-1)),1,0
         if sys.platform == "win32" and IN_NOTEBOOK and num_workers > 0: num_workers = 0       
         if sys.platform == "darwin" and num_workers > 0: num_workers = 0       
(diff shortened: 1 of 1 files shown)


jph00:
It is actually using num_workers, even although it isn't reporting it correctly. Nonetheless, I do agree it would be best if we reported it correctly!

Note that this is an nbdev project, so the change needs to be made in the source notebook, rather than just the py module.

PR #3960, opened 2023-09-18
Fix acl links

kevinunger:
Fix link that describes the BLEU metric used:
old: 
http://acl2014.org/acl2014/W14-33/pdf/W14-3346.pdf
new: 
https://aclanthology.org/W14-3346/


jph00:
Many thanks!

PR #3966, opened 2023-10-08
allow torch-2.1

stonebig:


jph00:
Have you tested it works correctly with cuda?

stonebig:
> Have you tested it works correctly with cuda?

no, I don't have nvidia card.

jph00:
We can't merge untested PRs, sorry. Once the new version is fully tested we can support it.

PR #3969, opened 2023-10-13
Add `MutableMapping` to `torch_core.apply` to Support Moving Transformers Dicts

warner-benjamin:
This PR adds `MutableMapping` to `torch_core.apply` to support moving Hugging Face Transformers dictionary objects with `to_device`. Transformers dictionaries inherit from `UserDict`, so the current `isinstance(x,dict)` check returns False and batches with `UserDict` won't be placed on the correct device with `to_device`.

### DIFF
--- a/fastai/imports.py
+++ b/fastai/imports.py
@@ -11,7 +11,7 @@ from copy import copy,deepcopy
 from multiprocessing import Lock,Process,Queue,queues
 from datetime import datetime
 from contextlib import redirect_stdout,contextmanager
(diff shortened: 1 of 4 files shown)


warner-benjamin:
Library is resynced in #3970.

jph00:
Can you please run sync in this one so the tests will run?

warner-benjamin:
This should work now that #3973 was merged, fixing the bug which stopped the CI.

PR #3975, opened 2023-10-22
Fix errors in distributed notebook example

eihli:
`distrib_ctx` is defined in `fastai.distributed.py`, there is no "accelerate" module.

Also, something in the data loader code initializes the CUDA device and if you try to run the `notebook_launcher` afterwards you get an error that "CUDA has been initialized before the `notebook_launcher` could create a forked subprocess".

The [documentation](https://huggingface.co/docs/accelerate/basic_tutorials/notebook#using-the-notebooklauncher) from 🤗 says:

	Next, you should make a get_dataloaders function that will
	return your built dataloaders for you. As mentioned earlier, if
	data is automatically sent to the GPU or a TPU device when
	building your DataLoaders, they must be built using this method.

And their example has the dataloader function called inside the training loop function that gets passed to `notebook_launcher`:

	def training_loop(mixed_precision="fp16", seed: int = 42, batch_size: int = 64):
	    set_seed(seed)
	    # Initialize accelerator
	    accelerator = Accelerator(mixed_precision=mixed_precision)
	    # Build dataloaders
	    train_dataloader, eval_dataloader = get_dataloaders(batch_size)
	    # ...

image

Making the changes above revealed another issue.

```
RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cuda:2!
```

You see that error because the batches that get sent to [encodes](https://github.com/fastai/fastai/blob/master/fastai/data/transforms.py#L377) are distributed across devices but the Transformer's `self.mean` and `self.std` both got set in the `rank0_first` process.

image

By setting each DataLoader's device after distributing the dataloaders across devices with the call to `accelerator.prepare`, the device's transforms get placed on the proper device. Another option is to add a `device` kwarg to `DistributedDL` and pass the `self.learn.model.device` into `_wrap_dl` in the `before_batch` method and let `DistributedDL.__init__` do the work of setting its device.


nbs/20a_distributed.ipynb
```diff
@@ -642,18 +642,17 @@
...
     "\n",
     "def train():\n",
-    "    with learn.accelerate_ctx(in_notebook=True):\n",
+    "    set_seed(99, True)\n",
+    "    path = untar_data(URLs.PETS)/'images'\n",
+    "    dls = ImageDataLoaders.from_name_func(\n",
+    "        path, get_image_files(path), valid_pct=0.2,\n",
+    "        label_func=lambda x: x[0].isupper(), item_tfms=Resize(224))\n",
+    "    \n",
+    "    learn = vision_learner(dls, resnet34, metrics=error_rate).to_fp16()\n",
```
eihli, line 654: If you don't wrap this line in a [rank0_first](https://docs.fast.ai/distributed.html#rank0_first), then you'll spawn 3 downloads of the model.

But if you _just_ wrap this line in a `rank0_first`, like this `learn = rank0_first(lambda: vision_learner(dls, resnet101, metrics=error_rate).to_fp16())`, then you'll get an error:

> RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cuda:1!

So something else needs to be done. I haven't tracked it down yet.

eihli: Ok. I've tracked it down to [here](https://github.com/fastai/fastai/blob/master/fastai/data/transforms.py#L372).

```
class Normalize(DisplayedTransform):
    "Normalize/denorm batch of `TensorImage`"
    parameters,order = L('mean', 'std'),99
    def __init__(self, mean=None, std=None, axes=(0,2,3)): store_attr()

    @classmethod
    def from_stats(cls, mean, std, dim=1, ndim=4, cuda=True): return cls(*broadcast_vec(dim, ndim, mean, std, cuda=cuda))

    def setups(self, dl:DataLoader):
        if self.mean is None or self.std is None:
            x,*_ = dl.one_batch()
            self.mean,self.std = x.mean(self.axes, keepdim=True),x.std(self.axes, keepdim=True)+1e-7

    def encodes(self, x:TensorImage): return (x-self.mean) / self.std
    def decodes(self, x:TensorImage):
        f = to_cpu if x.device.type=='cpu' else noop
        return (x*f(self.std) + f(self.mean))

    _docs=dict(encodes="Normalize batch", decodes="Denormalize batch")
```
That setup only happens once, so `self.mean` and `self.std` are going to be on whatever device was used during setup, but `x` is going to be on the current device.

I confirmed it by setting `self.mean` and `self.std` inside the `encodes` function. Now I'm trying to find a way to resolve this without having to recalculate the `mean` and `std` on every call to `encodes`.

eihli: It's like I'm looking for a way to say "Hey, if this Transform gets distributed, then run this callback."

I can't think of a good general-purpose way to handle this.

It's like there's no good way to do this without complicating DataLoaders/Transformers. I don't want a DataLoader to have to peak inside it's list of transformers and conditionally do stuff to specific ones in the case where the DataLoader gets distributed. And I don't want Transtormers to reach up to their DataLoader to tell them the same.

The way things are now, you just kind of have to "know" whether or not a Transformer is `distrib_ctx`-safe and handle it in a one-off way.

I'm not very familiar with this codebase though. I'd love to get the input of someone who is familiar with it.

eihli:
Per conversation on Discord: look at TfmdDL and see if it can/should be calling `to` on all of its Transforms.

warner-benjamin:
#3976 should fix the quoted issue from Discord.

> Per conversation on Discord: look at TfmdDL and see if it can/should be calling `to` on all of its Transforms.
> 
> > benjamin — Today at 1:23 PM
> > Maybe the distributed issue is related to different bug I encountered today. If you pass a device to a fastai DataLoader/TfmdDL on init, it doesn't move the transforms to that device.

warner-benjamin:
@eihli I'm going to respond to [your post in #3976](https://github.com/fastai/fastai/pull/3976#issuecomment-1778212214) here, so the discussion on this issue/PR doesn't end up in too many places.

> Still getting the error from #3975 on this branch. ("RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cuda:1!")
> 
> Here's a bit of info that I hope you can use to track down the issue. I'm happy to jump on a screenshare if you want to collaborate on my environment as it experiences the issues.
> 
> This first diff fixes the error. The second diff doesn't.
> 
> ```
> diff --git a/fastai/distributed.py b/fastai/distributed.py
> index b8d71083..8b537dd5 100644
> --- a/fastai/distributed.py
> +++ b/fastai/distributed.py
> @@ -87,10 +87,11 @@ class DistributedDL(TfmdDL):
>                  pin_memory=dl.pin_memory, timeout=dl.timeout, shuffle=shuffle, drop_last=dl.drop_last, persistent_workers=dl.persistent_workers)
>          self.bs,self.device,self.drop_last,self.dataset,fake,self.num_workers,self.offs,self.pin_memory = \
>              attrgetter('bs','device','drop_last','dataset','fake_l','num_workers','offs','pin_memory')(self.dl)
> +        self.dl.device = self.dl.device
>          self.fake_l = _FakeLoader(self, fake.pin_memory, fake.num_workers, fake.timeout, 
>                                    persistent_workers=fake.persistent_workers, 
> -                                  pin_memory_device=fake.pin_memory_device)
> -        
> +u                                 pin_memory_device=fake.pin_memory_device)
> +
>      def _broadcast(self,t,rank):
>          "Broadcasts t from rank `rank` to all other ranks. Returns t so t is same for all ranks after call."
>          t = LongTensor(t).cuda() # nccl only works with cuda tensors
> ```
> 
> ```
> diff --git a/fastai/distributed.py b/fastai/distributed.py
> index b8d71083..cf7fa24e 100644
> --- a/fastai/distributed.py
> +++ b/fastai/distributed.py
> @@ -83,7 +83,7 @@ class DistributedDL(TfmdDL):
>          store_attr()
>          if type(dl) == torch.utils.data.DataLoader:
>              shuffle = True if eq(type(dl.sampler), torch.utils.data.RandomSampler) else False
> -            self.dl = DataLoader(dataset=dl.dataset, bs=dl.batch_size, num_workers=dl.num_workers, \
> +            self.dl = DataLoader(device=dl.device,dataset=dl.dataset, bs=dl.batch_size, num_workers=dl.num_workers, \
>                  pin_memory=dl.pin_memory, timeout=dl.timeout, shuffle=shuffle, drop_last=dl.drop_last, persistent_workers=dl.persistent_workers)
>          self.bs,self.device,self.drop_last,self.dataset,fake,self.num_workers,self.offs,self.pin_memory = \
>              attrgetter('bs','device','drop_last','dataset','fake_l','num_workers','offs','pin_memory')(self.dl)
> ```
> 
> I found this surprising since the [DataLoader does set](https://github.com/fastai/fastai/blob/nil/fastai/data/load.py#L102) `device` in its init, so I expect it to be hitting the setter that you added, and expect that second diff to resolve the issue.

The solution you mention in the quoted post:
```python
self.dl.device = self.dl.device
```

differs from this PR's current change of:
```python
self.dl.to(self.device)
```
so I'm curious which one works. or do both? (I don't have access to a multi-GPU machine, so cannot check myself.)

Also, `dl` could be a PyTorch compatible DataLoader due to `if type(dl) == torch.utils.data.DataLoader`, (as an aside should be an `isinstance` check), so it's not guaranteed to have a `device` attribute. So, you'll need to check if `device` exists first before trying to grab it from `dl` if that's the current working solution.

eihli:
> so I'm curious which one works. or do both?

Both do.

> Also, dl could be a PyTorch compatible DataLoader...

If that's the case, stick with the `self.dl.to(...)` solution?

Also:

> > I found this surprising since the [DataLoader does set](https://github.com/fastai/fastai/blob/nil/fastai/data/load.py#L102) device in its init, so I expect it to be hitting the setter that you added, and expect that second diff to resolve the issue.

Is it worth digging into this? I still expect `self.dl = DataLoader(device=dl.device, ...` should have also fixed it and I'm still really curious as to why it didn't. I think that's the solution that feels the best (if it would work).

jph00:
Please let me know @eihli when it's ready for review and all seems to be working for you.

fastai/distributed.py
```diff
@@ -77,16 +77,16 @@ def _round_to_multiple(number,multiple): return int(math.ceil(number/multiple)*m
...
         if world_size is None: world_size=num_distrib()
         store_attr()
         if type(dl) == torch.utils.data.DataLoader:
             shuffle = True if eq(type(dl.sampler), torch.utils.data.RandomSampler) else False
             self.dl = DataLoader(dataset=dl.dataset, bs=dl.batch_size, num_workers=dl.num_workers, \
                 pin_memory=dl.pin_memory, timeout=dl.timeout, shuffle=shuffle, drop_last=dl.drop_last, persistent_workers=dl.persistent_workers)
-        self.bs,self.device,self.drop_last,self.dataset,fake,self.num_workers,self.offs,self.pin_memory = \
-            attrgetter('bs','device','drop_last','dataset','fake_l','num_workers','offs','pin_memory')(self.dl)
+        self.bs,self.drop_last,self.dataset,fake,self.num_workers,self.offs,self.pin_memory = \
+            attrgetter('bs','drop_last','dataset','fake_l','num_workers','offs','pin_memory')(self.dl)
```
eihli, line 89: I just realized that to make this a backwards-compatible and non-breaking change, we should continue to get `self.device` from `self.dl` in the case where device isn't passed in.

eihli: Fixed with the conditional check below.

jph00:
Great stuff - thanks!

PR #3976, opened 2023-10-24
Place Transforms on device on `DataLoader.__init__` and when calling `to`.

warner-benjamin:
This PR modifies the `DataLoader` to move transforms to the correct device on `DataLoader` init or when calling `DataLoader.to`.

I would have kept the device moving code on `TfmdDL`, but `Getattr` was not playing nicely with a device property & setter added to `TfmdDL`. Errors all over the place.

One note: `DisplayedTransform` don't always show the current device the transform is on, probably due to the `DisplayedTransform.name` using the cached `'__stored_args__'` instead of the current state.


eihli:
Here's a bit of info that I hope you can use to track down the issue. I'm happy to jump on a screenshare if you want to collaborate on my environment as it experiences the issues.

This first diff fixes the error ("RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cuda:1!")

The second diff doesn't fix the error.

```
diff --git a/fastai/distributed.py b/fastai/distributed.py
index b8d71083..8b537dd5 100644
--- a/fastai/distributed.py
+++ b/fastai/distributed.py
@@ -87,10 +87,11 @@ class DistributedDL(TfmdDL):
                 pin_memory=dl.pin_memory, timeout=dl.timeout, shuffle=shuffle, drop_last=dl.drop_last, persistent_workers=dl.persistent_workers)
         self.bs,self.device,self.drop_last,self.dataset,fake,self.num_workers,self.offs,self.pin_memory = \
             attrgetter('bs','device','drop_last','dataset','fake_l','num_workers','offs','pin_memory')(self.dl)
+        self.dl.device = self.dl.device
         self.fake_l = _FakeLoader(self, fake.pin_memory, fake.num_workers, fake.timeout, 
                                   persistent_workers=fake.persistent_workers, 
-                                  pin_memory_device=fake.pin_memory_device)
-        
+u                                 pin_memory_device=fake.pin_memory_device)
+
     def _broadcast(self,t,rank):
         "Broadcasts t from rank `rank` to all other ranks. Returns t so t is same for all ranks after call."
         t = LongTensor(t).cuda() # nccl only works with cuda tensors
```

```
diff --git a/fastai/distributed.py b/fastai/distributed.py
index b8d71083..cf7fa24e 100644
--- a/fastai/distributed.py
+++ b/fastai/distributed.py
@@ -83,7 +83,7 @@ class DistributedDL(TfmdDL):
         store_attr()
         if type(dl) == torch.utils.data.DataLoader:
             shuffle = True if eq(type(dl.sampler), torch.utils.data.RandomSampler) else False
-            self.dl = DataLoader(dataset=dl.dataset, bs=dl.batch_size, num_workers=dl.num_workers, \
+            self.dl = DataLoader(device=dl.device,dataset=dl.dataset, bs=dl.batch_size, num_workers=dl.num_workers, \
                 pin_memory=dl.pin_memory, timeout=dl.timeout, shuffle=shuffle, drop_last=dl.drop_last, persistent_workers=dl.persistent_workers)
         self.bs,self.device,self.drop_last,self.dataset,fake,self.num_workers,self.offs,self.pin_memory = \
             attrgetter('bs','device','drop_last','dataset','fake_l','num_workers','offs','pin_memory')(self.dl)
```

jph00:
Thank you!

PR #3980, opened 2023-11-22
Correct typo: worse to worst

sasilver75:
"And like before, we can check where the model did its `worse`:"

Should be

"And like before, we can check where the model did its `worst`:"`

Thanks for the course! :)


sasilver75:
As shit I included a little bit that I didn't meant to, sec

Issue #3981, opened 2023-11-24
The attirbute 'has_mps' is deprecated and should be replaced with the attribute "torch.backends.mps.is_built()"


TalosMann:
Please confirm you have the latest versions of fastai, fastcore, and nbdev prior to reporting a bug (delete one): YES 

Merely a deprecated attribute, it causes no actual issue to my knowledge but the warning sign popping up is annoying.

Error can be found at fastai/fastai/torch_core.py line:263

PR #3994, opened 2023-12-15
fixes fastai/fastai#3981 to stop deprecated warning messages

jkyamog:
I have used the current torch.backends.mps.is_built attribute to replace the older has_mps has been deprecated about 6 months ago on this [pytorch commit](https://github.com/pytorch/pytorch/commit/4cfa06f706811e2c3b70289b222461822537bd93)

I have tested on an M1 machine for the positive case should the mps.is_available (line above) return false.  I have also tested this on a x86/nvidia machine and it remove the warning signs.

I have attached here a screenshot on how it can look like on google collab when its not fixed.
Screenshot 2023-12-15 at 15 22 35


jph00:
Many thanks!

PR #4014, opened 2024-03-07
Remove use of undefined variable in comet.ml callback

CharlesGaydon:
Quick fix to instantiation error when using `CometCallback`.


CharlesGaydon:
Thank you, copy that. 
I'll keep this PR open for a week or two to give you feedback when I get around running my experiments!

Lothiraldan:
@CharlesGaydon Hi, did you had a chance to use our new integration? Is it working well for you?

PR #4022, opened 2024-04-15
Add quoting parameter to from_csv() for Text and Vision DataLoaders

NickOveracker:
I was running into errors trying to import a dataset, and the cause ended up being that it was inferring quoting in a dataset that did not use quoting. I have added a parameter to the from_csv() methods for TextDataLoaders and VisionDataLoaders in order to accommodate this situation.

### DIFF
--- a/fastai/text/data.py
+++ b/fastai/text/data.py
@@ -278,9 +278,9 @@ class TextDataLoaders(DataLoaders):
         return cls.from_dblock(dblock, df, path=path, seq_len=seq_len, **kwargs)
 
     @classmethod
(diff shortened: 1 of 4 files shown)


jph00:
Thanks @NickOveracker ! Note that this is an nbdev project, so the "true source" is the notebooks. Can you please update those? (Until you do our CI won't pass.)

NickOveracker:
> Thanks @NickOveracker ! Note that this is an nbdev project, so the "true source" is the notebooks. Can you please update those? (Until you do our CI won't pass.)

Sure, will do!

PR #4026, opened 2024-04-24
Support PyTorch 2.3

warner-benjamin:
It appears that no changes are needed for PyTorch 2.3, all tests pass as does an imagenette spot training run.


jph00:
Thanks @warner-benjamin

PR #4028, opened 2024-04-25
Fix: "cut" parameter not passed to vision_learner

adamamer20:


jph00:
Thanks! Note that this is an nbdev project so the source notebook needs to be edited -- that's the actual "source code" of the project.

jph00:
You'll need to use WSL, not native Windows.

PR #4034, opened 2024-06-24
Support for loss function pickling 

### LINKED ISSUE #3901: Some fastai Loss Functions Cannot be Exported due to Pickling Error

### LINKED ISSUE #3927: FocalLossFlat causing picking error upon model export
image

I was having issues with exporting my model due to a pickling error, then I saw issue #3901 and decided to check the loss function I was using: FocalLossFlat. Seems to not be pickleable. Any solutions?

kevin-vitro:
This PR closes #3901 by resolving type hints using `typing.get_type_hints`

The issue has to do with torch 1.13.x+.  I also added the tests from #3910 and ensured that all of them were pickle-able.  

Also closes #3927


jph00:
Thank you!

PR #4055, opened 2024-11-27
Fix FillMissing processor by removing inplace=True

nikhilmalkari8:
### Issue:
- The `FillMissing` processor in `tabular.core` raised a warning due to the deprecated `inplace=True` in Pandas.
- This caused compatibility issues with recent Pandas versions.

### Fix:
- Replaced `inplace=True` with reassignment in the `FillMissing` processor.
- Updated `FillMissing` to work with TabularPandas objects and tested the behavior.

### Test:
- Added a new test in `tests/test_tabular_core.py` to validate the `FillMissing` functionality.
- Verified that missing values are filled correctly in continuous columns.


jph00:
I do not see any mention in the pandas docs of a deprecation?

NB: this is an nbdev project, so the source files are the notebooks, and that's where tests go too.

Issue #3996, opened 2023-12-21
'_FakeLoader' object has no attribute 'pin_memory_device'


tiagu:
Hi

I am trying to run a dataloader  .show_batch() but getting an attribute error:

```
dls.show_batch()

File env/lib/python3.10/site-packages/fastai/data/core.py:100, in TfmdDL.show_batch(self, b, max_n, ctxs, show, unique, **kwargs)
     98     old_get_idxs = self.get_idxs
     99     self.get_idxs = lambda: Inf.zeros
--> 100 if b is None: b = self.one_batch()
    101 if not show: return self._pre_show_batch(b, max_n=max_n)
    102 show_batch(*self._pre_show_batch(b, max_n=max_n), ctxs=ctxs, max_n=max_n, **kwargs)

File env/lib/python3.10/site-packages/fastai/data/load.py:137, in DataLoader.one_batch(self)
    135 def one_batch(self):
    136     if self.n is not None and len(self)==0: raise ValueError(f'This DataLoader does not contain any batches')
--> 137     with self.fake_l.no_multiproc(): res = first(self)
    138     if hasattr(self, 'it'): delattr(self, 'it')
    139     return res

File env/lib/python3.10/site-packages/fastcore/basics.py:660, in first(x, f, negate, **kwargs)
    658 x = iter(x)
    659 if f: x = filter_ex(x, f=f, negate=negate, gen=True, **kwargs)
--> 660 return next(x, None)

File env/lib/python3.10/site-packages/fastai/data/load.py:101, in DataLoader.__iter__(self)
     99 self.before_iter()
    100 self.__idxs=self.get_idxs() # called in context of main process (not workers/subprocesses)
--> 101 for b in _loaders[self.fake_l.num_workers==0](self.fake_l):
    102     if self.device is not None: b = to_device(b, self.device)
    103     yield self.after_batch(b)

File env/lib/python3.10/site-packages/torch/utils/data/dataloader.py:659, in _SingleProcessDataLoaderIter.__init__(self, loader)
    658 def __init__(self, loader):
--> 659     super().__init__(loader)
    660     assert self._timeout == 0
    661     assert self._num_workers == 0

File env/lib/python3.10/site-packages/torch/utils/data/dataloader.py:587, in _BaseDataLoaderIter.__init__(self, loader)
    583 self._rank = rank
    584 # for other backends, pin_memory_device need to set. if not set
    585 # default behaviour is CUDA device. if pin_memory_device is selected
[...]

PR #4058, opened 2024-12-19
Updated install instructions for conda in readme.

rother:
The old instructions (`conda install -c fastai fastai`) installs version 2.1.10 of fast.ai instead of the latest one. To install the latest version, `conda install fastai::fastai `should be used.

Adding the PR, because this has been an issue for me on a couple of VM installs already and I was recently reminded of it. The old version will lead to issues with torch imports in xresnet.py and the pin_memory_device issue mentioned here: https://github.com/fastai/fastai/issues/3996 which could be closed after this is merged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ miniforge). First install PyTorch using the conda line shown
 [here](https://pytorch.org/get-started/locally/), and then run:
 
 ``` bash
-conda install -c fastai fastai
+conda install fastai::fastai
 ```
 
 To install with pip, use: `pip install fastai`.


jph00:
Many thanks. I wasn't aware that these syntaxes were different!

Since this is an nbdev project, the readme is autogenerated from index.ipynb, so could you please make the update there?

rother:
> Many thanks. I wasn't aware that these syntaxes were different!
> 
> Since this is an nbdev project, the readme is autogenerated from index.ipynb, so could you please make the update there?

I added a commit with the change in index.ipynb

PR #4061, opened 2025-01-07
get rid of inplace operation and warning

### LINKED ISSUE #4043: Is there plan to update FillMissing for fastai.tabular ?
Hi, I am new to fast.ai and I am playing around with `fastai.tabular` on Kaggle and got the below warning. I couldn't find any information on this on the forum so I wonder if there is a plan to fix it? If yes, could I try? 
Screenshot 2024-08-21 at 10 58 45 PM

gautam-e:
Fixes #4043

I tried to make the change in the notebook but ran into a Module Not Found error (get_embeddings) . I couldn't figure out how to fix that so here's the change directly in the py file.

### DIFF
--- a/fastai/tabular/core.py
+++ b/fastai/tabular/core.py
@@ -311,7 +311,6 @@ class FillMissing(TabularProc):
         for n in missing.any()[missing.any()].keys():
             assert n in self.na_dict, f"nan values in `{n}` but not in setup training set"
         for n in self.na_dict.keys():
-            #to[n].fillna(self.na_dict[n], inplace=True)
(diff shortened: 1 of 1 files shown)


bjarnehinkel:
I am currently trying to fix this issue as well. When I use the method you use in this pull request, it does get rid of the warning. But there seem to be some issues with the na_dict when fixing in this way. In the dataset I am using for validation (https://www.kaggle.com/c/house-prices-advanced-regression-techniques/data), missing values that should be getting filled with a median of 0 are being filled with 136 instead (This is referring to the MasVnrArea column, and taking into consideration only the rows that are in the TabularPandas training set when determining the median).

PR #4062, opened 2025-01-07
Fix ys in tabular

gautam-e:
When we have a traget of more than one column, `to.dataloaders()` throws an error (something is wrong with your data...).
This is because the multiple target columns are not handled correctly. This PR attempts to fix that.


gautam-e:
Their PR includes the change proposed in PR #4061. Hope that's OK.

gautam-e:
I believe I might've created a bug with something like `learn.predict(df.iloc[0])` so let me work on this a bit more.

PR #4073, opened 2025-02-26
Fix #4072 untar_data() ignores provided base

### LINKED ISSUE #4072: untar_data() ignores config base when provided and always uses default fastai_cfg()
Please confirm you have the latest versions of fastai, fastcore, and nbdev prior to reporting a bug (delete one): YES (fastai v2.7.18, fastcore v1.7.29)

**Describe the bug**
The `untar_data()` function takes an optional `base` parameter whose purpose is to specify the  "Directory containing config file and base of relative paths". However, the path is ignored and the default `fastai_cfg()` location is always used.

It looks like the `untar_data()` function [always passes](https://github.com/fastai/fastai/blob/b624c9f54a170e721e77be254e407be8624f294a/fastai/data/external.py#L137) `fastai_cfg()` and `base` to `FastDownload`. The result is that `FastDownload` [always uses](https://github.com/fastai/fastdownload/blob/9c729f96be9b930054dadfc1598964e61664f434/fastdownload/core.py#L70) `fastai_cfg()` and never creates a new config based upon the provided `base`. FastDownload only uses `base` to create a new config if no config is provided.

**To Reproduce**

1. Call `untar_data()` with a custom `base`: untar_data(URLs.PETS, base='/tmp/my_location')

**Expected behavior**
Data is downloaded and untarred to `/tmp/my_location`. Actually, it gets download to the default `~/.fastai` directory.

**Additional context**

A workaround is to set `FASTAI_HOME` envar to change the base, but this doesn't work well to programmatically change the base. Alternatively, don't use `untar_data()` and  use `FastDownload` directly:

```python
# This doesn't use /tmp/my_location as the base, it still uses ~/.fastai
# path = untar_data(URLs.PETS, base='/tmp/my_location')

# WORKAROUND: Use FastDownload directly. Pass cfg=None to FastDownload instead of fastai_cfg()
# so that FastDownload creates a new config based on the provided base.
d = FastDownload(cfg=None, module=fastai.data, base='/tmp/my_location')
path = d.get(URLs.PETS, extract_key='data')
```

jeraymond:
Updated untar_data() to pass a None config to FastDownload when an overridden base is provided. FastDownload _always_ uses a provided config object and ignores base if both are given.

### DIFF
--- a/fastai/data/external.py
+++ b/fastai/data/external.py
@@ -131,8 +131,14 @@ def untar_data(
     data:Path=None, # Optional override for `Config`'s `data` key
     c_key:str='data', # Key in `Config` where to extract file
     force_download:bool=False, # Setting to `True` will overwrite any existing copy of data
-    base:str='~/.fastai' # Directory containing config file and base of relative paths
+    base:str=None # Directory containing config file and base of relative paths
 ) -> Path: # Path to extracted file(s)
     "Download `url` using `FastDownload.get`"
-    d = FastDownload(fastai_cfg(), module=fastai.data, archive=archive, data=data, base=base)
+    cfg = None
+    if base is None:
+        cfg = fastai_cfg()
+        # A base must be provided as FastDownload initializes a Path with it even
+        # though the config provided is ultimately used instead.
+        base = '~/.fastai'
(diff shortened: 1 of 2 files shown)


jph00:
Thanks!

PR #4074, opened 2025-03-03
Use fasttransform

RensDimmendaal:
This PR uses [fasttransform](https://github.com/answerdotai/fasttransform) which is where fastcore's the dispatch and transform modules have been moved now they use plum-dispatch.

This is a draft PR as we still need to merge some other PR's, see: [release checklist](https://github.com/AnswerDotAI/fasttransform/issues/3).


nbs/07_vision.core.ipynb
RensDimmendaal, line 1: TODO: double check why fewer encodes/decodes methods.

Issue #4069, opened 2025-02-16
broken link report website


sensor111:
on the page https://docs.fast.ai/       link from:  See the fast.ai documentation on [Using Colab](https://course.fast.ai/start_colab) for more information.  generate 404

justindachille:
By searching the string "docs fast ai start colab" on Google, I found these two links:

https://course20.fast.ai/start_colab
https://course19.fast.ai/start_colab.html

I'm wondering if either of these pages is what was intended to be linked to? If so, I can swap it out.

PR #4077, opened 2025-03-10
Fix broken Colab link - Issue #4069

annakavosh:
Updated the Colab link in the README.


jph00:
Thanks. The readme is auto-generated from index.ipynb so it actually needs to be fixed there.

jph00:
This is great thanks.

PR #4095, opened 2025-05-23
update torch version to 2.7

tonyhoo:
This PR updates the PyTorch version constraint in settings.ini to allow compatibility with PyTorch 2.7.

## Changes
- Updated `pip_requirements` from `torch>=1.10,<2.7` to `torch>=1.10,<2.8`
- Updated `conda_requirements` from `pytorch>=1.10,<2.7` to `pytorch>=1.10,<2.8`

## Testing
I've tested this change locally with PyTorch 2.7.0 and confirmed basic functionality works:
- Model training with vision_learner
- DataLoaders functionality
- Basic operations with tensors

## Motivation
This change is needed to allow projects that depend on fastai (such as [AutoGluon](https://github.com/autogluon/autogluon)) to use the latest PyTorch versions without version conflicts.


jph00:
Thank you!

PR #4105, opened 2025-07-07
Fix: avoid pandas chained assignment warning in Tabular core.py

D2himself:
## Problem

This line in `fastai/tabular/core.py` triggered a warning when using newer versions of pandas:

```python
to[n].fillna(self.na_dict[n], inplace=True)

## Fix

```python
to[n] = to[n].fillna(self.na_dict[n])

@jeraymond @muellerzr

### DIFF
--- a/nbs/40_tabular.core.ipynb
+++ b/nbs/40_tabular.core.ipynb
@@ -2,7 +2,7 @@
  "cells": [
   {
    "cell_type": "code",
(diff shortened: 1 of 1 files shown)


jph00:
Thanks @D2himself . This is an nbdev project, so the source files are the notebooks -- please make the change there, then run `nbdev_prepare`.

PR #4120, opened 2025-08-19
set weights_only=False for load_model_text, fixes LRFinder / lr_find

austinvhuang:
PyTorch 2.6 set weights_only=True as a default:
https://docs.pytorch.org/docs/main/notes/serialization.html#torch-load-with-weights-only-true

This breaks fastai in a couple of places. Optimizer metadata is stored in an `L` type (see image):

image

 Users can somewhat work around this if they don't need to save optimizer state. However, for `LRFinder` and `lr_find`, `torch.save` and `torch.load` are invoked with the optimizer saved and consequently `torch.load` fails with the `weights_only=True` default.

This PR adds `weights_only=False` to the `load_model_text` invocation of `torch.load`, which fixes `LRFinder` and `lr_find`. `load_pretrain` and `load_encoder` in the module are also updated with  `weights_only=False`.


jph00:
Thanks!

PR #4124, opened 2025-08-31
Address deprecation warnings in fp16 Callback

FacuRoffet99:
Updates the `fp16` Callback to use the modern `torch.amp` API instead of the deprecated `torch.cuda.amp` functions. This resolves the following `FutureWarning` messages:

```
FutureWarning: `torch.cuda.amp.autocast(args...)` is deprecated. Please use `torch.amp.autocast('cuda', args...)` instead.
FutureWarning: `torch.cuda.amp.GradScaler(args...)` is deprecated. Please use `torch.amp.GradScaler('cuda', args...)` instead.
```

The deprecated `torch.cuda.amp.autocast` and `torch.cuda.amp.GradScaler` are replaced with `torch.amp.autocast('cuda', ...)` and `torch.amp.GradScaler('cuda', ...)` respectively.

### DIFF
--- a/fastai/callback/fp16.py
+++ b/fastai/callback/fp16.py
@@ -7,8 +7,8 @@ from __future__ import annotations
 from ..basics import *
 from .progress import *
 
-from torch.cuda.amp import GradScaler,autocast
-from torch.cuda.amp.grad_scaler import OptState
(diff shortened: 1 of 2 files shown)


jph00:
Thanks a lot! :D

PR #4134, opened 2025-11-24
use the torch way of setting default device

fzngagan:


fzngagan:
```python
def set_default_device(device):
    # to make sure we always return a torch.device object
    dev = torch.device(device)
    torch.set_default_device(dev)
    return dev

def default_device(use=-1):
    "Return or set default device; `use_cuda`: -1 - CUDA/mps if available; True - error if not available; False - CPU"
    if use == -1: use = defaults.use_cuda
    else: defaults.use_cuda=use
    if use is None:
        if torch.cuda.is_available() or _has_mps(): use = True
    if use:
        if torch.cuda.is_available():
            return set_default_device(torch.device(torch.cuda.current_device()))
        if _has_mps(): return set_default_device("mps")
    return set_default_device("cpu")
```

Just so that I don't lose the code while cleaning up the commits.

PR #4140, opened 2026-01-12
Fix broken link to paper in README

0xOllie:
The link to the fastai paper in the README is broken. I couldn't find a replacement on MDPI but there is a copy on arxiv.


jph00:
The current link works ok for me?...

PR #4147, opened 2026-04-28
Replace broken search_images_bing with SerpApi-backed implementation

sS373Nn:
The original search_images_bing used the Bing Search API which has since been shut down. This replaces it with two SerpApi-backed functions:
- search_images_serpapi(key, term, engine='bing_images', ...) — general function supporting bing_images for now, with future plans to expand to include google_images, yahoo_images and yandex_images
- search_images_bing(key, term, ...) — drop-in wrapper around the above for familiarity Both functions use EAFP-style error handling, warning the user if the API key is invalid or missing before re-raising the exception SerpApi key from https://serpapi.com.

@jph00


jph00:
Looks like you might have forgotten to run nbdev-export :)

sS373Nn:
Oh no! I very much forgot to run nbdev. Will fix that first thing this morning and send another PR

sS373Nn:
search_images_brave() adds an alternative image search function alongside the existing search_images_serpapi and search_images_bing. Brave uses its own independent search index via an official REST API rather than scraping, making it a more stable and reliable option for building image datasets.

sS373Nn:
Almost done! Need to finish fleshing out the SerpApi flexibility for the other available engines.

sS373Nn:
Updated functions to what I think should be an acceptable final submission. May be offline for the rest of the evening, so I wanted to push before I left. Please let me know if you would like any part changed or reverted.

sS373Nn:
@jph00

jph00:
Many thanks - sorry for the long delay. The style and narrative is some distance away from being mergable - partly because this notebook was already in a sorry state. I'm cleaning up the nb locally and reimplementing there. Will release this today.

